PR #147, opened 2026-05-24
feat(graph): symbol_in_degree() primitive (PR-0.2 of v1.15 code-map improvement)

aimasteracc:
## Summary

Adds the symbol-level fan-in primitive needed by **P4 (entry-point ranking honesty)** in the v1.15 code-map improvement PRD.

The current `_find_entry_points` in `codegraph_overview_tool.py` keys off `CallGraph._callers` — function-level call edges, not import-graph fan-in. The recon found two problems:

1. **Conflates 'no callers' with 'true entry point'** — purely inbound symbols (like a class imported by 5 files but never called within a function body) are missed.
2. **Collides with `dead_code` detection** — symbols with zero callers appear in BOTH `entry_points` and `dead_code` lists.

This PR ships **only the primitive**; P4's scoring formula + consumer integration land separately.

## API

```python
@dataclass(frozen=True)
class SymbolFanIn:
    symbol: str
    file_count: int                       # primary signal
    importer_files: tuple[str, ...]       # sorted, deduped
    defining_files: tuple[str, ...]
    ambiguous: bool                       # len(defining_files) > 1

class DependencyGraph:
    def symbol_in_degree(
        self,
        symbol: str,
        *,
        defining_file: str | None = None, # disambiguator
    ) -> SymbolFanIn: ...
```

## Design rationale (locked in `.recon/pr-0-2-design.md`)

- **Method on DependencyGraph** (not free function) — state lives with the graph cache; eviction is automatic via `_cache_key_for` fingerprint.
- **Rich frozen dataclass** (not bare int) — P4 needs both `file_count` AND `importer_files` (for evidence rendering).
- **`defining_file` kwarg** — disambiguates when multiple files define the same symbol name.
- **`file_count = len(deduped set)`**, not edge count — repeated `from X import Foo` in one file still counts as 1 importer.
- **Populated in the existing `_build()` walk** — zero new IO, no `.ast-cache/index.db` coupling, no networkx churn.

## Scope guardrails

- ✅ Python only in PR-0.2. Other languages return empty set as documented limitation.
- ✅ Additive change to `project_graph.py` — `dependents_of()` / `find_cycles()` / `BlastRadius` / `to_dict()` existing keys all unchanged (G1-G4 regression guards).
- ✅ Submodule-as-name guard: `from . import sub` resolving to `sub.py` doesn't falsely register `sub` as a symbol importer.
- ❌ No `__all__` / visibility / export-signal detection (that's P4's scoring concern).
- ❌ No alias tracking, wildcard imports, or transitive re-exports.

## Test plan

- [x] 24 new `symbol_extractors` tests (Python top-level def/class extractor; decorator unwrapping; nested-def exclusion; unicode identifiers)
- [x] 13 new `TestSymbolInDegree` cases — R1-R9 spec coverage (simple count, dedup, ambiguous, defining_file filter, relative import, stdlib non-indexing, JS empty symbols) + G1-G4 regression guards
- [x] 0 regression on the 251-file change-impact focused suite (150 tests across 8 affected modules)
- [x] Pre-commit hooks: ruff lint + ruff format + mypy + bandit + detect-secrets + pyupgrade all pass
- [ ] Manual macOS gate-check (CLAUDE.md §2.2 requirement for `project_graph.py` foundational changes) — covered by full unit suite passing on darwin

## Files touched

| Type | Path |
|---|---|
| ➕ new | `tree_sitter_analyzer/symbol_extractors.py` (~170 lines) |
| ➕ new | `tests/unit/test_symbol_extractors.py` (~195 lines, 24 tests) |
| ✏️ edit | `tree_sitter_analyzer/project_graph.py` (additive: dataclass + 3 state dicts + 1 method + 1 `to_dict` key) |
| ✏️ edit | `tests/unit/test_project_graph.py` (appended `TestSymbolInDegree` class, 13 tests) |

## Unblocks

- **P4** — entry-point ranking honesty: `entry_score = α·log(1+symbol_in_degree(sym).file_count) + β·is_public + γ·has_main_signature + δ·export_signal - dead_penalty·dead_overlap`

### DIFF
--- a/tests/unit/test_project_graph.py
+++ b/tests/unit/test_project_graph.py
@@ -205,6 +205,188 @@ class TestDependencyGraph:
 
 
 # ============================================================
+# PR-0.2: symbol_in_degree() — symbol-level fan-in primitive
+# ============================================================
+
+
+class TestSymbolInDegree:
+    """Test the PR-0.2 ``symbol_in_degree()`` API.
+
+    The primitive answers "how many project files import this symbol by
+    name?" — feeding P4's ranked entry-point detection. RED tests track
+    the planner's spec in ``.recon/pr-0-2-design.md``; GREEN tests are
+    regression guards on the existing public surface.
+    """
+
+    @pytest.fixture
+    def py_graph(self):
+        from tree_sitter_analyzer.project_graph import DependencyGraph
+
+        return DependencyGraph(str(PY_PROJECT))
+
+    @pytest.fixture
+    def custom_graph(self, tmp_path):
+        """Build a small focused project for the cases the PY_PROJECT
+        fixture doesn't cover (ambiguous defs, repeated imports, etc.).
+        """
+        from tree_sitter_analyzer.project_graph import DependencyGraph
+
+        (tmp_path / "utils.py").write_text(
+            "def format_thing(x):\n    return str(x)\n"
+            "\n"
+            "def never_imported_helper():\n    return None\n",
+            encoding="utf-8",
+        )
+        (tmp_path / "main.py").write_text(
+            "from utils import format_thing\n"
+            "from utils import format_thing  # duplicate import on purpose\n",
+            encoding="utf-8",
+        )
+        (tmp_path / "cli.py").write_text(
+            "from utils import format_thing\n", encoding="utf-8"
+        )
+        return DependencyGraph(str(tmp_path))
+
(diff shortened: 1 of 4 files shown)


aimasteracc:
Superseded by new PR targeting `develop` (per GitFlow rule: feat/* not allowed; must use feature/* and target develop). Same code, just renamed branch + retargeted base.

PR #151, opened 2026-05-25
chore(mcp): two startup UX fixes — drop fake LoggingCapability + non-blocking metrics

aimasteracc:
## Summary

Fix the `[error] Failed to set MCP server log level: Error: MPC -32601: Method not found` log entry that surfaces in every MCP client (VS Code, Claude Desktop, etc.) on every TSA server start.

## Root cause

`tree_sitter_analyzer/mcp/_server_helpers.py:build_initialization_options` declares `logging=LoggingCapability()` in `ServerCapabilities`. Per MCP spec, clients SHOULD react to that capability by calling `logging/setLevel` to adjust verbosity. But TSA never registered the corresponding handler — every client call gets back JSON-RPC `-32601 Method not found`, which clients log as `[error]`.

## Fix

Drop the `LoggingCapability()` advertisement. Honest answer: we don't implement it, so we shouldn't claim it.

TSA's actual verbosity controls (`TSA_DEBUG`, standard Python logging config) are unchanged.

## Verification

- 219/219 MCP unit tests pass (`tests/unit/mcp/ -k server`)
- mypy clean
- Smoke test: `build_initialization_options(...).capabilities.logging == None`
- Tool discovery still works (59 tools)

## Why now

User reported the error in real-world VS Code MCP logs:
```
2026-05-25 10:28:23.561 [error] Failed to set MCP server log level:
  Error: MPC -32601: Method not found
```

It's been silently noisy across all clients since the capability was added. The fix is a 4-line change, zero behaviour-change risk.

If we ever want real client-driven log-level routing, the inverse PR registers a `set_logging_level` handler and re-adds the capability — small follow-up.

### DIFF
--- a/tree_sitter_analyzer/mcp/_server_helpers.py
+++ b/tree_sitter_analyzer/mcp/_server_helpers.py
@@ -17,17 +17,24 @@ def build_initialization_options(
     """Build MCP initialization options without bloating the runtime method."""
     from mcp.server.models import ServerCapabilities
     from mcp.types import (
-        LoggingCapability,
         PromptsCapability,
         ResourcesCapability,
         ToolsCapability,
     )
 
+    # NOTE: We deliberately do NOT advertise ``LoggingCapability()``.
+    # The spec lets clients that see this capability call
+    # ``logging/setLevel`` to adjust server verbosity (see
+    # https://spec.modelcontextprotocol.io/specification/server/utilities/logging/),
+    # but we never registered a handler — so the call returned JSON-RPC
+    # ``-32601 Method not found`` and surfaced as ``[error]`` in every
+    # client log (e.g. VS Code MCP, Claude Desktop). Until we actually
+    # implement set-level routing through our logger, advertising the
+    # capability is a contract lie; dropping it is the honest fix.
     capabilities = ServerCapabilities(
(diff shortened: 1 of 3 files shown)


aimasteracc:
## Added 2nd commit: codegraph_metrics non-blocking on empty cache

User report: `codegraph_metrics({sections:[cache,health]})` 'never returns'. Reproduced as 50 s real time on a 1500-file repo with cold AST cache — MCP clients have 30 s default timeouts, so the call drops client-side.

**Root cause**: `_get_cache()` → `ensure_indexed()` synchronously triggers `cache.index_project(max_files=5000)` when AST cache is empty.

**Fix**: `auto_build=False` kwarg on `ensure_indexed`. `codegraph_metrics` is read-only, shouldn't build cache inside an MCP request. Empty-cache path already had the right hint (`Run ast_cache mode=index first`) — we just stopped blocking.

**Result**: cold-start 50.34 s → **0.01 s** with the hint surfaced.

4279 MCP unit tests pass. Zero behaviour change for warm cache or explicit-build callers (`codegraph_autoindex` etc.).

Both fixes are MCP UX cleanups (declared-but-broken contracts) — bundling into one chore PR.

PR #226, opened 2026-05-30
Merge release v1.17.2

aimasteracc:
### Changes

- Fix ~14ms cold-import penalty by deferring `asyncio` import in the hot module-load chain (every CLI invocation and MCP server spawn benefits)
- Fix O(files × symbols) linear scan for single-character symbol queries; route to FTS5 prefix search instead
- Fix path-traversal vulnerability in ASTDiffTool and SemanticClassifyTool: validate `file_path` against project root before constructing `git show` commands
- Fix CWE-703: replace seven `assert`-based security guards in MCP tools with unconditional `raise ValueError` so they survive Python `-O` optimisation
- Eliminate duplicate module-to-file mapping logic in `call_graph.py` to prevent future silent divergence between the build and cache-restore code paths


aimasteracc:
Superseded by v1.17.4. Verified read-only: v1.17.2's code is 100% contained in release/v1.17.4 (only the version string differs). v1.17.2 is already published on PyPI. Closing per the new CLAUDE.md 'Branch-divergence prevention' rule #6 (don't let CONFLICTING release PRs accumulate). Its code reaches main via the next release (v1.17.5).

PR #227, opened 2026-05-30
Merge release v1.17.3

aimasteracc:
Expert team improvements release v1.17.3

## Changes

- Fix critical path traversal vulnerability in ast_diff_tool: file paths from MCP callers are now validated against the project root boundary before any filesystem access
- Fix critical performance regression in get_stats() fallback: replaces per-row Python json.loads() deserialization (5,000+ blobs on large repos) with a single SQLite json_extract + json_array_length expression, 10-100x faster
- Fix hot-loop regex recompilation in route detector: three inline re.match() calls inside the AST node walker are promoted to module-level pre-compiled constants, eliminating repeated pattern compilation on large repos
- Fix depth-7 nesting and 88-line method in call_path._bidirectional_sql: duplicated BFS forward/backward expansion blocks extracted into a shared _bfs_expand_step() helper, halving method length and bringing nesting within TSA limits
- Fix uncached PRAGMA table_info in constraint evaluator hot path: schema check result is now memoized per DB file via functools.lru_cache, eliminating a repeated metadata round-trip on every constraint evaluation call

## Fixes Summary

- Replaced Python-level JSON deserialization loop in get_stats() fallback with a single SQLite native expression using json_extract() and json_array_length(). The old code fetched every row from ast_index, called json.loads() on each symbols_json blob, and computed the sum in Python. The new code pushes all computation into SQLite with COALESCE(SUM(json_array_length(json_extract(symbols_json, '$.symbols'))), 0), eliminating per-row Python allocation and JSON parsing.
- Added path traversal boundary checks to ASTDiffTool.validate_arguments() for diff_files mode. Both old_file and new_file are now validated via self.security_validator.validate_file_path(..., base_path=self.project_root), raising ValueError with a clear message if either path escapes the project root.
- Extracted three inline re.match() patterns inside the decorator-walking hot loop to module-level compiled constants. Added _FLASK_ROUTE_RE, _PY_SHORTCUT_RE, and _FASTAPI_SHORTCUT_RE after the existing constants.
- Extracted a private _bfs_expand_step static method from the two duplicated BFS expansion blocks inside _bidirectional_sql. Method shrank from 88 lines to ~48 lines; max nesting depth dropped from 7 to 4.
- Added @functools.lru_cache(maxsize=16) helper _has_callee_resolved_file(db_path: str) -> bool that runs PRAGMA table_info(ast_call_edges) once per unique DB file path and caches the boolean result.


aimasteracc:
Superseded by v1.17.4. Verified read-only: v1.17.3's code is 100% contained in release/v1.17.4 (only the version string differs). v1.17.3 is already published on PyPI. Closing per CLAUDE.md 'Branch-divergence prevention' rule #6. Code reaches main via v1.17.5.

PR #284, opened 2026-06-04
docs(rfc): RFC-0003 — coverage-aware test-gap analysis

aimasteracc:
## RFC-0003 (draft)

Spec-first, per the RFC process. Requesting review **of the design** before any implementation (RFC-0001's value was Codex catching an architecture dead-end at the spec stage).

### The problem (user's insight)
TSA's `test_gap` decides "is this tested?" by **naming convention** (`test_foo` ⇒ `foo` tested) — wrong both ways: fake tests hide real gaps; indirectly-covered code reports false gaps. We've spent **3 dogfood rounds** (#281–#283) manually doing *static finds candidates → pytest-cov confirms*. This RFC builds that into the tool.

### The proposal
- **Consume `coverage.json`** as ground truth (coverage.py already produces it; `check_patch_coverage.py` already reads it in-repo). Graceful fallback to naming, clearly labeled via a `source` field.
- **Don't reimplement runtime tracing** — that abandons TSA's static identity and duplicates coverage.py. Stand on its shoulders.
- **Layer TSA's unique graph value**: who-should-test-it, change impact, complexity priority, why-missing — things coverage.py has no graph to provide. 1+1>2.
- **Fix an orphan**: `CodeGraphTestGapTool` exists + is tested but is wired into no facade and no CLI flag (Wave-C cutover left it unreachable). Wire it into `health` + `--test-gap` with parity.

### Key boundary stated honestly
coverage.py is precise because it does **runtime tracing** — a static tool *cannot* match that (indirect dispatch/reflection need execution). Not a TSA gap; a paradigm boundary. Hence: consume, don't reimplement.

RED-first test plan + acceptance criteria + CLI↔MCP parity section included. Please review the design, open questions (3) especially.

### DIFF
--- a/rfcs/0003-coverage-aware-test-gap.md
+++ b/rfcs/0003-coverage-aware-test-gap.md
@@ -88,9 +88,22 @@ Resolution order for "is symbol S tested?":
 
 1. **coverage source** (preferred): if a `coverage.json` is found (explicit
    arg → `coverage.json` in project root → `.coverage` via `coverage.py` API),
-   a symbol is *tested* iff ≥1 executable line in its `[start_line, end_line]`
-   range is marked executed for its file. This is line-precise and immune to
-   both fake-test and indirect-dispatch errors.
+   a symbol is *tested* iff ≥1 **body** line in its range is marked executed —
+   **excluding the declaration lines** (the `def`/`class` line and any decorator
+   lines).
+
+   > **Why exclude declaration lines (Codex P1 on #284):** coverage.py records
+   > the `def`/`class` *statement* line as executed on mere **import**. If we
+   > counted any executed line in `[start_line, end_line]`, a function that is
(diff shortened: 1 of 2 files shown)


aimasteracc:
## Codex review triage

**P1 — Don't count definition-line execution as coverage** → ✅ **Accepted, RFC updated** (this is exactly why we spec-first). You're right: coverage.py marks the `def`/`class` line executed on mere import, so 'any executed line in [start,end]' would mark imported-but-uncalled symbols as tested — hiding the fake/missing tests this RFC exists to expose.

Updated the spec: a symbol is covered only when an executed line lies **strictly inside its body** (declaration + decorator lines excluded; body span derivable from the AST node the parser already extracts). Added a RED test ("declaration-line immunity") and an acceptance criterion. Caught at the spec stage before any code — the RFC process working as intended (cf. RFC-0001).

PR #286, opened 2026-06-04
fix(index): exclude build dirs (bin/obj/packages/target) — index full hang on C#/Java

aimasteracc:
## Bug (real user report)

A user running TSA via **VS Code + Copilot on a Windows C# project** reported `index full` **never finishes** on an otherwise small project.

## Root cause

The filesystem-walk exclude sets did not include compiled-language **build dirs**. `.vs`/`.git` are skipped via the `name.startswith('.')` check, but `bin`/`obj`/`packages`/`target`/`.gradle`/`vendor`/`Pods` are **not dot-prefixed** and were walked. A C# `bin`/`obj` or NuGet `packages/` dir holds huge generated/compiled output → indexing appears to hang.

Secondary issue: `_EXCLUDE_DIRS` was **duplicated across 12+ modules** and had drifted (6–23 entries each).

## Fix

- New shared **`constants.EXCLUDE_DIRS`**: VCS + language caches + build artifacts for C#/.NET, Java, Rust, Go, Swift, C++.
- Switched the three filesystem walkers users actually hit to it: **index** (`_ast_extraction`), **health scoring** (`health_scorer`), **project graph** (`project_graph`). `health_scorer` keeps its extra test-data excludes via set union.

## Verification

Mock C# project with `src/app.cs` + `bin/` + `obj/` + `packages/` → index now picks up **only `src/app.cs`**. 139 related tests green; ruff clean.

Remaining 9 `_EXCLUDE_DIRS` copies (call_graph, dead_code, etc.) can migrate to the shared constant in a follow-up — they're not on the user's index-full path.

### DIFF
--- a/tree_sitter_analyzer/_ast_extraction.py
+++ b/tree_sitter_analyzer/_ast_extraction.py
@@ -14,6 +14,7 @@ import os
 import sqlite3
 from typing import Any
 
+from .constants import EXCLUDE_DIRS
 from .core.parser import Parser
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


aimasteracc:
Clarification: the actual user project is **Java** (Maven/Gradle), not C# as the commit message says — `target/` (Maven compile output + unpacked dependency jars + `generated-sources/` containing annotation-processor-generated `.java`) is what made `index full` hang. The fix covers it: `target`/`build`/`.gradle`/`bin`/`out` are all in the new shared `EXCLUDE_DIRS`. Verified with a mock Maven layout — index picks up only `src/main/java/...`, skips `target/generated-sources` and `.gradle`.

PR #289, opened 2026-06-04
fix(mcp): defer ~54ms tool-registry build out of the initialize handshake

aimasteracc:
## Problem (startup)

`TreeSitterAnalyzerMCPServer.__init__` eagerly ran `_create_tool_registry` (~54ms) to build the 8 facade tools. But the registry is only needed for `tools/list` and `tools/call` — **never for the MCP `initialize` handshake**. Building it during `__init__`/`create_server` pushed spawn→initialize to the edge of the client's connect window, so a loaded machine intermittently saw the server stuck at `status: pending` (the headless-benchmark + company-laptop symptom).

## Fix

Defer the registry to a lazy, idempotent `_ensure_registry()` triggered on first access via the `tools` / `tool_instances` properties (and `call_tool` / `set_project_path`). `register_tools` reads the registry only **inside its handler bodies**, so `create_server` no longer materialises it and `initialize` returns before the 54ms is paid.

Cheap eager components stay eager: analysis engine + security validator (legacy code-scale + per-call security pre-check) and the legacy alias tools + optional universal tool — all lightweight and registry-independent (`attach_tool_aliases` never reads the registry, so it is safe to run before the registry exists).

## Measured (gin, warm cache)

| | before | after |
|---|---|---|
| `Server.__init__` | 61 ms | **29 ms** |
| spawn→`initialize` response | 215 ms | **181 ms** (−16%) |

The 54ms now lands on the first `tools/list`, after the client already considers the connection established. Combine with `MCP_TIMEOUT=30000` on the client for extreme-load belt-and-suspenders.

## Tests

RED-first `TestRegistryDeferral`: registry not built at init, not built by `create_server` (the property that keeps initialize fast), built exactly once on first `.tools` access, and the eager components are present before the build. Full `test_server.py` 43/43 + `tests/unit/mcp/` 4492/4492 green; ruff + mypy clean.

### DIFF
--- a/tests/unit/mcp/test_server.py
+++ b/tests/unit/mcp/test_server.py
@@ -208,6 +208,58 @@ class TestServerCreation:
                 server.create_server()
 
 
+class TestRegistryDeferral:
+    """Startup fix: the ~54ms facade tool registry is built lazily.
+
+    The MCP ``initialize`` handshake never needs the registry — only
+    ``tools/list`` / ``tools/call`` do. Building it during ``__init__`` /
+    ``create_server`` pushed spawn→initialize to the edge of the client's
+    connect window, causing intermittent ``status: pending``. These tests
+    pin the contract that the registry is NOT materialised until first access.
+    """
+
+    @pytest.fixture
+    def server(self):
+        with tempfile.TemporaryDirectory() as tmp:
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #301, opened 2026-06-04
fix(resolver+nav): bind callees to the correct definition (language + test-shadow + unresolved)

aimasteracc:
## Problem — found by dogfooding the chain DSL / nav callees on this repo

Querying the callees of the Python `_resolve_entry_points` returned **wrong same-name binds**: callees resolved to `examples/ModernJavaScript.js` (JS), a Swift `func sorted` body, and real cache methods (`fts_search`/`fts_search_ranked`) bound to their **test mocks** instead of source. One root bug class: **callee resolution binds a call to the wrong definition of the same name.** It's both a correctness bug and a token-cost bug (foreign/guessed bodies inflate the response — the very thing that makes TSA pricier than CG per call).

Callee resolution is **not centralized** — 4 independent paths each resolve callee→file, so each fix had to land in the right writer (traced via `provenance`+`callee_resolution` on the edge).

### A. Cross-language gate — a call must not bind across languages
- `synapse_resolver`: central `_is_cross_language` gate on the cascade output → falls through to `unknown`.
- `_ast_cache_unresolved._choose_candidate`: drop foreign-language candidates before scoring.
- `callee_resolution.CalleeResolver`: gate the global bare-name fallback to caller language.
- **Codex P2 ×2 fixed:** JS↔TS are one interop family (`_language_family.languages_compatible`); caller language derived from file extension when a file has no indexed functions.

### B. Body inlining — no foreign / guessed bodies
- `call_path_enrich._resolve_def`: carries language, gated by `lang_hint` → no cross-language body.
- `symbol_body_inline._record_span`: an **unresolved callee gets no inlined body** (it would be a bare-name guess — the Swift `sorted` case).

### C. Test-shadow demotion — production calls bind to source, not test mocks
- `_choose_candidate` and `CalleeResolver` global fallback: when the caller is **not** a test file, demote test-only definitions (only when a source def exists; test callers keep test helpers). Uses canonical `utils.test_detection.is_test_file` (#294).

## Live proof (in-process, fixed code; the long-running MCP server holds stale code mid-session)
- `get`/`entries`/… : ~~JS `ModernJavaScript.js`~~ → Python.
- `sorted` : resolution `unknown`, **no Swift body** inlined.
- `fts_search`/`fts_search_ranked` : ~~test mock~~ → **`tree_sitter_analyzer/_ast_cache_query.py`** (real source).

## Tests — RED-first, 0 regressions
New: cross-language (resolution+body), JS/TS family, function-less-file language, unknown-callee-no-body, test-shadow demotion (both writer paths). Full sweep green across synapse/unresolved/callee/cross_file/call_path/body-inline/context/nav (**~900 tests**), lint+mypy clean.

### DIFF
--- a/tests/unit/mcp/tools/test_symbol_body_inline.py
+++ b/tests/unit/mcp/tools/test_symbol_body_inline.py
@@ -104,6 +104,42 @@ def test_inline_neighbor_bodies_attaches_body_to_each(tmp_path):
     assert "SMALL_MARKER" in small["body"]["content"]
 
 
+def test_inline_neighbor_body_does_not_cross_languages(tmp_path):
+    """A Python callee with no Python def must not inline a foreign-language body.
+
+    Regression: ``sorted()`` (a Python builtin call, no Python definition) has a
+    call-site ``file`` but no matching def there, so ``_resolve_def`` fell back to
+    ``candidates[0]`` — grabbing a Swift ``func sorted`` body and inlining it
+    under the Python callee. The record carries ``language='python'``; a
+    cross-language body must be suppressed (callee stays body-less).
+    """
+    (tmp_path / "corpus.swift").write_text("func sorted() -> Int {\n    return 1\n}\n")
+    cache = _build_cache(tmp_path)
+    cache.get_conn.return_value.execute(
+        "INSERT INTO ast_index (file_path, symbols_json, language) VALUES (?,?,?)",
+        (
+            "corpus.swift",
+            json.dumps(
+                {
+                    "symbols": [
+                        {"name": "sorted", "kind": "function", "line": 1, "end_line": 3}
+                    ]
+                }
+            ),
+            "swift",
+        ),
+    )
+    cache.get_conn.return_value.commit()
+    # Python callee record: a ``sorted()`` call from a Python file, no end_line.
+    neighbors = [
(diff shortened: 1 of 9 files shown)


aimasteracc:
## Codex review triage (both P2s — fixed)

| # | Codex finding | Verdict | Action |
|---|---|---|---|
| P2-1 | `synapse_resolver/__init__.py:384` — strict equality rejects valid JS↔TS imports (`typescript != javascript`) in gradual-migration projects | **Real** | Introduced `languages_compatible()` treating `{javascript, typescript, jsx, tsx}` as one family (matching `CrossFileResolver._register_module_path`). Applied at all four gate sites. Test: `test_global_fallback_allows_js_ts_family`. |
| P2-2 | `callee_resolution.py:184` — `_source_language` returns `''` for a file with no indexed functions → ungated global fallback re-opens cross-language binds for module-level calls | **Real** | `_source_language` now falls back to `language_from_path()` (pure extension lookup, no stat). Test: `test_global_fallback_gated_by_language_for_function_less_file`. |

Both fixed in `c06b514a` via a shared `_language_family` helper. Full regression sweep (295 tests across synapse/unresolved/callee/call_path/body-inline/context/cross_file/nav) green, lint+mypy clean.

@codex review

aimasteracc:
@codex review

PR #327, opened 2026-06-07
fix(mcp): project metrics dead-code key mismatch — stop reporting every function as dead (F1)

aimasteracc:
## The trust-breaker (found by dogfooding)
`project action=metrics` / `overview` reported call-graph stats where **every function was simultaneously an entry point AND dead code**: `entry_points == dead_code_candidates == total_functions == 28239`, `files_with_functions == 1`, and `suggested_next_steps: "28239 dead code candidates found"`. An agent's first orienting call thus told it the entire codebase is dead — directly refuting TSA's core "trust the graph" thesis — while the dedicated `health action=dead` tool correctly reported **1372**. Two surfaces disagreed 20×.

## Root cause (one character)
`CachedCallGraph.all_functions()` dicts key the path under `"file"` (`FunctionRef.to_dict`, [call_graph.py:90](../blob/develop/tree_sitter_analyzer/call_graph.py)), but `_collect_call_graph_metrics` read `f.get('file_path','')`. Every function string became `"::name"` (empty path) and never matched the real-path call-edge keys, so `callers_map`/`callees_map` stayed empty → every function had 0 callers (entry) and 0 callees (dead).

## Fix
`'file_path'` → `'file'` (one line, + a comment pinning the key to `FunctionRef.to_dict`). On the TSA repo: **28239/28239/28239/1 → entry 19335 / dead 3394 / files 1666** (sane).

## Test plan (RED-first)
- [x] `TestCallGraphMetricsSanity`: indexes a 2-file project where `caller()` calls `helper()`; asserts `dead_code_candidates < total_functions`, `entry_points < total`, and `files_with_functions >= 2`. Verified RED on the old key, GREEN on the fix (independent reviewer reproduced).
- [x] Independent adversarial review: pass, 0 findings. Confirmed no same-class bug elsewhere (`call_graph_impact.py` already uses `"file"`).
- [x] 15 file tests + 573 blast-radius pass; ruff + mypy clean.

## Scope / follow-up
Surgical one-char fix only. The metric's naive dead-code count (3394: no-callers-AND-no-callees) still differs from `health action=dead` (1372: transitive reachability + test filtering) because they are different engines — **converging the two surfaces onto one engine is a separate follow-up** (panel's NEXT wave), not required to remove the trust-breaker.

Found via a dogfood + expert-panel sweep; this is item #1 ("the one thing") of the TSA summit roadmap.

### DIFF
--- a/tests/unit/test_codegraph_metrics_tool.py
+++ b/tests/unit/test_codegraph_metrics_tool.py
@@ -119,3 +119,54 @@ class TestExecuteWithCache:
         assert result["cache"]["total_files"] == 42
         assert result["cache"]["total_symbols"] == 1000
         assert result["cache_indexed"] is True
+
+
+class TestCallGraphMetricsSanity:
+    """F1 regression: project metrics must not degenerate to 'every function is
+    both an entry point AND dead code'. Root cause was a key mismatch —
+    all_functions() dicts key the path under 'file', but the metric collector
+    read 'file_path', so every function string had an empty path and never
+    matched the call-edge keys (real paths), making callers/callees maps useless.
+    """
+
+    def _index(self, tmp_path):
+        from tree_sitter_analyzer.ast_cache import ASTCache
+
+        proj = tmp_path / "pkg"
+        proj.mkdir(parents=True, exist_ok=True)
+        (proj / "__init__.py").write_text("# pkg\n")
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #331, opened 2026-06-07
feat(mcp): RFC-0009 — nav context inline full entry bodies + entry-first ranking

aimasteracc:
Implements [RFC-0009](https://github.com/aimasteracc/tree-sitter-analyzer/blob/develop/rfcs/0009-nav-context-self-sufficiency.md) parts **A + B** — the core self-sufficiency fix that closes TSA's turn-count gap vs CodeGraph (its only measured deficit).

## What & why
`nav context` is the documented PRIMARY call, but a dogfood (`resolve_callee → resolve_java_callee`) showed its response **did not contain the answer**, forcing follow-up reads — the turn explosion. Two fixes to `_build_code_blocks`:

- **A — two-tier body budget**: entry-point / task-named symbols inline their FULL body up to `_MAX_ENTRY_BODY_LINES=160` (covers the 127-line motivating target); tangential expansion nodes keep the 16-line cap. The blanket 16-line cap was **net-negative** — smaller first response, but extra turns, and turns dominate end-to-end cost.
- **B — entry-first ranking**: named entry points get a code block BEFORE high-degree hubs (a cache accessor was beating the actual dispatcher, which got no block). Entry nodes marked `is_entry=True` in `_nodes_from_hits`.

## Dogfood (before → after)
| | before | after |
|---|---|---|
| `resolve_java_callee` body | 16 lines + '… 80 more' | **full 127 lines inline** |
| `resolve_callee` dispatcher | **absent** | present |
| `_route_cache.get` noise | present | **gone** |

→ the task is now answerable in **one call**.

## Test plan (RED-first)
- [x] `test_entry_point_body_inlines_full_under_budget` — full body, no truncation marker (verified RED on the 16-line cap)
- [x] `test_entry_point_ranked_before_high_degree_noise` — entry wins the slot vs a high-degree hub (verified RED on degree-only ranking)
- [x] 42 context tests pass; ruff + mypy clean

## Scope / follow-ups (per RFC-0009, tracked)
- **C** query tokenisation (de-noise generic verbs like 'dispatch' that still match unrelated event dispatchers) — separate PR.
- The **n≥5 benchmark turn-drop measurement** (the RFC's final acceptance gate) — depends on the benchmark run-id infra fix landing first. RFC stays `accepted` until C + measurement complete.

### DIFF
--- a/tests/unit/test_codegraph_context_tool.py
+++ b/tests/unit/test_codegraph_context_tool.py
@@ -1384,3 +1384,72 @@ def test_build_related_symbols_groups_by_file() -> None:
 
     b_group = next(g for g in groups if g["file"] == "b.py")
     assert b_group["symbols"] == ["beta:5"]
+
+
+def test_entry_point_body_inlines_full_under_budget(tmp_path: Path) -> None:
+    """RFC-0009 A: an entry-point symbol whose body fits the entry budget inlines
+    in FULL — no '… more lines' truncation marker — so the agent answers in one
+    call. The old blanket 16-line cap truncated it (RED)."""
+    from tree_sitter_analyzer.mcp.tools.codegraph_context_tool import (
+        _build_code_blocks,
+        _node_id,
+    )
+
+    body = "def big():\n" + "".join(f"    x{i} = {i}\n" for i in range(1, 40))
+    f = tmp_path / "big.py"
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #335, opened 2026-06-07
refactor: single-source constants.EXCLUDE_DIRS for vendored/cache dir exclusion

aimasteracc:
## What

`dead_code_analyzer.py` defined its own private `_EXCLUDE_DIRS` set that duplicated — and was a strict subset of — the canonical `constants.EXCLUDE_DIRS` frozenset. This PR migrates the one consumer to import the canonical constant so the excluded-directory policy lives in **one** place and can't drift.

## Why

The set of vendored/cache/build dir names (`node_modules`, `.git`, `__pycache__`, `.venv`, `dist`, `build`, `.ast-cache`, ...) was duplicated. `constants.EXCLUDE_DIRS` already existed as the canonical source; `dead_code_analyzer` carried a near-duplicate. Single-sourcing prevents the two from diverging.

## Behavior

- The canonical set is a **strict superset** of the dirs `dead_code_analyzer` previously excluded, so every historical entry is still pruned — **behavior unchanged** for those dirs.
- It additionally covers vendored/build dirs the walker should already skip (`vendor`, `target`, `obj`, `Pods`, `out`, `bower_components`, ...).
- Dot-prefixed dirs remain pruned via the existing `startswith(".")` check in `_iter_candidate_files`.

## Scope guard

Only `dead_code_analyzer` (one consumer) is migrated to keep the diff focused and safe. `SecurityValidator` / `PathResolver` / `BaseMCPTool` / the plugin registry are **untouched** (CLAUDE.md locks).

## Tests (RED-first)

- `test_exclude_dirs_is_canonical_constant` — asserts `dead_code_analyzer._EXCLUDE_DIRS IS constants.EXCLUDE_DIRS`.
- `test_exclude_dirs_covers_previously_excluded_dirs` — asserts the canonical set still covers every dir the module historically excluded.

Verified RED (reverted impl → `is`-identity test fails), then GREEN. Full `tests/unit/test_dead_code_analyzer.py` (33 tests) passes. `ruff check` + `ruff format --check` + `mypy` clean on touched files.

### DIFF
--- a/tests/unit/test_dead_code_analyzer.py
+++ b/tests/unit/test_dead_code_analyzer.py
@@ -330,3 +330,50 @@ def test_excludes_dot_prefixed_vendored_dirs(tmp_path):
     assert not any(".benchmark-repos" in i.file.replace("\\", "/") for i in imports), [
         i.file for i in imports
     ]
+
+
+# ---------------------------------------------------------------------------
+# Canonical EXCLUDE_DIRS single-source (centralize-exclude-dirs)
+# ---------------------------------------------------------------------------
+
+
+def test_exclude_dirs_is_canonical_constant() -> None:
+    """dead_code_analyzer must source its excluded-dir set from the single
+    canonical ``constants.EXCLUDE_DIRS`` rather than defining its own private
+    copy that can drift. The module-level name must BE the canonical object."""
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #352, opened 2026-06-07
feat(resolver): conservative Ruby callee resolver (RFC-0010 second wave)

aimasteracc:
## Summary

New-files-only per-language **Ruby** callee resolver for RFC-0010's second wave. It replaces the generic Python cascade for `.rb` callers with a SAFE, self-contained resolver that never regresses — `unknown` when unsure, never a wrong binding.

Auto-discovered via `languages/__init__` (pkgutil) — **no existing file is edited**, so it lands without merge conflicts or worktree races alongside the other second-wave language PRs.

## Resolution tiers

| Tier | Rule |
|---|---|
| **local** | bare `foo()` → same-file top-level `function` (via `file_symbols`); `self.foo` → single-class gate. In Ruby `self` IS the current object (handled like Java's `this`), so `self.foo` binds only when the file defines exactly ONE class — never to a sibling class's method. |
| **project** | a bare name with exactly ONE Ruby-family project definition of a bare-callable kind (top-level `function`). Class/method/module globals are excluded (need a receiver / are constants). |
| **builtin** | namespaced Ruby-core call matched on the FULL dotted name. **Restricted to the frozen `Math` module** (`Math.sqrt`, `Math.sin`, …) — precision over recall per RFC-0008. Suppressed when the project owns the receiver name in a Ruby file. |
| **unknown** | conservative default. |

## THE MOAT (the #1 requirement)

This resolver **never binds a callee to a symbol in a non-Ruby file**. Ruby is in no interop family, so `languages_compatible("ruby", owner_lang)` only admits another `ruby` (or empty/unknown) owner. A same-name Python/Java symbol stays `unknown`. The mandatory no-cross-language-mis-wire test (`test_no_cross_language_mis_wire_global` + 2 companions) locks this in.

## Conservative builtin tier rationale

Ruby method names are extremely overloaded (open classes, monkey-patching, DSLs). Bare `puts`/`require`/`raise`/`new`/`each` carry no proof of being the core API, so they are **deliberately excluded** — an empty tier for them is correct. Only the `Math.`-namespaced static forms are kept (a frozen core module user code rarely re-creates with the same method).

## Tests (RED-first)

- 22 unit tests in `tests/unit/test_ruby_method_resolution.py`, including the mandatory cross-language moat tests, the `self`-method single-class / sibling-ambiguity guards, the bare-callable-kind gate, and the builtin shadow-suppression gate.
- RED verified: removing `ruby.py` makes the suite error (`ModuleNotFoundError`).
- Resolver regression green: `uv run pytest tests/ -k 'resolver or synapse or registry'` → 375 passed, 2 skipped (Java byte-parity intact).
- `ruff check` / `ruff format --check` / `mypy` clean on the new files.

### DIFF
--- a/tests/unit/test_ruby_method_resolution.py
+++ b/tests/unit/test_ruby_method_resolution.py
@@ -1,0 +1,383 @@
+"""RFC-0010 second wave: Ruby per-language callee resolver.
+
+Unit-level tests that drive ``resolve_ruby_callee`` against a hand-built context
+(no full index needed). They assert the four guarantees of a SAFE,
+self-contained resolver:
+
+* same-file ``local`` resolution from ``file_symbols`` / ``file_class_methods``
+  (Ruby ``self`` IS the current object, so ``self.foo`` binds via the
+  single-class gate; bare ``foo`` binds via the flat top-level scan),
+* CONSERVATIVE ``builtin`` classification on the FULL dotted call name only,
+* ``project`` resolution gated to Ruby, and
+* the MOAT: a callee whose name also exists in another language's file must
+  NEVER bind to that file — it stays ``unknown`` (or the Ruby same-file def).
+"""
+
+from __future__ import annotations
+
+from tree_sitter_analyzer.synapse_resolver._registry import (
+    get_language_resolver,
+    registered_languages,
+)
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

PR #353, opened 2026-06-07
feat(synapse): SAFE per-language C callee resolver (RFC-0010)

aimasteracc:
## Summary

Adds a self-contained, conservative, **SAFE** per-language callee resolver for **C** (`.c` / `.h`) under RFC-0010's second wave. It REPLACES the generic Python cascade for C callers without regressing — when unsure it returns `unknown`.

**New-files-only** (the `languages/` package auto-discovers the module, so no existing file is edited):
- `tree_sitter_analyzer/synapse_resolver/languages/c.py`
- `tree_sitter_analyzer/synapse_resolver/languages/_c_constants.py`
- `tests/unit/test_c_method_resolution.py`

## Cascade (most-specific first)

1. **local** — an unqualified call to a same-file free `function`.
2. **project (single-global)** — exactly one same-language project-wide definition of the name.
3. **libc (`stdlib`)** — a conservative libc free-function name, classified only when the project owns **no** compatible-language function of that name (project shadowing wins).
4. **unknown** — receiver-qualified calls (`obj->fn` / `cfg.fn` — struct-field / function-pointer access whose target type C does not let us prove), ambiguous multi-global names, and everything else.

## THE MOAT — no cross-language mis-wire

Every project binding is gated by `languages_compatible('c', owner_lang)`, which for C is **directional and asymmetric**: a C caller resolves only `c` (including `.h` headers, indexed as `c`) and must **NEVER** bind a `cpp` / `objc` / Python / Go definition that merely shares a name (`languages_compatible('c', 'cpp')` is `False`). A mandatory no-cross-language test covers this, including the directional `c → cpp` case.

## Conservative tiers (RFC-0008 lesson)

C has **no namespace qualifier** (no `std::` analogue) and **no member-method dispatch**, so unlike the C++ resolver there is no namespace tier and no owner-class map. The libc tier is intentionally tiny and high-confidence — `malloc`/`calloc`/`realloc`, the `printf` family, the `mem*` and `str*` families, and the `f*` stdio variants — while pruning collision-prone names (`free`, `open`, `read`, `write`, `close`, `find`, `sort`). An `unknown` edge is never a mis-wire.

## Test plan

- **26 pass + 1 strict xfail.** The xfail nails the known shared-extractor production gap (tree-sitter C `function_definition` exposes its identifier under `function_declarator`, so the generic `_walk_for_symbols` walker omits C free functions; the local tier flips to a hard PASS once that is fixed — root cause is out of this resolver's scope).
- RED-first, including the **mandatory** no-cross-language-mis-wire test (`TestNoCrossLanguageMisWire`, incl. directional `c → cpp`).
- Real-index integration: maps-independent libc tier + moat asserted on a freshly-parsed `.c` + `.py` index.
- `ruff check` + `ruff format --check` + `mypy` clean on all new files.
- Resolver regression suite green: `pytest tests/ -k 'resolver or synapse or registry'` → 375 passed, 2 skipped (Java byte-parity intact).

### DIFF
--- a/tests/unit/test_c_method_resolution.py
+++ b/tests/unit/test_c_method_resolution.py
@@ -1,0 +1,507 @@
+"""Unit + integration tests for the C resolver (RFC-0010 second wave).
+
+Covers the conservative cascade in
+``synapse_resolver/languages/c.py``: local same-file resolution, single-global
+project resolution, the conservative libc free-function tier (shadowed by
+project ownership), and — MANDATORY — the no-cross-language-mis-wire moat
+(a callee whose name also exists in a non-C file must NOT bind to that file).
+
+C is simpler than C++: there are no namespaces (no ``std::`` qualifier), no
+member methods to disambiguate by class. A C call is either an unqualified free
+function, a struct field access (``->`` / ``.`` on a value — never a resolvable
+free function), or a libc free function. The moat is also DIRECTIONAL: a C
+caller must NEVER bind to a C++ definition (``languages_compatible('c', 'cpp')``
+is False), while a C caller MAY resolve another C file / C-tagged ``.h`` header.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from tree_sitter_analyzer.ast_cache import ASTCache
+from tree_sitter_analyzer.synapse_resolver import ResolverContext, resolve_callee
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

PR #354, opened 2026-06-07
feat(synapse): RFC-0010 Kotlin per-language callee resolver

aimasteracc:
## Summary

Adds a self-contained, SAFE per-language callee resolver for **Kotlin** (`.kt`/`.kts`), RFC-0010 second wave. **New-files-only** — registered via the auto-discovery registry, so no existing file is edited.

New files:
- `tree_sitter_analyzer/synapse_resolver/languages/kotlin.py`
- `tree_sitter_analyzer/synapse_resolver/languages/_kotlin_constants.py`
- `tests/unit/test_kotlin_method_resolution.py`

## Resolution tiers (conservative — `unknown` when unsure)

1. **local (bare)** — receiver-less call to a top-level function / class (or single-class implicit-receiver member) in the **caller's own file**. Same file == same language → cannot cross-bind.
2. **local (`this`/`super`)** — member call binds a same-file `method` row **only when unambiguous** (one class defines the name). Two classes defining it → `unknown` (no caller-owner class is threaded through — wave-1 Codex P2 lesson). A receiver call never binds a top-level function.
3. **stdlib** — a tiny set of auto-imported bare stdlib builders (`listOf`/`println`/`require`/…), gated on the project not owning the name in a compatible-language file (shadowing preserved).
4. **unknown** — receiver method calls on a variable (`items.map` — receiver type not carried), cross-file project calls, qualified third-party calls.

## THE MOAT — never cross-language bind

A Kotlin callee is **never** bound to a symbol in another language's file. Local tiers consult only the caller's own file; the stdlib gate is language-aware via `languages_compatible`. Critically, **Kotlin and Java are not one `languages_compatible` family**, so a Kotlin caller never binds a Java symbol and a Java `listOf` never suppresses Kotlin's stdlib tier. Real Kotlin/Java interop would need import/type evidence this resolver doesn't carry — deliberately out of scope.

Mandatory no-cross-language-mis-wire tests included: vs **Java** (same JVM, same name) and vs **Python**.

## Conservative tiers (RFC-0008 lesson)

Bare-method-name `STDLIB_METHODS_KOTLIN` / `EXTERNAL_METHODS_KOTLIN` sets are **intentionally empty** — Kotlin DSL/builder idiom reuses exactly such names; without receiver-type evidence they'd mis-classify. The bare auto-imported top-level builder set is the only classified tier.

## Note on activation

Kotlin call-edge extraction is not yet enabled in `function_extraction._CALL_NODE_TYPES` (python/js/ts/java/go/c/cpp only), so the resolver is **dormant end-to-end** until that separate extractor change lands — same as the merged Rust resolver. It's correct and tested at its contract surface now and activates with zero further work once Kotlin edges are produced.

## Verification

- `ruff check` + `ruff format --check` + `mypy` clean on all new files.
- `tests/unit/test_kotlin_method_resolution.py` — 20 passed (RED-first verified before implementation).
- Resolver regression `uv run pytest tests/ -k 'resolver or synapse or registry'` — 375 passed, 2 skipped (Java byte-parity intact).
- `git show --stat HEAD` — new-files-only (3 files, +761, 0 modifications).

### DIFF
--- a/tests/unit/test_kotlin_method_resolution.py
+++ b/tests/unit/test_kotlin_method_resolution.py
@@ -1,0 +1,374 @@
+"""RFC-0010 Kotlin resolver — SAFE, self-contained per-language callee resolution.
+
+Kotlin call-edge extraction is not yet wired into ``function_extraction.py``
+(only python / js / ts / java / go / c / cpp appear in ``_CALL_NODE_TYPES``), so
+a real ``index_project`` run produces zero Kotlin ``calls`` edges and the
+resolver is never invoked end-to-end. These tests therefore drive the resolver's
+CONTRACT SURFACE directly: build the context from synthetic ``file_symbols`` /
+``global_name_table`` / ``file_languages`` maps (the shapes
+``_build_resolver_context_uncached`` passes) and call ``resolve_kotlin_callee`` —
+the same function the registry dispatch in ``synapse_resolver/__init__.py`` calls.
+
+The resolver is deliberately CONSERVATIVE (RFC-0008 lesson): it resolves
+same-file / same-language local calls (bare top-level functions, and
+``this.``/``super.`` member calls only when the method name is unique in the
+file), classifies a tiny set of auto-imported stdlib TOP-LEVEL functions
+(``listOf`` / ``println`` / ``require``) gated on project non-ownership, and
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

PR #355, opened 2026-06-07
feat(resolver): conservative C# callee resolver (RFC-0010 second wave)

aimasteracc:
## Summary

Self-contained, **SAFE**, per-language callee resolver for **C# (`.cs`)** — RFC-0010 second wave. **New-files-only**: auto-discovered via `languages/__init__`, no existing file edited (zero merge-conflict / worktree-race risk).

Files:
- `tree_sitter_analyzer/synapse_resolver/languages/csharp.py`
- `tree_sitter_analyzer/synapse_resolver/languages/_csharp_constants.py`
- `tests/unit/test_csharp_method_resolution.py`

## Cascade (conservative, moat-safe)

1. **`System.*` fully-qualified BCL namespace** → `stdlib`. The C# analogue of C++ `std::` — `namespace System` is reserved, never project code, so no shadow gate is needed.
2. **local** — same-file `method` / free function / `class`. A `this.`/`base.` member call binds only an **unambiguous** (single-owner-class) same-file method; a name owned by >1 class in the file stays `unknown` (caller's class is unprovable).
3. **project (single-global)** — exactly **one same-language** project-wide definition of an unqualified name. Excluded for `this.`/`base.` member calls.
4. **BCL static-type qualifier tier** — `Console` / `Math` / `Convert` / `Enumerable` / `String` qualified calls → `stdlib`, **shadow-gated** by `_project_owns`: a *same-language* project (or untagged) symbol of that type name suppresses the claim. Pruned HARD — `File`/`Path`/`Directory`/`Task`/`Environment`/`Guid` are deliberately LEFT OUT (collide with domain types).
5/6. **bare-method tiers EMPTY** (RFC-0008): without receiver-type inference, bare C# method names (`Add`/`Select`/`ToString`) collide too heavily to classify. Everything else → `unknown`.

## THE MOAT — never cross-language bind

Every project binding is gated by `languages_compatible('csharp', owner_lang)`. C# is its own family (no symmetric/directional compat with any other tag), so a Python/Java/Go symbol that merely shares a name resolves to `unknown` or a same-language def — **never** the foreign file.

## Tests (RED-first)

31 new tests, including the **mandatory no-cross-language-mis-wire** suite (unit + public-API dispatch + real-index). Real-index integration drives the full `ASTCache` pipeline: the C# symbol walker records methods/classes, so the `local` tier fires **end-to-end** (no xfail needed, unlike C++).

- Targeted: `tests/unit/test_csharp_method_resolution.py` → 31 passed
- Regression `-k 'resolver or synapse or registry'` → **375 passed, 2 skipped** (Java byte-parity intact)
- `ruff check` + `ruff format --check` + `mypy` clean on new files

### DIFF
--- a/tests/unit/test_csharp_method_resolution.py
+++ b/tests/unit/test_csharp_method_resolution.py
@@ -1,0 +1,565 @@
+"""Unit + integration tests for the C# resolver (RFC-0010 second wave).
+
+Covers the conservative cascade in
+``synapse_resolver/languages/csharp.py``: ``System.``-qualified stdlib
+classification, the near-exclusive BCL static-type qualifier tier (shadow-gated
+by ``_project_owns``), local same-file resolution, single-global project
+resolution, the empty bare-method-name tiers, and — MANDATORY — the
+no-cross-language-mis-wire moat (a callee whose name also exists in a non-C#
+file must NEVER bind to that file).
+"""
+
+from __future__ import annotations
+
+from tree_sitter_analyzer.ast_cache import ASTCache
+from tree_sitter_analyzer.synapse_resolver import ResolverContext, resolve_callee
+from tree_sitter_analyzer.synapse_resolver._context import build_resolver_context
+from tree_sitter_analyzer.synapse_resolver._registry import get_language_resolver
+from tree_sitter_analyzer.synapse_resolver.languages._csharp_constants import (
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

PR #356, opened 2026-06-07
feat(resolver): conservative PHP callee resolver (RFC-0010 second wave)

aimasteracc:
## Summary

RFC-0010 second wave: a self-contained, SAFE, **new-files-only** per-language callee resolver for **PHP** (`.php`). Auto-discovered via `languages/__init__` — no edit to any shared file, so it lands without merge/worktree races.

### What it resolves
- **local** — same-file bare free-function calls (incl. namespaced `\App\helper` / `App\helper`, resolved on the last segment) via the caller file's own `file_symbols`; `$this->m()` / `self::m()` / `static::m()` / `parent::m()` method calls bound **only** through the unambiguous single-class gate (the caller's enclosing class is not carried by the edge, so a name defined by two classes stays `unknown` — never the file-wide first match).
- **builtin** — a curated, hand-audited set of PHP **global free functions** (`strlen`, `array_map`, `json_encode`, `count`, `preg_match`, `in_array`, `sprintf`, …) classified only when the project does not itself define a free function of that name (shadowing preserved).
- **unknown** — everything else: `Foo::bar` static / `$obj->bar` instance / `App\Util::run` namespaced-static receiver calls (the class type the edge does not record), ambiguous self-methods, and unresolved bare names. The **external** tier is intentionally **empty** (PHP libraries expose namespaced classes/methods, not distinctive global functions — RFC-0008 precision-over-recall).

### THE MOAT — never cross-language bind
The resolver only ever consults the **caller file's own** maps for a `local` match and never looks up a symbol in another file. The built-in shadow gate is **language-aware** via `languages_compatible` — PHP has no interop family, so a same-named Python/Java symbol never suppresses (nor is bound by) PHP classification. A **mandatory** no-cross-language-mis-wire test covers this both directly and **end-to-end through the production `build_resolver_context`** (a polyglot repo where `helper` exists as both a Python def and a PHP free function — the PHP caller resolves to `app.php`'s symbol id, never `py/helpers.py`).

### Conservative-tier note (PHP has no index-time call edges yet)
PHP symbols (functions/methods/classes) are indexed, but the synapse call-edge extractor does not currently emit `calls` edges for PHP, so the resolver activates whenever such edges appear and is fully unit- + registry-dispatch-tested now (no DB-edge dependency for correctness).

## Test plan
- [x] RED-verified: the 27 tests fail (`ModuleNotFoundError`) without `php.py`, pass with it.
- [x] Mandatory no-cross-language-mis-wire test (direct + end-to-end via production context builder).
- [x] `ruff check` + `ruff format --check` + `mypy` clean on all 3 new files.
- [x] Resolver regression green: `pytest tests/ -k 'resolver or synapse or registry'` → **377 passed, 2 skipped** (Java byte-parity intact).
- [x] `git show --stat HEAD` = new-files-only (3 files, +918, −0).

## Files (new only)
- `tree_sitter_analyzer/synapse_resolver/languages/php.py`
- `tree_sitter_analyzer/synapse_resolver/languages/_php_constants.py`
- `tests/unit/test_php_method_resolution.py`

### DIFF
--- a/tests/unit/test_php_method_resolution.py
+++ b/tests/unit/test_php_method_resolution.py
@@ -1,0 +1,459 @@
+"""RFC-0010 second wave: the PHP per-language callee resolver.
+
+PHP calls come in several syntactic shapes the resolver must classify
+correctly without ever guessing:
+
+* bare / same-file       — ``helper()`` resolves to a same-file ``function``.
+* namespaced free call   — ``\\App\\helper()`` / ``App\\helper()`` resolve on the
+  last segment as a same-file free function (the ``\\`` split is a namespace,
+  not a class/instance receiver).
+* global built-in        — ``strlen()`` / ``array_map()`` / ``json_encode()``
+  classify as ``builtin`` ONLY when the project does not itself define a free
+  function of that name (shadowing preserved).
+* self method call       — ``$this->m()`` / ``self::m()`` / ``static::m()`` /
+  ``parent::m()`` bind ``local`` ONLY when exactly ONE class in the caller file
+  defines ``m`` (the enclosing class is not carried by the edge → ambiguous
+  across classes stays ``unknown``).
+* receiver call          — ``Foo::bar()`` (static) / ``$obj->bar()`` (instance)
+  / ``App\\Util::run()`` carry a class type the edge does not record; the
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

PR #384, opened 2026-06-07
fix(formatters): apply CSV control-char sanitizer to the two remaining writers

aimasteracc:
## What

PR #381 added `tree_sitter_analyzer/formatters/_csv_safety.py` (`csv_safe_row` / `csv_safe_cell`) and applied it to the CSV writers it touched, but left **two writers unguarded** against the Python 3.10 `csv` divergence. This finishes the same coherent concern so **every** CSV writer is py3.10-safe.

The two remaining writers:

1. **`formatters/base_formatter.py` → `_format_csv`** (the public `format_type="csv"` path used by every `BaseTableFormatter` subclass). Previously used `csv.writer(..., escapechar="\\")`. The `escapechar` **doubled literal backslashes** in ordinary fields (`C:\tmp` → `C:\\tmp`) — a format regression for Windows paths and regex strings that Codex flagged on #381. Removed `escapechar` and wrapped each data row (`_csv_field_row` / `_csv_method_row` results) with `csv_safe_row`.
2. **`_legacy_table_formatter_csv.py` → `format_csv`** (the `--table csv` CLI path). Wrapped its class/method/field data rows with `csv_safe_row`. Imported lazily inside a small `_csv_safe_row` wrapper to avoid a `formatters`-package circular import (`formatters.__init__` → `formatter_registry` → … → this module).

Headers are constant string literals — **not** sanitized; only data rows.

## Why

On **Python 3.10**, `csv.writer` with no `escapechar`:
- (a) raises `_csv.Error: need to escape, but no escapechar set` on a NULL byte, and
- (b) writes a bare `\r` **unquoted**, so the CSV fails to read back (`new-line character seen in unquoted field`).

Python 3.11+ does neither. Setting `escapechar="\\"` silences (a) but doubles backslashes — the wrong fix. The right fix is the sanitizer: it strips C0/DEL controls + bare CR, preserves `\t`/`\n`, and leaves the dialect (and backslashes) untouched.

## Real Python 3.10 verification

Ran CR / lone-CR / CRLF / NUL / a Windows-path name (`C:\tmp\Foo`) / a regex name (`\d+\w*`) through **both** writers on real CPython 3.10.20:

```
PY 3.10.20
ALL CHECKS PASSED: both writers round-trip, no bare CR, backslashes not doubled
```

Asserting, for every input and both writers: (a) `csv.reader` round-trips with no exception, (b) no bare `\r` in output, (c) backslashes are NOT doubled (`C:\\tmp` absent).

## Tests

- New focused file `tests/unit/formatters/test_csv_control_char_safety_remaining.py` mirroring the #381 pattern: parametrized control-char + CR round-trip + backslash-preservation assertions for **both** writers. RED on py3.10 before the fix, GREEN after.

## Golden masters stayed green

The sanitizer is a no-op on normal input, so output bytes are unchanged for normal data:

```
uv run pytest tests/integration/core/test_golden_master_regression.py -q -k csv
18 passed
```

Also green: full `tests/unit/formatters/` (2058 passed), legacy/table formatter suite (289 passed), `ruff`, `mypy`, and all pre-commit hooks (bandit, detect-secrets, etc.).

### DIFF
--- a/tests/unit/formatters/test_csv_control_char_safety_remaining.py
+++ b/tests/unit/formatters/test_csv_control_char_safety_remaining.py
@@ -1,0 +1,160 @@
+"""CSV control-character safety for the two remaining CSV writers.
+
+Companion to ``test_csv_control_char_safety.py``, covering the two writers that
+PR #381 left out:
+
+1. ``BaseTableFormatter._format_csv`` (the public ``format_type="csv"`` path,
+   exercised here via ``PythonTableFormatter``). It previously used
+   ``escapechar="\\"`` — which doubled literal backslashes in ordinary fields
+   (Windows paths, regex) — now switched to the strip-via-``csv_safe_row``
+   approach.
+2. ``_legacy_table_formatter_csv.format_csv`` (the ``--table csv`` CLI path).
+
+Guards the same Python 3.10 vs 3.11+ ``csv`` divergence: on 3.10 a NULL byte
+raises ``_csv.Error: need to escape, but no escapechar set`` and a bare ``\\r``
+is written unquoted (producing an unreadable CSV). These tests assert the
+contract version-independently — RED on Python 3.10 before the fix, GREEN after.
+"""
+
+from __future__ import annotations
+
+import csv
+import io
+
+import pytest
+
+from tree_sitter_analyzer._legacy_table_formatter_csv import (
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

PR #482, opened 2026-06-11
fix(structure): class_detail delivers full contract — fields, visibility, extends, no closures (#455)

### LINKED ISSUE #455: [P2][dogfood] class_detail returns half its contract — no fields/visibility/inherited; closure listed as method
## Summary
`structure action=class_detail` delivers about half its advertised surface. The facade schema promises "detailed class inspection: **fields, methods, visibility, inherited members**" — the response contains only a flat method list.

## Repro (develop @ 32412d6e)
```
structure action=class_detail symbol=BaseMCPTool
```
Returns `method_count: 17` + methods (name/line/file/is_override) and nothing else. Missing vs the contract:
- **fields**: none returned (BaseMCPTool has instance attributes)
- **visibility**: no field at all (underscore-private vs public not distinguished)
- **inherited members**: none (no base class info either — not even `extends: ABC`)
- `is_override` is `false` for all 17 — including `execute`, which subclasses override; on the base it's trivially false everywhere, providing no signal.

Bonus defect: `_wrapped` (line 316–320) is a **local closure inside `__init_subclass__`**, listed as a class method. A nested function is not part of the class API; this mirrors the nested-scope attribution problem in #452 from the other direction.

## Impact
class_detail is the "one call instead of reading the class" tool; an agent using it to plan an override or to enumerate the API gets a list that (a) omits what it was promised and (b) includes a phantom method.

## Suggested fix
Return fields + visibility + base classes + inherited members (the AST cache has receiver/contains data for this), exclude nested closures from method lists, and align the schema text with whatever is actually supported until then.

aimasteracc:
Closes #455.

## What was missing (vs the advertised "fields, methods, visibility, inherited members")
Response had only a flat method list. Now delivers:
- **extends**: base class list from the AST cache's `parents`
- **fields**: class-level attrs + `self.x` instance attrs scanned across the WHOLE class body (not just `__init__` — BaseMCPTool, the issue's own repro, delegates writes to `_apply_project_root`; an init-only scan returned `[]`; guard test pins its exact 4 fields)
- **visibility** on every method/field (public / protected `_x` / private `__x`, dunder=public)
- **inherited_methods**: from nearest in-project ancestor when indexed; otherwise honest `inherited: {available: false, reason}` — no faking
- **closure fix**: `_wrapped` (local closure inside `__init_subclass__`) no longer listed as a method — line-range containment filter

## Tests
48 RED-first tests (25 RED initially) incl. the BaseMCPTool repro guard; tools dir 946 passed; contracts 53 passed; ruff clean; class_inspect_tool.py coverage 95.3%.

### DIFF
--- a/tests/unit/mcp/tools/test_class_inspect_tool.py
+++ b/tests/unit/mcp/tools/test_class_inspect_tool.py
@@ -406,7 +406,7 @@ class TestEdgeCases:
         assert response["success"] is True
         # fields should be empty (OSError fallback) but rest of response is intact
         assert response["fields"] == []
-        assert len(response["methods"]) > 0  # methods still work
+        assert len(response["methods"]) == 4  # methods still work (Animal pin)
 
     def test_class_with_duplicate_self_assignments_deduplicated(
         self, tmp_path: Path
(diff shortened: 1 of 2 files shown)


aimasteracc:
## Codex review triage

|---|---------|---------|--------|
| P2 | annotation-only 字段(`name: str` 无默认值)被静默漏掉 — dataclass/Pydantic 常见形态 | **真问题,已修** | 正则补 annotation-only 分支;新测试钉 Config 类精确 3 字段 ["name","retries","_token"] + visibility/kind(commit 711aa605);BaseMCPTool 回归检查 4 字段不变 |
| P2 | OSError 回退测试用 `> 0` 违反精确断言锁定规则 | **真问题,已修** | 改为 `== 4`(Animal 在他处已钉 4)(commit 711aa605) |

PR #514, opened 2026-06-11
fix(kotlin): bind import extraction to the node type the grammar actually emits

aimasteracc:
Closes #512

## Root cause (two layers)

1. **Wrong binding**: `extract_imports()` dispatched on `"import_header"` — the installed tree-sitter-kotlin grammar emits `'import'` for import statements, so the handler never fired and every Kotlin file reported 0 imports.
2. **Keyword-leaf double-count**: each `'import'` statement node contains a leaf token of the SAME type (the `import` keyword). Naive binding yields 3×2=6. Guarded in `extract_import()`: a node whose text has no module path (`len(parts) < 2`) is the keyword leaf → skip.

`"import_header"` stays in the dict for older grammar versions that do emit it.

## How it surfaced

Ratchet batch 6 (#511): the old `assert len(imports) >= 0` was vacuously green over the broken extraction; the batch pinned `== 0` with an extraction-gap marker. This PR flips the pin to `== 3` (RED-first: pin flipped before the fix, observed `assert 0 == 3` failure, then fixed to green) — the ratchet closing its loop as designed.

## Verification

- RED→GREEN on `test_extract_imports_extractor` (`0 == 3` fail → pass)
- 137 passed (full kotlin plugin test file); 278 passed (all kotlin tests in tests/unit/languages)
- `--show-languages` / `--show-extensions` sane (CLAUDE.md plugin-interface rule 7)
- Lead independently reproduced: 3 imports extracted with correct module names (`kotlin.collections.List`, `java.util.HashMap`, `kotlinx.coroutines.launch`)

## Diff

17 insertions / 5 deletions across kotlin_plugin.py (binding), kotlin_helpers.py (leaf guard), test re-pin.

PR #564, opened 2026-06-12
fix(csharp): real return types (primitive+generic) and implicit-public interface members

### LINKED ISSUE #536: [P2][QC-sweep] C#: 4 return types void-ified + interface members marked private
## C#: return types void-ified for non-void methods + interface members marked private
Develop @ 9fb13638, examples/Sample.cs:
1. **4 wrong return types** (shown `void`): `User.IsValid`→bool, `User.ToString`→string, `IUserRepository.GetById`→User, `GetAll`→IEnumerable&lt;User&gt;. Primitive-keyword and generic returns both affected; C# was otherwise the best performer in the sweep (D/I/A/B/E/G/H all ✓).
2. **Interface member visibility**: all 5 IUserRepository methods show `visibility=private`; C# interface members are implicitly **public**.

aimasteracc:
Closes #536 (P2, QC-sweep)

## Root causes (AST-dump confirmed before any code)

1. **Return types void-ified**: `extract_method_declaration` read `child_by_field_name("type")` — the tree-sitter-c-sharp grammar names that field **`returns`** for method_declaration, so the lookup always returned None → `void`. One-word fix + the 4 existing unit tests that had pinned the WRONG field name corrected.
2. **Interface members private**: `determine_visibility([])` defaulted to private; C# interface members are implicitly public. Now: no explicit modifier + parent chain reaches interface_declaration → public (class members without modifiers correctly stay private).

## Golden masters (5-kind protocol, #514 precedent)

full/compact/csv: exactly the 4 corrected signatures (`():bool`, `():string`, `(id:int):User`, `():IEnumerable<User>`); toon: 5 interface members private→public; plugins/csharp.json unchanged (tracks names/counts only). Nothing else moved.

## Verification

- 9 new RED→GREEN integration tests pinning exact values; 263 total green
- Lead independently ran Sample.cs: all 4 signatures correct (pasted in PR discussion)
- Ratchet exit 0


aimasteracc:
@codex review

aimasteracc:
## Codex review triage

| Comment | Verdict | Action |
|---|---|---|
| P2 (csharp_helpers:260): 隐式 public 只盖了 method_declaration——接口的 property/event 仍被 determine_visibility([]) 判 private | **真问题** | 已修:抽公共 `_apply_interface_implicit_public` helper,method/property/event 三处统一应用;新增 2 个 RED 测试(接口 property `Name` 与 event `Changed` → public)。20+ 集成测试绿,csharp golden 套件绿(Sample.cs 接口只有方法,goldens 不再变动) |

PR #558, opened 2026-06-12
test: pin exact assertions — batch 11 (integration) + ratchet baseline down

aimasteracc:
## Summary

Layer-2 loose-assertion cleanup, batch 11 — first integration-tests batch. 38 enforcement-pattern hits → 25 exact pins + 13 documented exemptions; ratchet baseline 1332 → 1307 (re-measured after the parallel batch-10 merge — reconciliation commit included).

| File | Hits → Remaining |
|---|---|
| tests/integration/cli/test_cli_async.py | 20 → 0 |
| tests/integration/mcp/test_user_story_4_integration.py | 18 → 0 |

## Exemption analysis (hard-step C, three-question test each)

13 exemptions, all in cli_async, all genuinely nondeterministic output-length asserts: 10× JSON output embeds a UUID temp path (length varies per run), 1× argparse help wraps to terminal width (COLUMNS), 1× version string embeds sys.executable basename (python vs python3), plus 1 vacuous dead-branch assert REMOVED (rc=1 always; else-branch never ran).

## Exact pins of note

- user_story_4: all 18 pinned over tempfile fixtures (import occurrences ==14, java percentage ==31.78, Service.java ==318 bytes…)
- cli_async fixed-content messages pinned by exact length (==441/==78/==1413/==101/==54) — wording drift goes red by design

## Test plan

- [x] 19 + 16 tests green (`-n 0`); ratchet exit 0
- [x] Baseline 1307 re-measured post-merge (raw outputs in pod report)

PR #586, opened 2026-06-12
feat(ci): AST-based loose-assertion ratchet — multiline asserts no longer evade

aimasteracc:
Closes the ratchet-blindspot tech-debt issue (multiline assert evasion found during #558).

## What changed

`check_loose_assertions.sh` → 4-line wrapper over a new Python AST checker (same CLI contract: diff-scoped, baseline mode, exit codes, exemption marker — the workflow step is byte-identical). Assert nodes' test expressions are inspected for GtE/Gt against int literals; the `# ratchet: nondeterministic` marker now works ANYWHERE within the assert's source segment (the batch-11 marker-placement dance is over).

## Baseline conscious re-pin: 1180 (grep) → 1083 (AST), methodology note in the file

AST catches ~multiline evaders the grep missed, AND drops grep's false skips (asserts with `==` elsewhere on the line were skipped wholesale; version-spec strings excluded naturally). Net −97.

## Verification

- 12 RED→GREEN self-tests (multiline caught, closing-paren marker exempts, ==/string not flagged, property-file skip, exact counter)
- Clean-tree run vs origin/develop: exit 0 (pasted); synthetic bad diff: exit 1 with the finding (pasted)
- Workflow untouched

### DIFF
--- a/scripts/check_loose_assertions.py
+++ b/scripts/check_loose_assertions.py
@@ -1,0 +1,278 @@
+#!/usr/bin/env python3
+"""AST-based loose-assertion ratchet — Layer-1 CI gate.
+
+Parses Python test files with the ``ast`` module and flags ``assert``
+statements whose test expression contains a Compare node with a ``>=`` or
+``>`` operator applied against an integer literal (or where the left side is a
+``len()`` call with the same operators).
+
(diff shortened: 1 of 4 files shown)


aimasteracc:
@codex review

PR #585, opened 2026-06-12
fix(kotlin,scala): inheritance extraction + Kotlin primary constructors

aimasteracc:
Closes #561, closes #562; completes #567's scope-B (Kotlin primary_constructor).

## Three extraction gaps, one contract surface

1. **Kotlin inheritance**: `delegation_specifiers` now read — `constructor_invocation` child → superclass, bare `user_type` → interfaces (live: `Success : Result(value)` → superclass=Result; `UserManager : Displayable, Serializable` → 2 interfaces). The #560 resolver spellings, so outline extends/implements light up for Kotlin.
2. **Scala inheritance**: `extends_clause` read — first type_identifier = superclass, `with`+type_identifier = traits (live: `Circle extends Shape with Drawable with Serializable`).
3. **Kotlin primary constructors**: `primary_constructor` in the dispatch map — Function with is_constructor=True, name = class name (the C++ same-name convention), params from class_parameters.

## Verification

- 22 RED→GREEN tests; 3,528 suite green (2 pre-existing Swift-grammar fails unrelated)
- Kotlin goldens regenerated, ALL 5 kinds (#514 protocol): Function 8→12 / element_total 22→26 — exactly the 4 primary ctors; Scala has no golden (verified)
- 1 conscious re-pin: `len(functions) == 3 → == 4` (data class primary ctor now emitted)
- Ratchet exit 0

PR #587, opened 2026-06-12
fix(ratchet): scope loose-assertion gate to added diff lines; include renames

aimasteracc:
## Summary

Follow-up to the Codex review on #586 (both findings verified live before fixing):

- **P1 (live-confirmed on #585)**: `check_diff` scanned entire changed files, so a one-line re-pin in `test_kotlin_plugin.py` re-flagged an unrelated pre-existing baseline assert and failed CI. Violations now count only when the assert's source segment (`lineno..end_lineno`) overlaps a line the PR actually added (parsed from `git diff -U0` hunk headers).
- **P2**: `--diff-filter=AM` omitted renamed files — a rename+edit could smuggle a loose assertion past the gate. Now `AMR`; `+++ b/` carries the new path for renames.

Deliberate behavior: editing any line *inside* an existing multiline loose assert still flags it — touching a loose assert means fixing it.

## Test plan
- [x] 4 new git-fixture self-tests: unrelated-edit passes (P1), added assert flagged, rename+add flagged (P2), multiline partial-touch flagged
- [x] 16/16 self-tests green (`-n 0`)
- [x] Live sanity: checker exit=0 on this branch vs origin/develop; ruff + mypy clean

### DIFF
--- a/scripts/check_loose_assertions.py
+++ b/scripts/check_loose_assertions.py
@@ -60,6 +60,7 @@ class Violation(NamedTuple):
     path: str
     lineno: int
     snippet: str
+    end_lineno: int = 0
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

Issue #538, opened 2026-06-12
[P3][QC-sweep] Theme I residual: Go alias/interface members, Rust mod/trait abstract, C++ namespace, Python ABC, TS abstract flag


aimasteracc:
## Theme I residual + per-language container gaps (sweep digest)
Develop @ 9fb13638 — smaller container/typing gaps from the 15-language sweep, grouped:
- **Go**: `type StringSlice = []string` alias absent (9/10 containers); interface method signatures not nested (Reader/Writer/ReadWriter all methods=[], N4); interface embedding not reflected (N5).
- **Rust**: `mod sample_module` not a container; trait's abstract method skipped (`Displayable.display` missing — got 1/2 methods, only the default-impl `summary`).
- **C++**: `namespace math` not captured; `math::max` loses its qualifier (N3).
- **Python**: `class Animal(ABC)` typed `class`, not `abstract_class` (TS keyword-based detection works; ABC pattern doesn't).
- **TS**: `abstract validate()` carries no `abstract: true` on the method even when the container is abstract_class.
- **Kotlin**: `LegacyUserManager.get` return shown `Unit`; source infers String from `= "legacy"` (expression-body inference).

aimasteracc:
Fixed by #583 — per-item verdicts in the PR description. Recovered: Go type_alias + interface embedding, Rust trait abstract methods (trait-only via _inside_trait guard; extern FFI declarations excluded per Codex P2), Python ABC/abc.ABC → abstract_class (qualified bases now reach superclass extraction), TS abstract method flag. C++ namespace was a false positive (already a Package).

Scope-B items split into focused follow-ups: #588 (Go interface method nesting), #589 (Rust mod container), #590 (C++ qualified-name loss), #591 (Kotlin Unit inference — already fixed by #593 in flight).

aimasteracc:
**Retested on develop @ 5c1f476e (#583 + #595 + #593) — 6 of 8 legs CONFIRMED FIXED ✓**
| Leg | Status |
|---|---|
| Go `StringSlice` alias in containers | ✓ |
| Go interface methods nested (Reader.Read 1/1) | ✓ (#595) |
| Go ReadWriter embedding → implements ['Reader','Writer'] | ✓ |
| Python `Animal(ABC)` → type=abstract_class | ✓ |
| Rust trait abstract method (Displayable 2/2 incl. `display`) | ✓ |
[...]

Issue #537, opened 2026-06-12
[P1][QC-sweep] Swift claimed supported but RuntimeError on fresh install — grammar not in base deps, no hint


aimasteracc:
## Swift: "supported" but RuntimeError on a fresh install — grammar dependency not in base package
Fresh worktree + `uv` install of the package (develop @ 9fb13638): `examples/sample.swift` → `RuntimeError: Unsupported language: swift`; root cause `ModuleNotFoundError: tree_sitter_swift`. Yet `--show-supported-languages` lists swift, and README counts it among the 13 actively-resolved languages.
If the swift grammar lives in an extras group (`[all]`?), then (a) `--show-supported-languages` must report install-state honestly (supported vs available-with-extra), and (b) the README language claims should state the install requirement. A PyPI `[mcp]` user hitting a Swift repo gets a hard RuntimeError with no hint to `pip install tree-sitter-swift`. (Error message should carry the fix.)

aimasteracc:
Persona-1 UX walkthrough adds the full root-cause picture for this issue (develop @ 2e5a3680): the README's canonical Get Started one-liner and all 3 MCP config snippets install `[mcp]` — but tree-sitter-swift (and the other 5-language grammars) live in `[all]`. Worse, the troubleshooting row for `unsupported language` says *'Update to ≥ 1.12.x'* — wrong fix for anyone on 1.23. Complete fix set: (a) decide `[mcp]` includes grammars OR switch the 4 README snippets to `[all,mcp]`; (b) replace the stale troubleshooting row with the extras command; (c) the RuntimeError itself should print `pip install "tree-sitter-analyzer[all]"`. A Swift iOS newcomer currently dead-ends with zero recovery path at all three layers.

aimasteracc:
已由 #559 修复并合入 develop(手动关闭)。三腿:缺语法时错误信息直接给 pip 命令(extras 名读自真实 pyproject);--show-supported-languages 每语言报 installed/install_hint(importlib find_spec 探针,Codex P2 加固:loader-less 的 json 也诚实报 not-installed);README 三语 troubleshooting 注明 extras 安装。info_commands 史上首批测试落地(96.21% 覆盖,规则 7 的零测试警告终结)。

aimasteracc:
**Retested on develop @ 0a89d055 (#559) — CONFIRMED FIXED ✓**
[...]

PR #593, opened 2026-06-12
fix(kotlin): expression-body functions no longer claim Unit return

### LINKED ISSUE #591: Kotlin: expression-body functions report Unit instead of inferred return type
Split from #538 scope-B.

**Symptom**: `fun get(key: String) = "legacy"` reports `return_type=Unit`; the expression body infers String. Any expression-body function without an explicit type is mislabeled Unit — actively wrong, worse than empty.

**Where**: `tree_sitter_analyzer/languages/kotlin_helpers.py` return-type extraction (defaults to Unit when no explicit type node).

**Scope note**: full inference is a non-goal; minimum honest fix = stop claiming Unit when an expression body exists (emit the literal's type for trivial literals, else unknown/empty), per the honesty-over-fabrication convention from #537.

**Acceptance**: expression-body samples stop reporting Unit; explicit-type and block-body behavior unchanged; goldens re-pinned exactly.

aimasteracc:
Closes #591 (split from #538 scope-B).

## Summary
`fun get(key: String) = "legacy"` reported `return_type=Unit` — the Unit default applied whenever no explicit `:` type exists, but expression bodies infer their type. Claiming Unit was actively wrong.

**Fix** (honesty over fabrication, #537 precedent — no full inference):
- expression body + trivial literal → literal type (String / Int / Double / Boolean)
- expression body + anything else (calls, `42L`, `0xFF`) → `""` (unknown; matches the Go plugin convention, and the formatter suppresses the empty suffix)
- explicit type / block body → unchanged (Unit stays correct for block bodies)

## Test plan
- [x] RED-first: 9 expression-body cases failed pre-fix, 3 baseline cases green
- [x] GREEN: 14/14 new tests; kotlin regression 164 + 126 + formatter 37 all passed (-n 0)
- [x] Goldens: full/compact/csv each change exactly 1 line (`Sample.kt` `get` → `fun(): String`); toon/plugins-json legitimately byte-identical (no return_type in those surfaces)
- [x] Ratchet gate exit=0; new code fully patch-covered (stub nodes with explicit parent=None)

Lead independently spot-checked live: `get→'String' n→'Int' f(call)→'' explicit→'Long' block→'Unit'`.

Note: Scala has the same Unit-sentinel bug (`_scala_return_type`) — tracked separately.

### DIFF
--- a/tests/golden_masters/compact/kotlin_sample_compact.md
+++ b/tests/golden_masters/compact/kotlin_sample_compact.md
@@ -21,5 +21,5 @@
 | fetchUserAsync | (1):Result<User> | pub | - | 37-40 | - |
 | display | (0):String | pub | - | 42-44 | - |
 | toTitleCase | (0):String | pub | - | 52-54 | - |
-| get | (0) | pub | - | 59-59 | - |
+| get | (0):String | pub | - | 59-59 | - |
 | main | (1) | pub | - | 62-65 | - |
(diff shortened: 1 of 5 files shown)


aimasteracc:
@codex review

Issue #532, opened 2026-06-12
[P2][QC-sweep] Theme A residual: nested-container methods misattributed in 5 languages (inner class/namespace/module/impl<T>)


aimasteracc:
## Theme A residual: methods misattributed when the owner is a NESTED container (5 languages)
Top-level class method ownership is fixed (#436/#474/#482), but every *nested* container shape still leaks methods to the wrong owner (develop @ 9fb13638):
- **Java**: InnerClass.innerMethod + StaticNestedClass.nestedMethod attributed to outer `Test`; both nested classes show 0 methods.
- **TS**: class inside namespace — `DataProcessing.BatchProcessor` shows methods=[]; its 4 methods land on the namespace.
- **Scala**: Outer/Inner under NestedExample → 4 methods pulled up; owners show 0.
- **Ruby**: classes inside a module — User(15)/AdminUser(6)/Session(3) ALL show 0 methods; all 24 land on module `Authentication`.
- **Rust**: `impl<T> Container<T>` block → `Container.new` lands top-level; Container shows 0 methods (plain `impl User` is correct 4/4).
Pattern: ownership resolution stops at the first enclosing container instead of the innermost. Same innermost-span idea as #484 (nested functions) but for the container axis.

aimasteracc:
Datapoint from the #535 retest (develop @ f443c246): Ruby module-nested ownership unchanged — `Authentication` module carries all 24 member records (now bare-named post-#565), `User`/`AdminUser`/`Session` still 0; top-level `UserRepository` is correctly 6/6. The Ruby row of this issue is now the canonical tracker for that case.

aimasteracc:
已由 #570 修复并合入 develop。五语言嵌套容器归属双层修复:API 面 find_class_name 改最内层 span(首匹配 bug)+全函数元素归属;DB 面补 Ruby method/singleton_method 节点与 Rust impl_item type 字段(泛型剥离);Codex P2 加固:局部函数(函数 span 内)保持无主。innermost 规则第四次应用(#474/#484/#485/#532)。

aimasteracc:
**Retested on develop @ b7d75cad (#570) — NOT fixed at the agent surface. Keep open.**

All five original repros are byte-identical to the pre-fix measurements (GetCodeOutlineTool):
| Case | Expected | Got (unchanged) |
|---|---|---|
| Java InnerClass / StaticNestedClass | 1 / 1 | **0 / 0** (Test still holds 14) |
[...]

PR #595, opened 2026-06-12
fix(go): interface method signatures extracted and owned by their interface

### LINKED ISSUE #588: Go: interface method signatures not nested under interface container (methods=[])
Split from #538 scope-B (PR #583 fixed type_alias + embedding; this needs its own pass).

**Symptom** (sweep N4): `Reader`/`Writer`/`ReadWriter` interfaces all report `methods=[]` — `method_elem` signatures inside `interface_type` are not extracted/attached to the interface container, so an agent sees empty interfaces.

**Where**: `tree_sitter_analyzer/languages/_go_type_helpers.py` (interface walk now collects embedded types via `_go_interface_embedded` — method signatures need the same treatment, plus ownership attachment per the innermost-span rule used in #532/#474).

**Acceptance**: golden go.json Reader/Writer show their method signatures with name+params+return; ReadWriter shows embedding (already) + no phantom methods; exact-pin counts.

aimasteracc:
Closes #588 (split from #538 scope-B).

## Summary
`type Reader interface { Read(p []byte) (n int, err error) }` reported methods=[] — `method_elem` nodes inside `interface_type` were never extracted. Now each emits a Function with `receiver_type` = interface name (innermost-span ownership convention, #532/#474), `is_method=True`. Embedded interfaces (`type_elem`) skipped — no phantoms. `type_alias` interfaces included (mirrors extract_type_declaration).

method_elem carries the same name/parameters/result fields as function_declaration, so the existing shared builders apply — the fix is one new helper + dispatch entries.

## Test plan
- [x] RED-first: 2 failed pre-fix → 12 passed (pins: params `["p []byte"]`, return `"(n int, err error)"`, receiver_type Reader; ReadWriter zero phantoms; empty/anonymous interfaces; 5 guard stubs with explicit parent=None)
- [x] Regression 44 + association 25 + golden suites 78+18 passed (-n 0)
- [x] Goldens both halves regenerated: go.json Function 18→20 (+Read +Write from examples/sample.go), tables +2 rows, toon counts 18→20 — every line traced
- [x] Conscious re-pin: method_count 18→20 with docstring
- [x] Ratchet exit=0; new lines patch-covered; ruff/mypy clean

Lead independently spot-checked live: Read → receiver_type=Reader, params/return exact, no ReadWriter phantoms.

### DIFF
--- a/tests/golden_masters/compact/go_sample_compact.md
+++ b/tests/golden_masters/compact/go_sample_compact.md
@@ -10,6 +10,8 @@
 ## Funcs
 | Func | Sig | V | L | Doc |
 |------|-----|---|---|-----|
+| Read | (1):(n in | E | 55-55 | - |
+| Write | (1):(n in | E | 61-61 | - |
 | NewService | (2):*Ser | E | 80-86 | - |
 | Name | (0):s | E | 89-91 | - |
 | IsRunning | (0):b | E | 94-98 | - |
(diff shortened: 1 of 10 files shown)


aimasteracc:
@codex review

PR #592, opened 2026-06-12
test: ratchet batch-14 — pin 56 loose assertions in top-4 AST-ranked files

aimasteracc:
## Summary
Loose-assertion ratchet batch 14 — first batch ranked by the new AST counter (#586). Top-4 files (14 violations each), every `>=`/`> N` pinned to the measured exact value:

| File | Pins |
|---|---|
| tests/unit/test_symbol_resolver.py | 14 |
| tests/unit/test_project_graph.py | 14 |
| tests/unit/core/test_queries_javascript_enhanced.py | 14 |
| tests/integration/mcp/test_user_story_2_integration.py | 14 |

Baseline: **1083 → 1026** (recorded; clean develop actually measures 1082 on this base — the 1-count drift is documented in the baseline history line, true batch delta = 56).

Contract 5b notes: no dead-branch mocks found; byte-count pins guarded with `write_text(..., newline="\n")` (6 fixtures); zero exemption markers; several pins upgraded to exact list/set equality.

## Test plan
- [x] All 4 files re-run individually `-n 0`: 22/51/27/15 passed
- [x] `check_file` per-file re-run: 0 violations each
- [x] `check_loose_assertions.py origin/develop` exit=0
- [x] `--baseline` measures 1026 == recorded (lead independently re-verified)

PR #596, opened 2026-06-12
fix(rust): mod blocks extracted as container elements

### LINKED ISSUE #589: Rust: mod blocks not captured as container elements
Split from #538 scope-B.

**Symptom**: `mod sample_module { ... }` produces no container element — items inside are extracted but the module itself is invisible, so module-level structure/ownership is lost (compare: C++ namespace → Package works).

**Where**: `tree_sitter_analyzer/languages/rust_plugin.py` extractor dispatch (no `mod_item` handler).

**Acceptance**: `mod_item` yields a container (Package-like, matching the C++ namespace convention); nested items remain owned per innermost-span rule; goldens re-pinned exactly.

aimasteracc:
Closes #589 (split from #538 scope-B).

## Summary
`mod sample_module { ... }` produced no container element. New `extract_packages` emits a Package per `mod_item` (mirrors the C++ namespace → Package convention): inline mods with body span, declaration-only `mod tests;` with its single line (it's the only trace of the file-module mapping in lib.rs — skipping it reproduces the invisibility this issue fixes). Nested mods each emit their own Package. No parent walks — no depth-cap surface.

## Test plan
- [x] RED-first: 6 failed pre-fix → 9 passed; regression across all rust test files 164 passed (-n 0)
- [x] Goldens both halves: rust.json +Package(sample_module) element_total 21→22; toon `package: null` → `{name: sample_module, line_range: (2, 88)}` (verified against examples/sample.rs); full/compact/csv byte-identical (rust table views don't render package metadata) — regenerated and confirmed
- [x] Ratchet exit=0; all new lines patch-covered (guards via stubs with explicit parent=None)
- [x] Rebased onto develop after #592 landed mid-task; gates re-run green

Lead independently spot-checked live: nested `outer`/`inner` both Packages with correct spans, `mod decl_only;` single-line span, nested fn still extracted.

### DIFF
--- a/tests/golden_masters/plugins/rust.json
+++ b/tests/golden_masters/plugins/rust.json
@@ -3,9 +3,10 @@
     "Class": 5,
     "Function": 9,
     "Import": 1,
+    "Package": 1,
     "Variable": 6
   },
-  "element_total": 21,
+  "element_total": 22,
   "names_by_type": {
     "Class": [
       "BorrowedItem",
@@ -26,6 +27,9 @@
     "Import": [
       "use std::collections::HashMap;"
     ],
+    "Package": [
+      "sample_module"
+    ],
     "Variable": [
       "active",
       "email",
@@ -39,6 +43,7 @@
     "Class",
     "Function",
     "Import",
+    "Package",
     "Variable"
   ]
 }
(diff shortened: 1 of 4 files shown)


aimasteracc:
@codex review

PR #597, opened 2026-06-12
fix(scala): expression-body defs no longer claim Unit return

### LINKED ISSUE #594: Scala: expression-body defs report Unit instead of inferred return type (same family as #591)
Found during #591 (Kotlin expression-body Unit misreport): `_scala_return_type` in the Scala plugin uses the same unconditional Unit sentinel, so `def get(key: String) = "legacy"` likely reports Unit.

**Fix shape**: port the #591 approach — trivial literal → literal type; other expression bodies → "" (unknown, Go convention); explicit-type and block-body behavior unchanged. Verify first with a live parse (the pod did not confirm Scala's exact node shapes — diagnosis is from code reading only).

**Acceptance**: mirror of #591 — RED-first tests, exact pins, goldens re-pinned if scala sample shows expression-body defs.

aimasteracc:
Closes #594 (same family as the merged Kotlin fix #593).

## Summary
`def get(key: String) = "legacy"` reported `return_type=Unit` — live-confirmed (the issue was filed from code reading; the pod reproduced it first). Mirrors the #593 contract exactly:
- explicit type / block body → unchanged (Unit stays correct for block bodies)
- expression body + trivial literal → String/Int/Double/Boolean — `string` in tree-sitter-scala covers raw `"""x"""` too, so the Kotlin Codex-P2 raw-string gap cannot recur here
- suffixed/hex ints (`42L`/`0xFF`), calls, blocks → `""` (unknown, honesty over fabrication)

## Test plan
- [x] RED-first: 13 failed pre-fix, 4 unchanged-behavior pins green → GREEN 17/17
- [x] Regression: languages -k scala 142 passed (-n 0)
- [x] Goldens: zero scala golden surfaces render return types (no plugins/scala.json, no scala table goldens; corpus histogram + 86-symbol wiring pin unaffected — verified, both suites green)
- [x] Ratchet exit=0

Lead independently spot-checked live: get→String, n→Int, f(call)→'', raw→String, explicit→Long, block→Unit, 42L→''.

### DIFF
--- a/tests/unit/languages/test_scala_expression_body_return.py
+++ b/tests/unit/languages/test_scala_expression_body_return.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python3
+"""Scala expression-body defs must not claim ``Unit`` (issue #594).
+
+``def get(key: String) = "legacy"`` used to extract ``return_type="Unit"``.
+The expression body infers ``String``; claiming ``Unit`` is actively wrong.
+Same family as the Kotlin fix for #591/#593 — the contract mirrors
+``test_kotlin_expression_body_return.py`` exactly:
+
+* explicit return type → use it (unchanged)
+* block body ``{ ... }`` with no explicit type → ``Unit`` (correct, kept)
+* expression body ``= <expr>`` with no explicit type:
+  - trivial literal → its pinned type (String / Int / Boolean / Double)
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #567, opened 2026-06-12
fix(languages): is_constructor for Python/Ruby/PHP/Scala/C++ keyword constructors

aimasteracc:
Closes #531 (P2, QC-sweep)

## Scope (lead-decided)

Unambiguous keyword constructors only: Python `__init__` (class-scoped — module-level `def __init__()` stays False), Ruby `initialize`, PHP `__construct`, Scala `def this(...)`, C++ same-name methods (destructors `~Name` stay False, ≤12-level parent walk with the OOM-lesson cap). **Deliberately skipped**: Go `New*` / Rust `fn new` (convention → a separate is_factory flag is a product decision); **Scope B reported**: Kotlin primary constructors are an extraction gap (`primary_constructor` node not in the dispatch map) — follow-up material, not a flag bug.

## Verification

- 13 new live-parse tests (exact `is True`/`is False` pins incl. negative cases); 89+192+102 plugin suites green; 2151 formatter tests green
- Lead independently verified Python: class `__init__`→True, sibling method→False, module-level `__init__`→False
- Goldens unchanged (they track names/counts, not the flag) — verified, no regeneration needed
- Zero re-pins required (no existing test pinned the False states); ratchet exit 0

PR #598, opened 2026-06-12
fix(cpp): qualified out-of-scope definitions keep their qualifier

### LINKED ISSUE #590: C++: out-of-namespace qualified definitions lose their qualifier (math::max → max)
Split from #538 scope-B (the 'namespace not captured' half was a false positive — namespace → Package already works; this is the remaining half).

**Symptom** (sweep N3): `int math::max(...)` defined outside the namespace block extracts as bare `max` — the `math::` qualifier is dropped, so the function can't be associated with its namespace.

**Where**: `tree_sitter_analyzer/languages/cpp_plugin.py` — `qualified_identifier` declarator path; same family as the two-path constructor handling from #567.

**Acceptance**: qualified definitions carry the qualifier (full_qualified_name or receiver_type = `math`), consistent with how out-of-class methods resolve; goldens re-pinned exactly.

aimasteracc:
Closes #590 (split from #538 scope-B, sweep N3).

## Summary
Diagnosis nuance (live-parsed): the qualifier wasn't literally dropped — `math::max` kept it in the NAME while in-namespace `min` lost the association entirely, so two definition styles of the same logical symbol diverged. Fix converges all forms on the house convention (names stay bare, owner travels in `receiver_type` — #429/#474/#535):

| Case | Before | After |
|---|---|---|
| in-namespace fn | `min`, recv=None | `min`, recv=`math` |
| out-of-ns `math::max` | `math::max`, recv=None | `max`, recv=`math` |
| out-of-class `math::Foo::bar` | `math::Foo::bar`, recv=None | `bar`, recv=`math::Foo` |
| out-of-class ctor `math::Foo::Foo` | is_constructor=false | true (#567 family) |

Conversion operators (`operator std::string`) never split; anonymous namespaces skipped; C++17 `namespace a::b` supported; parent walk depth-capped. Full-table surface displays `math::max` (the sweep's observation point). In-class members keep recv=None (owner travels via class nesting — full Rust-parity receivers would be a larger follow-up).

## Test plan
- [x] RED-first 11 failed → GREEN 16; full relevant sweep 527 passed + agent contracts 53 (-n 0)
- [x] Goldens both halves: cpp.json zero diff (bare-name snapshots); full table 1 line `| max |`→`| math::max |`; csv/toon byte-identical after newline-artifact restore
- [x] Ratchet exit=0 (rebased over #592); patch lines covered; change-impact risk=none

Lead independently spot-checked live: all six cases above reproduce exactly.

### DIFF
--- a/tests/golden_masters/full/cpp_sample_full.md
+++ b/tests/golden_masters/full/cpp_sample_full.md
@@ -108,7 +108,7 @@ using StringList = vector<string>;
 ## Global Functions
 | Method | Signature | Vis | Lines | Cols | Cx | Doc |
 |--------|-----------|-----|-------|------|----|----|
-| max | (a:T, b:T):T | + | 25-27 | 5-6 | 2 | - |
+| math::max | (a:T, b:T):T | + | 25-27 | 5-6 | 2 | - |
 | operator<< | (os:ostream&, c:const Circle&):ostream& | + | 160-163 | 5-6 | 1 | - |
 | swap_values | (a:T&, b:T&):void | + | 176-180 | 5-6 | 1 | - |
 | add | (a:int, 0:int b =):int | + | 185-187 | 5-6 | 1 | - |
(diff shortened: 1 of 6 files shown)


aimasteracc:
@codex review

PR #601, opened 2026-06-12
docs(rfc): RFC-0016 — semantic symbol search via sqlite-vec (draft)

aimasteracc:
Tracking issue: #517 (dogfood evidence: #441 residual — conceptual queries dead-end on BM25, entry_points empty).

## Summary
Draft RFC for opt-in semantic symbol search: sqlite-vec virtual table in the same `.ast-cache` SQLite (no new daemon, 100% local moat intact), `search action=semantic` + `--search-semantic` CLI twin, optional `nav context` rerank.

Key design points for review:
- Embedding input = signature + docstring; graceful-degrade verdicts for extension/index/model absence (#537 honesty convention)
- `[semantic]` pip extra, default install stays lean; ONNX MiniLM default vs ollama endpoint = **Open Q1 (panel)**
- Phase-1 brute-force cosine option (Alt C) to decouple API from sqlite-vec platform risk = **Open Q2**
- Rule-11 cost invariants pinned in the test plan (index bytes/symbol, build throughput)
- Hybrid lexical tiebreak so exact-name hits never lose to fuzzy (Sourcegraph lesson)

Status: **draft** — seeking adversarial review on the open questions before any implementation lands.

Issue #517, opened 2026-06-11
[Feature][dogfood] Semantic symbol search via sqlite-vec in .ast-cache — evidence from #441 residual recall gap


aimasteracc:
## Feature request (user perspective): semantic / conceptual symbol search
As an agent user, BM25-only search fails me on **conceptual** queries — the exact gap measured in this dogfood cycle:
- `nav action=context "how does the MCP server route a tool call to the right facade action"` → after the #487 stop-word fix, candidates are honest but **entry_points is empty** (#441 residual): the relevant symbols (`legacy_to_facade`, `handle_call_tool`) exist but no lexical token reaches them.
- `search action=symbol` requires knowing the name fragment; "where is request dispatching handled" finds nothing.

Every "how does X work" question that doesn't share tokens with the symbol names dead-ends, and the agent falls back to grep loops — the exact behavior TSA exists to replace.

## Proposal sketch (keeps the locked architecture)
- Embed symbol signatures+docstrings into a vector index stored **in the same `.ast-cache/` SQLite file** via the `sqlite-vec` extension + HNSW — no new daemon, no engine change, "100% local" story intact.
- Surface as `search action=semantic` (and a rerank option for `nav context` candidate recall).
- Needs an RFC (public API addition) — filing as the user-demand evidence for it.

Related context: HelixDB/market scan shows semantic search is table-stakes for "AI memory/knowledge" positioning; TSA's wedge is correct-edges + local-zero-dep, which sqlite-vec preserves.

aimasteracc:
RFC-0016 (draft, revision 2) merged via #603 — rfcs/0016-semantic-symbol-search.md. Two-reviewer adversarial panel reshaped the design: phase-0 = FTS5 porter stemming (#604, ships first), phase-1 = brute-force-over-BLOB (measured 1.03ms @ 44k symbols, no sqlite-vec/extension surface), retrieval-quality pilot is the acceptance gate before any implementation. Issue stays open tracking the pilot + implementation.

aimasteracc:
Phase-0 re-baseline (post-#606 porter stemming, measured on this repo's real index):

- `dispatching` recall: **0 → 39 hits (16 production)** — the stemming half of the motivation is closed.
- Exact-token queries: production sets identical (16/16, same order) — no regression.
[...]

Issue #441, opened 2026-06-10
[P2][dogfood] nav context: stop-words as candidates, examples/ dominate entry_points, misleading next_step


aimasteracc:
## Summary
`nav action=context` (the documented START-HERE call) degrades badly on natural-language tasks: stop-words become search candidates, demo/example files dominate the result, and `next_step` confidently tells the agent to answer from the wrong code.

## Repro (develop @ 32412d6e)
```
nav action=context query="how does the MCP server route a tool call to the right facade action"
```
Returns:
- `candidates: [MCP, server, tool, right, facade, action]` — raw token split; "right" is used as a search term.
- `entry_points`: 4 of 6 are `examples/file_output_factory_demo.py` and `scripts/sync_version_minimal.py`; `code_blocks` spends ~4KB inlining the demo script's source.
- The actually-relevant symbols (`facade_map.legacy_to_facade`, `server.handle_call_tool`) only appear buried in `related_symbols` with no source.
- `agent_summary.next_step`: "Answer from code_blocks now." — an agent obeying this answers the routing question from a file-output demo.

## Impact
This is the tool the server instructions tell agents to call FIRST for every how-does-X-work question. On multi-word natural tasks it actively misleads, and the per-block source inlining makes the wrong answer expensive too.

## Suggested fix
1. Stop-word filter + light keyword scoring for candidate extraction (drop "right", "how", "does"...).
2. Rank production code above `examples/` / `scripts/` / `tests/` in entry_points and code_blocks (the test-de-prioritisation rule already exists for search ranking — apply it here).
3. Make `next_step` honest when match confidence is low: suggest a refined query or `search action=symbol`, instead of "answer from code_blocks now".

aimasteracc:
Fixed by #487 (merged)。三项修复:① stop-words 过滤(显式小集合)——且 Codex P1 加固:只过滤「未引用且全小写的纯散文 token」,`Request`/Server 等显式符号引用一律保留;② entry_points 生产优先排序(共享 test_gap_analyzer._NON_PROD_DIRS 单一事实源,#479 同源)——Codex P2 加固:wants_tests 查询不被 non-prod tier 降级;③ next_step 诚实化:仅非生产匹配时点名警告并建议 search action=symbol,有生产匹配时锚定真实符号名。candidates 实测从 6 个噪声词降到 2 个真符号;codegraph_context_tool 覆盖 96%。

aimasteracc:
[...]

PR #603, opened 2026-06-12
docs(rfc): RFC-0016 — semantic symbol search via sqlite-vec (draft)

aimasteracc:
Tracking issue: #517 (dogfood evidence: #441 residual — conceptual queries dead-end on BM25, entry_points empty). Reopens #601 under a GitFlow-compatible branch name.

## Summary
Draft RFC for opt-in semantic symbol search: sqlite-vec virtual table in the same `.ast-cache` SQLite (no new daemon, 100% local moat intact), `search action=semantic` + `--search-semantic` CLI twin, optional `nav context` rerank.

Key design points for review:
- Embedding input = signature + docstring; graceful-degrade verdicts for extension/index/model absence (#537 honesty convention)
- `[semantic]` pip extra, default install stays lean; ONNX MiniLM default vs ollama endpoint = **Open Q1 (panel)**
- Phase-1 brute-force cosine option (Alt C) to decouple API from sqlite-vec platform risk = **Open Q2**
- Rule-11 cost invariants pinned in the test plan (index bytes/symbol, build throughput)
- Hybrid lexical tiebreak so exact-name hits never lose to fuzzy (Sourcegraph lesson)

Status: **draft** — seeking adversarial review on the open questions before any implementation lands.

### DIFF
--- a/rfcs/0016-semantic-symbol-search.md
+++ b/rfcs/0016-semantic-symbol-search.md
@@ -318,7 +318,10 @@ versions; its prior-art value is SQL ergonomics, not ANN.
 
 ## What this RFC does NOT do (deferred)
 
-- sqlite-vec/ANN indexing (phase-2, behind a measured trigger).
+- sqlite-vec/ANN indexing (phase-2, behind a measured trigger; if adopted,
+  the vec0 column MUST pin `distance_metric=cosine` — the default is L2,
+  which would silently disagree with the phase-1 cosine semantics (Codex P2
+  on #603 rev 1)).
 - `nav action=context` semantic rerank (phase-2, pilot-gated).
 - Body-content embeddings; cross-repo federation; replacing BM25 anywhere;
   auto-building on first connect. (Unchanged.)
--- a/rfcs/0016-semantic-symbol-search.md
+++ b/rfcs/0016-semantic-symbol-search.md
@@ -1,0 +1,236 @@
+# RFC-0016: Semantic symbol search via sqlite-vec in `.ast-cache`
+
+- **Status**: draft
+- **Author(s)**: lead (autonomous), on behalf of dogfood evidence in #517
+- **Created**: 2026-06-13
+- **Last updated**: 2026-06-13
+- **Tracking issue**: #517
+- **Affected source paths** (pin them — reviewers watch for drift here):
+  - `tree_sitter_analyzer/mcp/tools/` (search facade: new `semantic` action)
+  - `tree_sitter_analyzer/core/ast_cache*` (new `symbol_embeddings` virtual table, schema version bump)
+  - `tree_sitter_analyzer/cli_main.py` (new `--search-semantic` flag, CLI twin)
+  - `tests/unit/mcp/`, `tests/integration/` (RED-first suites below)
+
+## Summary
+
+Add an opt-in semantic (vector) symbol search to the existing `.ast-cache`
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

aimasteracc:
## Adversarial panel round 1 — triage summary (revision 2 pushed)

Two independent reviewers (platform/ops + product/cost), all findings probe-backed. Full reports in session log; key verdicts:

| Finding | Verdict | Action in rev 2 |
|---|---|---|
| **P1** Motivation doesn't reproduce — query 1 returns `legacy_to_facade` rank #1 on HEAD | confirmed (reviewer ran the tool live) | Motivation rewritten: real gap = confidently-wrong results + stemming, not emptiness; baseline becomes an executable test |
| **P1** Embedding input fields absent from cache (0/1885 rows carry docstring) | confirmed (sqlite probe) | serialization extension now an explicit prerequisite + full-reindex cost stated |
| **P1** Brute-force ~119× overestimated; measured 1.03ms @ 44k symbols | confirmed (numpy bench) | phase-1 = BLOB brute-force; sqlite-vec demoted to measured-need phase-2 (vec0 is also linear scan — 2.38ms @ 20k) |
| **P1/P2** rowid FK dangles after re-index (both reviewers independently) | confirmed (`_ast_cache_write.py:26`) | schema rekeyed `(file_path, symbol_key)` + same-transaction file-scoped delete (ast_symbol_activation precedent) |
| P2 `enable_load_extension` absent on python.org/system macOS builds | confirmed (probe) | absence-mode table row added; phase-1 needs no extension at all |
| P2 onnxruntime matrix vs py3.10 support; no Intel-mac wheel | confirmed (PyPI probe) | pinned support table = acceptance deliverable |
| P2 "~100ms warmup" falsified (606ms measured) | confirmed | measured number in text |
| P2 model+tokenizer distribution unspecified | confirmed | Open Q1/Q2 sharpened (model wheel vs download; tokenizers wheel vs vendored WordPiece) |
| P2 TOON example violated locked control surface | confirmed | example replaced with rule statement (RFC-0012) |
| P2 hybrid tiebreak unimplementable | confirmed | respecified: FTS5 union leg + total order |
| P2 writer contention / silent drop | confirmed | ≤200-row batches, inference outside transactions, checkpoint policy |
| P2 33% substantive docstrings; no retrieval pilot | confirmed | pilot is now an acceptance PRECONDITION with explicit no-go path |
| P3 FTS5 porter stemmer ignored in alternatives | confirmed | adopted as phase 0, ships separately first |
| P3 path pins drifted; int8 phrasing; embedding_meta no PK | confirmed | all fixed |

**Surviving attacks** (explicitly): sqlite-vec wheel works end-to-end locally; size math correct; old TSA versions read a new cache without crashing; zero-daemon posture and TOON locks unviolated; the conceptual-recall gap itself is real (query 2 returns garbage).

Status stays **draft**: pilot (≈1 day) is the gate before any implementation. Codex review still welcome on rev 2.

aimasteracc:
Codex rev-1 comments triage: (1) embedding cleanup before symbol-row replacement — **already absorbed in revision 2** (schema rekeyed (file_path, symbol_key) + same-transaction file-scoped delete; both panel reviewers hit the same finding). (2) vec0 distance_metric=cosine — vec0 left phase-1 entirely in rev 2; requirement now pinned in the phase-2 deferral note so it can't be lost.

PR #608, opened 2026-06-12
test: ratchet batch-17 — pin 41 loose assertions in top-4 AST-ranked files

aimasteracc:
## Summary
Loose-assertion ratchet batch 17. Top-4 files by AST counter:

| File | Pins |
|---|---|
| tests/unit/core/test_query_loader.py | 11 |
| tests/unit/test_unreachable_code.py | 10 (+1 `<=` bound) |
| tests/unit/test_incremental_sync.py | 10 (+1 `<=` bound) |
| tests/integration/core/test_analyze_scale_tool_file_output.py | 10 |

Baseline: **933 → 892** (= exactly the 41 flagged pins; the 2 `<=` bounds are outside the AST counter and noted in the history line for auditability — lead re-verified live).

Contract 5b highlights: **real dead-branch mock fixes before pinning** — analyze_scale mocks lacked `element_type` so summary counts were silently 0 and the hotspots guard was dead (now pinned to live values incl. full hotspot-entry); `MagicMock(name=...)` repr-name gotcha fixed via helper; 8 tree-sitter Node mocks got explicit `parent=None`; byte-size pins guarded with `newline="\n"`; query counts pinned as count+min exact facts per language; zero exemptions.

## Test plan
- [x] All 4 files individually `-n 0`: 21/34/19/15 passed
- [x] Diff gate exit=0
- [x] `--baseline` measures 892 == recorded (lead independently re-verified)

### DIFF
--- a/scripts/loose_assertion_baseline.txt
+++ b/scripts/loose_assertion_baseline.txt
@@ -33,7 +33,9 @@
 #   files).
 #   batch-16 (AST): 977 -> 933, 2026-06-13 (44 pins in top-4 AST-ranked
 #   files).
+#   batch-17 (AST): 933 -> 892, 2026-06-13 (41 flagged pins in top-4
+#   AST-ranked files, plus 2 unflagged <= bounds pinned exact).
 
-BASELINE_COUNT=933
+BASELINE_COUNT=892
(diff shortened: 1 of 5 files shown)


aimasteracc:
@codex review

PR #612, opened 2026-06-12
feat(index): module-level constants indexed as symbols

aimasteracc:
Closes #610 (scope study + approved decision in the issue thread; found via RFC-0016 pilot Q3).

## Summary
Python module-level assignments now index as **kind=constant** when the target matches `^_?[A-Z][A-Z0-9_]+$` ∪ annotated-with-value ∪ `__dunder__`. `_STOP_WORDS`, `TOON_CONTROL_SURFACE`, `CANONICAL_VERDICTS`, `LEGACY_TOOL_MAP` — all previously unfindable — now searchable AND go-to-definition-able.

- `_walk_for_symbols` gains a top-down `enclosed` scope parameter (no parent walks — zero Mock-OOM surface); reads the `left` field (tree-sitter-python has no `name` field on assignment)
- `_AST_CACHE_EXTRACTOR_VERSION` 2→3 (both sites) — existing caches re-extract cleanly
- kind=constant lights up the pre-existing dead ranking weight `_KIND_BONUS["constant"]`
- Definition-kind filters extended in symbol_resolver AND codegraph_query_backend (the study cited only the former; resolve() routes through the latter — caught by the RED test staying red)
- Bare annotations (`x: int` no value) excluded — declaration, not definition (pinned)
- Cost (measured, this repo): 1,916 constant rows ≈ +3.9% / study's microbenchmark ~719 KB (0.33% of index.db)

## Test plan
- [x] RED-first: 6 failed pre-impl → 13 new tests green; negative pins (lowercase/class-body/function-body/tuple/bare-annotation excluded)
- [x] Regression 248 + golden -k python 8 (zero drift) + count-pin suites 172 + ast_cache family 60 (-n 0)
- [x] E2E probe: fresh index → `search action=symbol _STOP_WORDS` returns the constant (match_tier=exact) — pasted in issue-thread report
- [x] Zero re-pins needed; ruff clean; ratchet exit=0
- [x] Lead independently re-ran 70 tests + verified both version-bump sites + push diff=0

Follow-ups (filed next): Go/Rust same-shape gap; symbols_json serialization enrichment (RFC-0016 prerequisite).

Issue #613, opened 2026-06-12
Go/Rust: module/package-level constants not indexed (same shape as #610)


aimasteracc:
Same gap as the fixed Python case (#610/PR #612): the #610 scope study live-verified Go `const`/package-level `var` and Rust `const`/`static` items produce **0 ast_symbol_rows** (study: .recon/issue-610-constants-scope-study.md, root-cause section).

**Fix shape**: mirror #612 — extend `_walk_for_symbols` for Go `const_declaration`/`var_declaration` (package scope) and Rust `const_item`/`static_item`, kind=constant; extractor version bump; definition-kind filters already accept constant after #612. Separate PRs per language per house rules.

**Note from study**: JS/TS/Java currently OVER-capture (function-local variables become rows) — that's a third, different cleanup, file separately if pursued.

aimasteracc:
Go half merged via #615 (const_spec all names; var_spec const-style pattern only; extractor v4; e2e probe green). Rust half (const_item/static_item, v5) next — issue stays open.

aimasteracc:
Both halves complete: Go via #615 (const_spec all names, var_spec const-style only, extractor v4), Rust via #618 (const_item/static_item all names incl. static mut + associated consts, function/closure/block locals excluded, anonymous const _ skipped per Codex P2, extractor v5; child_enclosed now language-gated so each grammar's scope sets can't cross-contaminate). JS/TS/Java over-capture remains a separate cleanup if pursued (noted in the original issue).

Issue #610, opened 2026-06-12
Module-level variables/constants not indexed as symbols (_STOP_WORDS unfindable)


aimasteracc:
Found during #607 diagnosis (RFC-0016 pilot Q3): `SELECT ... WHERE name='_STOP_WORDS'` → 0 rows — module-level variable/constant assignments aren't extracted into ast_symbol_rows, so agents cannot find constants by name (`search action=symbol "_STOP_WORDS"` dead-ends).

**Scope question to settle first**: which module-level bindings deserve symbol rows (ALL_CAPS constants only? any top-level assignment? annotated only?) — measure index-size impact on this repo before choosing.

**Evidence**: .recon/rfc0016-pilot-baseline.md Q3 + PR #609 root-cause report (hypothesis 3 caveat).

aimasteracc:
## Scope study complete + lead decision (study: .recon/issue-610-constants-scope-study.md)

**Root cause (live-verified)**: `_walk_for_symbols` (_ast_extraction.py:497-572) never extracts Python `assignment` nodes — not in `_VAR_DECL_LIKE`, and they carry a `left` field not `name`. Never extracted (not extracted-then-dropped). Go/Rust same; JS/TS/Java over-capture instead.

**Approved scope** (decision under standing mandate): module-level simple assignments matching `^_?[A-Z][A-Z0-9_]+$` ∪ annotated assignments ∪ `__dunder__`, as **kind=constant**.
- Cost: +3.83% rows / ~719 KB (0.33% of index.db) — measured with real-DDL microbenchmark (400.3 B/row)
- Recall probes 4/4: _STOP_WORDS, TOON_CONTROL_SURFACE, CANONICAL_VERDICTS, LEGACY_TOOL_MAP all covered
- kind=constant lights up the pre-existing dead weight `_KIND_BONUS["constant"]: 0.04` (_ast_cache_search.py:17) instead of inflating variable semantics
- Safety verified: resolver/ownership paths filter kind IN (function,method,class) — unaffected
- Include the navigate kind-filter one-liner so constants are go-to-definition-able, not just searchable

**Implementation requirements**: single-point change in `_walk_for_symbols`; MUST bump `_AST_CACHE_EXTRACTOR_VERSION` 2→3 (both sites); re-pin test_ast_extraction symbol lists + status kind breakdown; golden_masters zero-drift (plugin path). Python first; Go/Rust as same-shape follow-ups in separate PRs.

[...]

PR #615, opened 2026-06-12
feat(index): Go package-level constants indexed as symbols

aimasteracc:
Go half of #613 (Rust half follows in a separate PR — issue stays open). Mirrors the merged Python fix #612.

## Summary
`const_declaration`/`var_declaration` wrappers carry no `name` field (why the existing `_VAR_DECL_LIKE` branch never fired) — names live on `const_spec`/`var_spec` with a REPEATED name field that leaks comma tokens (filtered by type). Now:
- **const_spec → ALL names** (Go consts are constants by compiler definition — a name-style gate would only add false negatives)
- **var_spec → const-style pattern only** (`^_?[A-Z][A-Z0-9_]+$`): package vars are mutable state; lowercase vars are exactly the FTS noise the #610 study rejected
- Blank identifier `_` skipped (not referenceable); function-locals excluded via the #612 top-down `enclosed` mechanism + `_GO_SCOPE_BODY_NODES`
- `_AST_CACHE_EXTRACTOR_VERSION` 3→4 (both sites)

## Test plan
- [x] RED-first: 6 capture tests failed pre-fix → 56 passed; exclusions pinned (locals, lowercase var, blank, iota continuation captured)
- [x] Regression 186 passed; golden -k go 18 passed zero drift (-n 0)
- [x] E2E probe (tmp Go project → ASTCache → SQL): 4 constant rows, locals/lowercase = 0 rows — pasted in report
- [x] New lines 100% covered; ruff/mypy clean; ratchet exit=0
- [x] Lead independently re-ran 56 tests + verified both version-bump sites + push diff=0

### DIFF
--- a/tests/unit/test_ast_extraction.py
+++ b/tests/unit/test_ast_extraction.py
@@ -561,3 +561,130 @@ class TestPythonModuleConstants:
         # Non-simple targets (tuple unpacking) are excluded by the scope rule.
         syms = _symbols_for("X_A, Y_B = 1, 2\n", "python")
         assert [s for s in syms if s["kind"] == "constant"] == []
+
+
+# ---------------------------------------------------------------------------
+# Issue #613: Go package-level constants indexed as kind="constant"
+# ---------------------------------------------------------------------------
+
+
+_GO_CONST_SRC = """\
+package main
+
+const A = 1
+
+const (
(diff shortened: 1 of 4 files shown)


aimasteracc:
@codex review

PR #616, opened 2026-06-12
test: ratchet batch-19 — pin 36 loose assertions in top-4 AST-ranked files

aimasteracc:
## Summary
Loose-assertion ratchet batch 19. Top-4 (8-way tie at 9 violations; tie-break documented):

| File | Pins |
|---|---|
| tests/unit/languages/test_swift_plugin_enhanced.py | 9 |
| tests/unit/languages/test_php_plugin.py | 9 |
| tests/unit/languages/test_html_plugin_tags_attrs.py | 9 |
| tests/unit/grammar_coverage/test_auto_discovery.py | 8 + 1 exemption |

Baseline: **852 → 816** (lead re-verified live). One exemption marker total (`elapsed_ms > 0` — genuine wall-clock, passes the 3-question test); the report-length pin was made deterministic via splitlines (byte length embeds timing).

**Product finding while pinning**: PHP extractor returns 0 imports even with `use` statements present — pinned as `imports == []` with gap-documented comments (forces conscious re-pin when extraction lands); filed separately.

Grammar-version-dependent pins carry provenance comments (swift 0.7.2 / php 0.24.1 / html 0.23.2) — grammar bumps go RED by design.

## Test plan
- [x] All 4 files individually `-n 0`: 44/49/17/40 passed (re-run post-ruff-format)
- [x] Diff gate exit=0 (pre- and post-commit)
- [x] `--baseline` measures 816 == recorded (lead independently re-verified)

PR #618, opened 2026-06-12
feat(index): Rust const/static items indexed as symbols

aimasteracc:
Rust half of #613 — completes the issue (Go half merged via #615; mirrors Python #612).

## Summary
`const_item`/`static_item` carry a `name` field but were never in `_VAR_DECL_LIKE` (it lists Go's node names) — 0 rows. Now a dedicated branch (mirroring #615) emits **kind=constant**:
- ALL names, no pattern gate (rustc enforces `non_upper_case_globals`; language-level constants/globals)
- `static mut` included; **associated consts (impl/trait) captured** — compiler-enforced, addressable as `Type::CONST`, unlike the mutable Python class attributes #612 excludes
- Function/closure locals excluded via `_RUST_SCOPE_BODY_NODES = {function_item, closure_expression}` — deliberately NOT `block` (shared cross-grammar; would break #612's if/try-wrapped module-assignment guarantee)
- mod-nested captured (consistent with #596 mod containers)
- `_AST_CACHE_EXTRACTOR_VERSION` 4→5 (both sites)

## Test plan
- [x] RED-first: 6 failed pre-fix → 8 new tests green; regression battery 212 passed; golden -k rust zero drift (-n 0)
- [x] E2E probe (tmp Rust project → ASTCache → SQL): 5 constant rows (incl. static mut, mod-nested, associated), locals 0 — pasted in report
- [x] New lines 100% covered; ruff/mypy clean; ratchet exit=0 (rebased over #616 mid-task)
- [x] Lead independently re-ran 64 tests + verified both version-bump sites + push diff=0

### DIFF
--- a/tests/unit/test_ast_extraction.py
+++ b/tests/unit/test_ast_extraction.py
@@ -688,3 +688,127 @@ class TestGoPackageConstants:
         syms = _symbols_for(src, "go")
         consts = sorted(s["name"] for s in syms if s["kind"] == "constant")
         assert consts == ["KindA", "KindB", "Typed"]
+
+
+# ---------------------------------------------------------------------------
+# Issue #613: Rust const/static items indexed as kind="constant"
+# ---------------------------------------------------------------------------
+
+
+_RUST_CONST_SRC = """\
+const MAX: u32 = 10;
+
+static GLOBAL: &str = "x";
+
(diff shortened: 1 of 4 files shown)


aimasteracc:
@codex review

PR #619, opened 2026-06-12
fix(php): use statements extracted as imports

### LINKED ISSUE #617: PHP: use statements never extracted as imports (extractor returns 0 imports)
Found by ratchet batch-19 while pinning tests/unit/languages/test_php_plugin.py: the PHP extractor returns **0 imports** even for fixtures containing `use` statements — the old `>= 0` asserts were vacuously green over this gap for their entire life (the exact false-green failure mode the exact-assertion rule exists to catch).

The batch-19 pins codify current behavior as `imports == []` with gap comments — when extraction lands, those pins go RED and force conscious re-pins (intended).

**Fix shape**: PHP plugin import extraction for `namespace_use_declaration` (verify node type with a live parse first); mirror how other languages populate imports; exact-pin tests; check golden/plugin JSON surfaces for re-pin needs. Also noted in the same file: `MAX_USERS` const not in variables (PHP half of the #610/#613 constants family).

aimasteracc:
Closes #617 (found by ratchet batch-19's gap pins — the old `>= 0` asserts were vacuously green over this for their entire life).

## Root cause
`_build_use_clause_import` did `child_by_field_name("name")` — that field never exists on `namespace_use_clause` in tree-sitter-php 0.24, so PHP imports were 0 forever. Group-use clauses (nested in `namespace_use_group`) were additionally never reached.

## Fix (php_helpers.py, surgical)
- Name lookup falls back to first `qualified_name`/`name` child (alias `name` always follows `as` — document order safe); group use expands prefix + one Import per member; alias convention matches Python (`import x as y`: FQN in name, binding in alias)
- **Trait `use` inside class bodies is NOT an import** — composition, different node type (`use_declaration`), pinned by test

## Test plan
- [x] RED-first: 6 failed pre-fix → 51 passed; batch-19's gap pins consciously re-pinned (old→new table in report)
- [x] Goldens BOTH halves (5-kind): php.json +Import:2 / full +Imports section / toon import_count 0→2 / compact+csv byte-identical (don't render imports) — every line traced
- [x] Ratchet exit=0 (rebased over #616); ruff clean
- [x] Lead independently re-ran 51 tests + live probe (simple/group/alias all extract, alias=O)

Noted not fixed: `MAX_USERS` class const = PHP half of the constants family (#610/#612/#613).

### DIFF
--- a/tests/golden_masters/full/php_sample_full.md
+++ b/tests/golden_masters/full/php_sample_full.md
@@ -1,5 +1,11 @@
 # Sample.php
 
+## Imports
+```php
+import App\Contracts\UserInterface
+import App\Traits\Timestampable
+```
+
 ## Classes Overview
 | Class | Type | Visibility | Lines | Methods | Fields |
 |-------|------|------------|-------|---------|--------|
--- a/tests/golden_masters/plugins/php.json
+++ b/tests/golden_masters/plugins/php.json
@@ -2,9 +2,10 @@
   "counts_by_type": {
     "Class": 5,
     "Function": 21,
+    "Import": 2,
     "Variable": 9
   },
-  "element_total": 35,
+  "element_total": 37,
   "names_by_type": {
     "Class": [
       "App\\Models\\AdminUser",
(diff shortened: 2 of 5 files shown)


aimasteracc:
@codex review

PR #625, opened 2026-06-12
feat(index): PHP constants indexed as symbols

### LINKED ISSUE #624: PHP: const declarations not extracted (class consts missing from variables; module-level const/define not indexed)
PHP `const` declarations are invisible on BOTH surfaces, completing the constants family (#610/#612 Python, #613/#615 Go, #618 Rust) and confirming the MAX_USERS gap noted on #619. **(A) ast_symbol_rows**: tree-sitter-php puts const names on `const_element` children of `const_declaration`; `const_declaration` sits in `_VAR_DECL_LIKE` but carries no `name` field, so the name gate never matches and no `kind="constant"` rows are produced — neither top-level `const MAX = 1;` nor class consts (`class C { const MAX_USERS = 1000; }`). **(B) plugin extract path**: `extract_variables` in php_plugin already routes `const_declaration` → `_build_php_constant_variable` (php_helpers.py), but that helper does `element_node.child_by_field_name("name")` — live-parse proves `const_element` has NO `name` field (name is the first child of type `name`) — so every const is silently dropped (verified: `class Config { const MAX_USERS = 1000; public const X = 2; }` yields only the $prop property, zero consts). `define()` calls are `function_call_expression` nodes (runtime registration, not declarations) and should stay out of scope. Plan mirrors #615/#618: ALL top-level const names captured (compiler-enforced immutable, like Go consts); class/interface/trait/enum consts captured (addressable as `C::MAX_USERS`, like Rust associated consts); function-body consts excluded via the enclosed gate (illegal PHP that tree-sitter still parses); extractor version bump 6→7.

aimasteracc:
Closes #624 — PHP half of the constants family (#612 Python / #615 Go / #618 Rust), plus the plugin-surface MAX_USERS gap noted in #619.

## Root cause (both surfaces, same family as #617/#619)
`const_element` has NO `name` field in tree-sitter-php — the walker's name-field gate made `const_declaration` (already in _VAR_DECL_LIKE) permanently invisible, and the plugin's `_build_php_constant_variable` silently dropped every class const the same way.

## Scope (mirrors the family precedents, each pinned)
- Top-level const: ALL names (compiler-enforced, Go reasoning)
- Class/interface/trait/enum consts: captured (`Config::MAX_USERS`, Rust associated-const reasoning)
- Function-body consts excluded via `_PHP_SCOPE_BODY_NODES` (keyed on function node types so braced-namespace consts stay captured); `define()` skipped (runtime call, dynamic names); multi-element declarations → one row each
- `_AST_CACHE_EXTRACTOR_VERSION` 6→7 (both sites); plugin surface fixed in 6 lines (field-then-child-scan, #617 pattern)

## Test plan
- [x] RED-first: 7+2 failed pre-fix → all green; regression 258 + edge_store 22 + golden -k php (-n 0)
- [x] Conscious re-pins: class-const names 4→5, variables 4→5, version 6→7; golden php.json Variable 9→12 (+exactly the 3 Sample.php consts) — every line explained
- [x] E2E probe: 5 constant rows, define()/fn-body leakage = 0 — pasted
- [x] Ratchet exit=0; ruff/mypy clean
- [x] Lead independently re-ran 134 tests + verified both version sites + push diff=0

### DIFF
--- a/tests/golden_masters/compact/php_sample_compact.md
+++ b/tests/golden_masters/compact/php_sample_compact.md
@@ -5,7 +5,7 @@
 |----------|-------|
 | Namespace |  |
 | Methods | 21 |
-| Fields | 9 |
+| Fields | 12 |
 
 ## Methods
 | Method | Sig | V | L | Cx | Doc |
--- a/tests/golden_masters/csv/php_sample_csv.csv
+++ b/tests/golden_masters/csv/php_sample_csv.csv
@@ -1,4 +1,7 @@
 Type,Name,Signature,Visibility,Lines,Complexity,Doc
+Field,User::STATUS_ACTIVE,User::STATUS_ACTIVE:const,public,20-20,,-
+Field,User::STATUS_INACTIVE,User::STATUS_INACTIVE:const,public,21-21,,-
+Field,User::MAX_LOGIN_ATTEMPTS,User::MAX_LOGIN_ATTEMPTS:const,private,22-22,,-
 Field,User::id,User::id:int,private,25-25,,-
 Field,User::username,User::username:string,public,26-26,,-
 Field,User::email,User::email:string,public,27-27,,-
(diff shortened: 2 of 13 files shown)


aimasteracc:
@codex review

Issue #628, opened 2026-06-12
C#: function-local variables over-captured (same as #626 family)


aimasteracc:
The #626 study measured 10 C# variable rows from the same ungated `_VAR_DECL_LIKE` path. Same contraction shape as the approved #626 rule once JS/TS (v8) and Java (v9) land; C# scope-body set needs its own live-parse (method_declaration, local_function_statement, lambda_expression, ...). Evidence: .recon/issue-626-overcapture-scope-study.md.

aimasteracc:
Fixed by #633 — overcapture family now closed across all four languages: JS/TS #629 (v8), Java #630 (v9), C# #633 (v10). C# divergence: block deliberately absent from the scope set (C#9 top-level statements live in compilation-unit blocks — function-keying verified complete across 27 live-probed shapes); accessor bodies gated by accessor_declaration; top-level statement vars kept as module scope; ERROR regions hardened. No shadow-semantics impact (only JS/TS consume kind=variable shadow sets).

Issue #626, opened 2026-06-12
JS/TS/Java: function-local variables over-captured into ast_symbol_rows (inverse of the constants family)


aimasteracc:
The #610 scope study (root-cause section) verified: while Python/Go/Rust/PHP under-captured (fixed in #612/#615/#618/#625), **JS/TS/Java over-capture** — their declaration node types reach `_VAR_DECL_LIKE` without scope gating, so function-local variables become symbol rows (~397 variable rows on this repo at study time). Local `const x = ...` inside a JS function is search noise with no definition-site value.

**Fix shape**: apply the same language-gated `enclosed` scope mechanism (#612's top-down flag + per-language scope-body sets, now with 4 precedents). **This is a behavior CONTRACTION** — needs a conscious scope decision first (module/class-level keep, function-local drop? what about TS module-namespace bodies?), measured row-count delta, extractor version bump, and a golden/test re-pin sweep. Recommend a short scope study (mirror .recon/issue-610-constants-scope-study.md) before implementation.

Evidence: .recon/issue-610-constants-scope-study.md; #613 closing note.

aimasteracc:
## Scope study complete + lead decision (.recon/issue-626-overcapture-scope-study.md)

**Approved contraction** (standing mandate): language-gated `enclosed` skip in the `_VAR_DECL_LIKE` branch for js/ts/java. Keep module/top-level declarators (const+let+var, NO const-style gate — this is contraction, not a constants feature), TS namespace-level, Java fields/interface constants; drop all function/method/ctor/lambda/initializer locals. Scope sets as proposed (JS/TS shared incl. class_static_block, statement_block deliberately absent preserving the #612 if/try guarantee; Java incl. block for instance initializers).

Measured: −240 rows on this repo (−60.9% of family variable rows); JS −56.9% / TS −53.8% / Java −69.3% on the in-repo corpus.

**Per-finding verdicts**:
1. builtin-shadow narrowing (synapse js:108/ts:125): **approved as a precision gain** — the current file-INsensitive suppression (a local `const Math` anywhere kills builtin classification project-wide) is coarser than what we ship after; the fix PR must state this verdict and pin the new module-scope-only behavior with tests.
2. Kind re-classification (module const / static final → constant): **SEPARATE change**, agreed — schema + shadow-query implications need their own round.
3. Two PRs, two bumps: JS+TS (v8) then Java (v9), serial (same file).
4. C# same disease (10 rows): file as follow-up.

[...]

PR #629, opened 2026-06-12
fix(index): JS/TS function-local variables no longer over-captured

aimasteracc:
JS/TS half of #626 (approved contraction; Java = next PR with v9; C# = #628).

## Summary
JS/TS `variable_declarator` rows were emitted ungated — function-local variables became searchable symbols (~60% of the family's 397 rows were noise). Now the constants-family `enclosed` mechanism gates them: new shared `_JSTS_SCOPE_BODY_NODES` (8 node types incl. class_static_block; `statement_block` deliberately absent — if/try/catch-wrapped module declarators stay captured, pinned per the #612 guarantee). Module/top-level const+let+var and TS namespace-level declarators kept (no name gate — contraction, not a constants feature). `_AST_CACHE_EXTRACTOR_VERSION` 7→8.

**Language-id completeness verified**: cache path only delivers javascript/typescript (.mjs/.cjs/.mts/.cts aren't in EXT_TO_LANG at all — never indexed); gate is complete.

**Builtin-shadow narrowing (#346 semantic change — approved as precision gain)**: the synapse shadow gates consume kind=variable rows; previously a function-local `const Math` ANYWHERE suppressed builtin classification project-wide (file-insensitive). After this PR only module-scope shadows suppress. Pinned through the real cache: module `const Math` still suppresses; function-local one no longer does (exact pins, JS + TS).

Known caveat (documented, not pinned): .tsx parses with the typescript grammar; JSX ERROR-recovery can flatten function bodies so locals in broken regions may survive — pre-existing grammar behavior, outside #626.

## Test plan
- [x] RED-first: 11 failed pre-fix → consolidated narrow run 486 passed (-n 0)
- [x] Golden FULL suites: 96 passed, 1 fail = pre-existing swift local-env (proven by stash-rerun on clean develop) — zero drift
- [x] E2E probe: 7 module/namespace rows kept, local rows [] — pasted; extractor_version 8
- [x] Re-pin: only the version test (7→8); ratchet exit=0; ruff clean
- [x] Lead independently re-ran 104 tests + verified both version sites + push diff=0

### DIFF
--- a/tests/unit/test_ast_extraction.py
+++ b/tests/unit/test_ast_extraction.py
@@ -1152,3 +1152,18 @@ class TestJsTsLocalVariableContraction:
         # locals must still be captured by this change.
         src = "class A { void m() { int local = 1; } }\n"
         assert self._variables(src, "java") == ["local"]
+
+
+class TestJsTsErrorRegionHardening:
+    """Codex P2 on #629: declarations inside error-recovered regions have
+    undecidable scope — they must not be emitted (better unindexed than a
+    function-local masquerading as a module symbol)."""
+
+    def test_declaration_inside_error_node_not_emitted(self):
+        # Deliberately broken TSX-ish source that forces ERROR recovery at
+        # the point of the declaration.
+        src = "const Comp = () => {\n  const Math = 1;\n  return <div//;\n};\nconst TOP = 2;\n"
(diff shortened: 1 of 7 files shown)


aimasteracc:
@codex review

PR #630, opened 2026-06-12
fix(index): Java method-local variables no longer over-captured

aimasteracc:
Java half of #626 — completes the approved contraction (JS/TS merged as #629; C# tracked as #628).

## Summary
`_JAVA_SCOPE_BODY_NODES = {method_declaration, constructor_declaration, compact_constructor_declaration, lambda_expression, static_initializer, block, ERROR}` gates the `_VAR_DECL_LIKE` branch for java. Live-parse verified: instance initializers are bare `block`s under class_body while `field_declaration` rows never route through any block (so `block` is safe); constructor bodies are `constructor_body` nodes — `constructor_declaration` is what gates ctor locals. ERROR-region hardening mirrors #629 (live-probed: a shattered lambda-in-field WAS leaking a local row).

- KEEP: class fields, static finals, interface constants, record fields (e2e probe: exactly 4 rows, locals [])
- DROP: method/ctor/compact-ctor/lambda/static-init/instance-init locals (−69.3% on the in-repo corpus per the study)
- `_AST_CACHE_EXTRACTOR_VERSION` 8→9 (both sites)
- **No shadow-semantics change**: Java has no builtin-shadow mechanism (grep-verified — only JS/TS consume kind=variable shadow sets)
- C# explicitly NOT widened (pinned: csharp locals still captured — #628's scope decision)

## Test plan
- [x] RED-first: 4 new tests failed pre-fix → 218 + 103 regression passed (-n 0)
- [x] Golden FULL suites: 96 passed, swift fail = pre-existing (stash-proven) — zero drift
- [x] Conscious re-pins: version 8→9; JS/TS-era `java_locals_unaffected` flipped to `now_contracted_too`
- [x] Ratchet exit=0; ruff clean
- [x] Lead independently re-ran 110 tests + verified both version sites + push diff=0

### DIFF
--- a/tests/unit/test_ast_extraction.py
+++ b/tests/unit/test_ast_extraction.py
@@ -1147,11 +1147,12 @@ class TestJsTsLocalVariableContraction:
         src = "const TOP = 1;\nfunction Comp() { const inner = 2; }\n"
         assert self._variables(src, "typescript") == ["TOP"]
 
-    def test_java_locals_unaffected(self):
-        # Java is the separate follow-up PR (#626 part 2) — its function
-        # locals must still be captured by this change.
+    def test_java_locals_now_contracted_too(self):
+        # Re-pinned when the Java half of #626 landed (extractor v9): Java
+        # method locals are dropped as well — full Java surface pinned in
+        # TestJavaLocalVariableContraction below.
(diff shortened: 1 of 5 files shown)


aimasteracc:
@codex review

PR #633, opened 2026-06-13
fix(index): C# method-local variables no longer over-captured

### LINKED ISSUE #628: C#: function-local variables over-captured (same as #626 family)

aimasteracc:
Closes #628 — completes the over-capture family (JS/TS #629, Java #630).

## Summary
C# scope set gates the `_VAR_DECL_LIKE` branch: `{method, constructor, destructor, operator, conversion_operator declarations, local_function_statement, accessor_declaration, lambda_expression, anonymous_method_expression, ERROR}`. 27 declaration shapes live-probed.

**The review-worthy decision — `block` deliberately ABSENT (diverges from Java #630)**: C# 9 top-level statements put blocks at compilation-unit level (`block < if_statement < global_statement`); a block-keyed set would drop if/try-wrapped top-level declarators and break the #612 module-scope guarantee. Function-keying is complete — every probed local routes through a set node. Accessor bodies are blocks gated by `accessor_declaration`; field rows never route through any set node (verified).

- KEEP: fields/const/static-readonly/interface+record constants, top-level statement vars (the program's module scope)
- DROP: method/ctor/dtor/operator/local-fn/lambda/anon-method/accessor locals + ERROR regions (leak demonstrated live pre-fix)
- `_AST_CACHE_EXTRACTOR_VERSION` 9→10; **no shadow-semantics change** (grep-verified: only JS/TS consume kind=variable shadow sets)

## Test plan
- [x] RED-first: 5 failed pre-fix → 313 regression passed; #630's csharp pin consciously flipped (cites v10)
- [x] Golden suites: 36 passed, swift fail stash-proven pre-existing — zero drift
- [x] E2E probe: 5 kept rows (fields + top-level), method/lambda locals absent — pasted
- [x] Ratchet exit=0; ruff clean; lead independently re-ran 115 tests + verified both version sites + push diff=0

### DIFF
--- a/tests/unit/test_ast_extraction.py
+++ b/tests/unit/test_ast_extraction.py
@@ -1249,11 +1249,12 @@ class TestJavaLocalVariableContraction:
         for local in ("lambdaFieldLocal", "staticInitLocal", "instanceInitLocal"):
             assert local not in names
 
-    def test_csharp_locals_unaffected(self):
-        # C# has the same over-capture disease (10 live rows) but is a
-        # separate follow-up per the #626 decision — must NOT be widened here.
+    def test_csharp_locals_now_contracted_too(self):
+        # Re-pinned when the C# follow-up (#628) landed (extractor v10): C#
+        # method locals are dropped as well — full C# surface pinned in
+        # TestCSharpLocalVariableContraction below.
         src = "class A { void M() { int local = 1; } }\n"
-        assert self._variables(src, "csharp") == ["local"]
+        assert self._variables(src, "csharp") == []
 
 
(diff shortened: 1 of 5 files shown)


aimasteracc:
@codex review

PR #634, opened 2026-06-13
fix(html): self-closing tags no longer captured twice

### LINKED ISSUE #632: HTML: self-closing tags captured twice (element + self_closing_tag nodes)
Found by ratchet batch-23 while pinning tests/unit/languages/test_html_plugin_coverage_boost.py: `<br/>` and `<input ... />` each yield **2 entries** — the extractor walks both the element node and its self_closing_tag child as separate captures. The batch-23 pins codify current behavior (`== 2` entries with explanatory comments) — fixing the duplicate flips those pins consciously.

**Fix shape**: dedup at the walk level (skip self_closing_tag when its parent element was already captured, or vice versa — live-parse to pick); re-pin the batch-23 pins to `== 1`; check HTML goldens (tags counts) for re-pin needs; golden 验收 run test_golden_master_regression.py full.

aimasteracc:
Closes #632 (found by ratchet batch-23's pins).

## Root cause
tree-sitter-html wraps every self-closing tag as `element > self_closing_tag` (parent/child); `_is_html_element_node` matches BOTH types, so the walk captured the parent (whose helpers already extract tag_name/attributes by descending into the child) then recursed and captured the child again — `<br/>` yielded 2 entries.

## Fix (6 lines)
Skip `self_closing_tag` children at the recursion site — the parent capture carries the attributes. The type stays in `_is_html_element_node` so unwrapped self_closing_tags (error-recovery trees) still extract.

Void elements without slash (`<br>`, ``) live-parsed: `element > start_tag`, no self_closing_tag node, already 1 entry — pinned.

## Test plan
- [x] RED-first: 3 failed pre-fix; batch-23's `== 2` pins consciously flipped to `== 1` (comments cite #632); new exact-list + attribute-survival + void-element pins
- [x] 5 HTML test files 105 passed (-n 0); accuracy/tags_attrs needed zero re-pins (no slash-form fixtures)
- [x] Goldens: zero `/>` in html sample — no regen needed; golden FULL suites 96 passed (swift fail stash-proven pre-existing)
- [x] Ratchet exit=0; lead independently re-ran 45 tests + push diff=0

### DIFF
--- a/tests/unit/languages/test_html_plugin_coverage_boost.py
+++ b/tests/unit/languages/test_html_plugin_coverage_boost.py
@@ -356,19 +356,40 @@ class TestRealParsingPaths:
         """Lines 168-179: self_closing_tag node type"""
         tree = _parse_html("<br/>")
         elements = extractor.extract_html_elements(tree, "<br/>")
-        # Exact current behavior: the self_closing_tag is captured twice
-        # (element node + self_closing_tag node) -> 2 entries
-        assert [e.tag_name for e in elements] == ["br", "br"]
+        # #632: the element node wraps the self_closing_tag child; only the
+        # parent element is captured (the child would be a duplicate) -> 1 entry
+        assert [e.tag_name for e in elements] == ["br"]
 
     def test_input_with_boolean_attribute(self, extractor):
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #642, opened 2026-06-13
fix(toon): heterogeneous arrays no longer drop fields silently

### LINKED ISSUE #637: [P1][dogfood] TOON encoder silently drops fields absent from a heterogeneous array's first row — callers' 49 bodies vanish
Dogfood patrol 2026-06-13 (develop dbda9a3e), lead-reproduced:

```python
ToonFormatter().format({'results':[{'a':1},{'a':2,'body':'IMPORTANT'}]})
# → results:\n  [2]{a}:\n    1\n    2     — body GONE, silently
```

Real-world impact: `nav action=callers` TOON output (8.6KB) loses all 49 caller `body` fields that JSON (29KB) carries — the first row is a ghost caller without body (separate issue), so the table header is built from its keys and every later row's body is dropped. The 'token-efficient' format is silently LOSSY on heterogeneous arrays.

**Root cause area**: toon_encoder builds the tabular header from the FIRST dict's keys; toon_encoder.py:340 has a homogeneity check but the executing path doesn't route through it.

**Fix shape options** (pick with evidence): (a) header = union of all rows' keys (absent cells empty); (b) fall back to non-tabular encoding for heterogeneous arrays (homogeneity check actually enforced); (c) both with size guard. MUST add a disjoint/lossless invariant test at the handle_call_tool boundary (RFC-0012 lesson — denylist-style fixes are 假修复). Note: this is the 'fix the divergence' clause of CLAUDE.md §1 — TOON default stays locked.

aimasteracc:
Closes #637 (P1, dogfood patrol 2026-06-13; lead-reproduced).

## Root cause
`_handle_dict_key` routed ANY first-item-is-dict list straight into ENCODE_ARRAY_TABLE — the homogeneity check at toon_encoder.py:340 only runs for top-level/list-nested lists, never the universal MCP shape `{'results': [...]}`. `_handle_array_table` built the schema from row 0's keys and the row encoder only visits schema keys → every key absent from row 0 silently vanished. Worst case found while testing: `[{}, {'a':1}]` dropped EVERY field of every row.

## Fix: union-of-keys header (measured)
Schema = union of all rows' keys in first-seen order; missing cells render empty. Mixed dict/scalar lists encode inline (pre-fix: AttributeError → full-JSON fallback). Nested annotations sampled from the first row HAVING the key.

| case | TOON before | TOON after | JSON |
|---|---|---|---|
| heterogeneous (real 49-caller offender) | 2,009 B — **lossy, 0/48 bodies** | 6,559 B, all bodies (71.8% of JSON) | 9,132 B |
| homogeneous | 6,519 B | **byte-identical** | 9,049 B |

The token-efficiency story survives (−28% vs JSON, now honest); the old "win" was deletion. This is §1's "fix the divergence" clause — defaults untouched.

## Losslessness invariant (Rule 11)
New boundary test drives the real `handle_call_tool` closure with the offender shape and STRUCTURALLY asserts every key of every dict in any list at any depth appears in toon_content — no denylist/allowlist (the RFC-0012 假修复 trap).

## Test plan
- [x] RED-first 9 failed → 12/12; encoder/formatter/CLI/MCP suites 142+147+201 passed; **cost invariants hold, nothing relaxed**
- [x] Golden acceptance 78/78, zero byte changes (all goldens homogeneous)
- [x] Patch coverage 96.88% (sole miss pre-existing); ratchet exit=0 (rebased over batch-25)
- [x] Lead independently re-reproduced + re-ran invariant suites + push diff=0

Known residual (pre-existing, out of scope): nested-dict CELL values with divergent subkeys remain positionally ambiguous — filed separately.

### DIFF
--- a/tests/unit/formatters/test_toon_heterogeneous_637.py
+++ b/tests/unit/formatters/test_toon_heterogeneous_637.py
@@ -1,0 +1,94 @@
+"""Issue #637 — TOON encoder silently dropped fields absent from a
+heterogeneous array's first row.
+
+The array-table header was built from ``items[0].keys()`` only
+(``_handle_array_table`` / ``_infer_schema``), and ``_handle_dict_key``
+routed ANY first-item-is-dict list into the table path without the
+homogeneity check that ``_handle_list_start`` carries.  Result: a ghost
+first row without ``body`` erased every later caller's ``body`` — the
+"token-efficient" format was silently LOSSY.
+
+Fix shape (option a): the table schema is the UNION of all rows' keys in
+first-seen order; rows missing a key render an empty cell.  Homogeneous
+arrays are byte-identical to the pre-fix output (union of identical key
+sets == first row's keys).
+
+All pins are exact ``==`` (user-locked rule — no substring-only/loose
+assertions for deterministic encoder output).
+"""
+
+from tree_sitter_analyzer.formatters.toon_encoder import ToonEncoder
+from tree_sitter_analyzer.formatters.toon_formatter import ToonFormatter
+
+
+class TestHeterogeneousArrayLosslessness:
+    """Fields present in ANY row must appear in the TOON output."""
+
+    def test_issue_637_minimal_repro_body_survives(self) -> None:
+        """The lead-verified repro: second row's ``body`` must survive."""
+        out = ToonFormatter().format(
+            {"results": [{"a": 1}, {"a": 2, "body": "IMPORTANT"}]}
+        )
+        assert out == "results:\n  [2]{a,body}:\n    1,\n    2,IMPORTANT"
+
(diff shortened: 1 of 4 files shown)


aimasteracc:
@codex review

aimasteracc:
Codex P2 triage (absent vs empty-string indistinguishable after union): **real, accepted as a documented limitation** — TOON's tabular syntax has no absence marker; introducing a sentinel (`-`/null literal) creates new ambiguity with genuine data containing those values, and absent-vs-intentionally-empty rarely changes agent behavior for text fields. Folded into #643 (TOON residual semantics issue) so a future TOON syntax revision can address absent/empty/null three-state properly. The #637 losslessness invariant (every key + leaf value recoverable) holds; this edge is about the empty-string corner only.

Issue #452, opened 2026-06-10
[P2][dogfood] Nested functions have zero call edges; NOT_FOUND verdict for an indexed symbol


aimasteracc:
## Summary
Functions defined **nested inside another function** (closures — e.g. handlers registered via decorator inside a registration function) have **zero call edges**, and the miss is reported with a misleading verdict.

## Repro (develop @ 32412d6e)
```
nav action=callees symbol=handle_call_tool
→ verdict: NOT_FOUND, callee_count: 0, callees: []
```
`handle_call_tool` (`tree_sitter_analyzer/mcp/server_utils/tool_registration.py:59`, nested `async def` inside the registration function) visibly calls `_dispatch_tool`, `ensure_canonical_error_envelope`, `ensure_canonical_success_envelope`, `reduce_to_control_surface`, `build_agent_friendly_error`, `TextContent`, ... — none indexed. Conversely `nav action=call_path source=handle_call_tool target=apply_toon_format_to_response` *does* find and inline the symbol's body but reports `NO_PATH` with `source_endpoint.callers: []`, `callees: []`.

## Impact
1. **Coverage**: nested functions are the backbone of Python registration/decorator patterns (this repo's own MCP boundary!). Every flow question through them dead-ends — the server cannot trace its own request path.
2. **Honesty**: `NOT_FOUND` for a symbol the index demonstrably has (call_path inlined its body) misleads the agent into "symbol doesn't exist" instead of "edges missing for nested scope". Should be INFO + `callee_count: 0` + an explicit "nested function — call edges not extracted" note, until extraction is fixed.

## Suggested fix
Extract call edges for nested function definitions (attribute them to the nested symbol, not the enclosing one); fix the verdict semantics for exists-but-no-edges in the meantime.

aimasteracc:
Fixed by #484 (merged)。根因:_extract_call_edges 把 call 归给第一个行区间包含它的函数(深度优先=外层先注册永远赢)→ 嵌套函数零边却有索引符号 = 矛盾 NOT_FOUND。修复:最小 span 归属(最内层函数)+ Codex P2 列感知加固(definitions 带 start/end col,6 语言 call-info 带 col,同行紧凑代码的兄弟调用不再被偷)。⚠️ index-time 修复:既有索引需 clean re-index 才能拿到修正后的边。

aimasteracc:
[...]

PR #646, opened 2026-06-13
fix(nav): ghost caller rows eliminated — call sites attributed or counted

### LINKED ISSUE #637: [P1][dogfood] TOON encoder silently drops fields absent from a heterogeneous array's first row — callers' 49 bodies vanish

### LINKED ISSUE #638: [P2][dogfood] nav callers returns ghost caller (name='', line=0) ranked first
Dogfood patrol 2026-06-13: `nav action=callers` for apply_toon_format_to_response returns first row `{name:'', line:0}` — real call sites are inside methods at hyphae_subscribe_tool.py:112/165. Un-navigable; also the trigger for the TOON field-loss P1 (its keys built the table header). Fix = resolve enclosing function for the call site or drop unresolvable rows with a counted note.

aimasteracc:
Closes #638 (dogfood patrol 2026-06-13; the ghost row also triggered the now-fixed #637 TOON header).

## Root cause (_ast_extraction.py:1023-1029, extraction-time)
The per-file definition-span lookup was a NAME-KEYED dict — hyphae_subscribe_tool.py has two `execute` methods, the second span overwrote the first, the line-112 call found no containing span → edge written as `caller_name='' / caller_line=0` → ghost row ranked first. The same defect was cloned in the parse tier (call_graph.py:191/206) where the fallback additionally produced a WRONG-name attribution.

## Fix
- **Extraction**: keep ALL spans (list; parse tier keyed by name:start_line); innermost-span selection (#452/#484 family) unchanged. `_AST_CACHE_EXTRACTOR_VERSION` 10→11 (both sites).
- **Query-time**: genuinely module-level call sites (no enclosing function by design) are now EXCLUDED with a counted note `unattributed_call_sites: N` (only when >0) — never emitted as name:'' line:0. Chose exclusion over a `<module>` pseudo-caller (would ripple through qualified lookup/body-inlining/TOON for marginal value).

## E2E re-run of the original repro
Before: ghost `{name:'', line:0}` first, line-112 site missing. After: ghost rows [], both edges attributed to their real `execute` methods (caller_line 81/143).

## Test plan
- [x] RED-first: 5/9 failed exactly on the predicted bugs → 9 passed; version re-pin 10→11
- [x] Regression: callers/callees/extraction 148 + call-graph 254 + nav/TOON 107 + parity/contracts 205 (-n 0)
- [x] Golden FULL (6c): 96 passed, swift stash-proven pre-existing; ratchet exit=0
- [x] Lead independently re-ran 25 tests + verified both version sites + push diff=0

Two adjacent pre-existing quirks observed, untouched (surgical), filed separately: SQL-tier dedup collapses same-named methods in one file; parse-tier fallback misattributes module-level calls to the nearest preceding function.

### DIFF
--- a/tests/unit/test_ghost_caller_rows_638.py
+++ b/tests/unit/test_ghost_caller_rows_638.py
@@ -1,0 +1,234 @@
+"""RED-first tests for issue #638: ghost caller rows (name='', line=0).
+
+Root cause (extraction tier): ``_extract_call_edges`` keyed its per-file
+definition spans by bare function name.  Two same-named methods in different
+classes (the hyphae_subscribe_tool.py shape: two ``execute`` methods) collide
+— only the LAST span survives, so calls inside the EARLIER same-named method
+find no containing span and are written as ``caller_name='' / caller_line=0``
+edges.  ``nav action=callers`` then emits the un-navigable ghost row
+``{name: '', line: 0}`` ranked first.
+
+Same defect cloned in the parse tier: ``CallGraph.build`` collected
+``file_funcs`` into a name-keyed dict before ``_find_enclosing_func``.
+
+Query tier decision (#638): genuinely module-level call sites have no
+enclosing function BY DESIGN (the edge writer maps them to a file node).
+The callers tool must never emit them as ghost rows — it excludes them and
+reports ``unattributed_call_sites: N`` so nothing is silently dropped.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+from tree_sitter_analyzer._ast_extraction import _extract_call_edges
+from tree_sitter_analyzer.ast_cache import ASTCache
+from tree_sitter_analyzer.call_graph import CallGraph
+from tree_sitter_analyzer.core.parser import Parser
+from tree_sitter_analyzer.mcp.tools.callers_tool import CodeGraphCallersTool
+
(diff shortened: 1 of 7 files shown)


aimasteracc:
@codex review

Issue #648, opened 2026-06-13
[P3] parse-tier _find_enclosing_func attributes module-level calls to the nearest preceding function


aimasteracc:
Observed during #638: call_graph.py's parse-tier fallback, when a call site is module-level (no containing span), attributes it to the nearest PRECEDING function instead of excluding/marking it — wrong-name edges (e.g. _uri_to_selector@32 got someone else's call). The SQL tier now excludes-with-count (#638 fix); the parse tier should mirror. Evidence: #638 root-cause report.

aimasteracc:
Fixed by #650 — parse-tier _find_enclosing_func fallback (nearest preceding function) removed; module-level calls now return None and the existing 'caller_ref is None: continue' in CallGraph.build drops them correctly (no consumer needs every call to have a caller, so no counted-note needed — the parse tier just produces a smaller correct edge set, mirroring the SQL tier's #638 behavior).

PR #650, opened 2026-06-13
fix(nav): callers dedup keeps same-named methods; parse tier stops misattributing module calls

### LINKED ISSUE #647: [P3] callers SQL-tier dedup by (file, caller_name) collapses same-named methods in one file
Observed during #638: two `execute` methods in hyphae_subscribe_tool.py produce two correctly-attributed edges (caller_line 81/143) but the listed response shows ONE row — pre-existing dedup key (file, caller_name) ignores line. Both call sites are real distinct callers. Fix shape: dedup key gains caller_line (count impact on caller_count vs listed needs the honest-truncation conventions). Evidence: PR #646 e2e paste.

aimasteracc:
Closes #647 and #648 (adjacent quirks found during #638).

## #647 — SQL-tier dedup collapsed same-named methods
callers_tool.py:237 dedup key was `{caller_file}:{caller_name}` — two `execute` methods in one file collapsed to one row. Fix: key += `:{caller_line}`. Honesty check: before `caller_count=1, listed=1`; after `caller_count=2, listed=2` (both methods shown); `unattributed_call_sites` unchanged.

## #648 — parse-tier misattributed module-level calls
call_graph.py:341-344 fallback picked the nearest PRECEDING function for module-level call sites → wrong-name edges. Fix: removed the fallback; `_find_enclosing_func` returns None for module-level calls; the existing `if caller_ref is None: continue` in CallGraph.build already drops them correctly. **Consumer analysis**: no consumer of callers_of/callees_of needs every call to have a caller — the counted-note pattern (needed for the SQL tier) is NOT required here; the parse tier just produces a smaller correct edge set.

## Test plan
- [x] RED-first: 5 new tests + 3 pre-existing flipped (tested removed fallback) → 7 new + regression 255 passed (-n 0)
- [x] Re-pin: #638's `test_real_call_sites_resolve_to_enclosing_methods` → 2 rows, caller_count 1→2
- [x] Lead independently re-ran 16 tests + push diff=0

### DIFF
--- a/tests/unit/test_call_graph_cached.py
+++ b/tests/unit/test_call_graph_cached.py
@@ -249,13 +249,19 @@ class TestFindEnclosingFuncRange:
         result = cg.find_enclosing_func(file_funcs, 15)
         assert result.name == "inner"
 
-    def test_fallback_closest_start(self):
+    def test_module_level_gap_returns_none(self):
+        """A call between two non-overlapping functions has no enclosing function.
+
+        Before #648: the fallback returned the nearest preceding function (foo).
+        After #648: returns None — the call is module-level and must not be
+        attributed to any function (mirrors the SQL tier's counted-exclusion).
+        """
         cg = CallGraph("/tmp")
         f1 = FunctionRef("a.py", "foo", 5, "python", end_line=5)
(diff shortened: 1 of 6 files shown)


aimasteracc:
@codex review

PR #661, opened 2026-06-13
fix(edit): guard caller count excludes imports and non-call references, reconciled with nav callers

### LINKED ISSUE #655: [P1][dogfood] edit guard counts import statements and string literals as callers — instructs agent to update phantom sites
Dogfood patrol round 2 (develop e0c9e909), grep-confirmed:

`edit action=guard` for `_walk_for_symbols` reports **7 callers** (callers_by_file: test_cpp_method_resolution.py:1, test_ast_extraction.py:3, _ast_extraction.py:3). `nav action=callers` and `caller_tree` agree on **4** (ground truth). The 3 phantom:
- `tests/unit/test_ast_extraction.py:17` — `_walk_for_symbols,` is an **import** line, not a call (guard counts it; real calls there are only 262 + 308 = 2, guard says 3)
- `tests/unit/languages/test_cpp_method_resolution.py` — a **string literal** `"(_ast_extraction._walk_for_symbols)"` inside a test assertion

**Why P1**: the guard's `required_actions` tells the agent to update 'all 7 call sites' — an agent acting on it chases 3 non-existent sites. This is a correctness bug in the modification-safety surface, worse than a count discrepancy because it drives wrong action.

**Root cause area**: the guard's caller-counting path (likely a different tier than nav/callers which is correct post-#650) counts call-EDGE rows that include import edges / text matches. Fix shape: guard must count only `calls`-kind edges (exclude imports + any non-call reference), reconciling with nav callers' count. Extends #546 (pre-flight chain count incoherence).

aimasteracc:
Closes #655 (P1, dogfood patrol round 2; grep-confirmed correctness bug).

## Root cause: two different data sources
The guard (`trace_impact_tool.py`) uses **ripgrep text search**; nav callers uses `EdgeStore.query_callers()` with `WHERE kind='calls'` SQL. For `_walk_for_symbols` ripgrep found 7 text matches but only 4 are real call sites — the 3 phantom survived the pre-existing comment/docstring filter:
- `test_ast_extraction.py:17` — `_walk_for_symbols,` (import-block continuation line)
- `test_cpp_method_resolution.py:628` — symbol inside an inline `reason="(...)"` string arg

The guard's `required_actions` then told the agent to update all 7 — chasing 3 non-existent sites.

## Fix: extend the ripgrep filter (heuristic, consistent with J7 design)
Excludes (1) import lines (`from x import (` opener + multi-line continuations), (2) inline single/double-quoted string-literal matches. Chose this over replacing ripgrep with graph-edge lookup (which would need a full index rebuild in the guard path — much larger change, against J7's "AST-per-hit too expensive" philosophy).

## Reconciliation invariant (the key deliverable)
`test_reconciliation_invariant_guard_matches_nav_callers`: fixture with 2 real calls + 1 import → asserts `guard call_count == nav_caller_count == 2`. The two surfaces now AGREE.

## Test plan
- [x] RED-first: 11 tests (import-detection, inline-string, filter integration, e2e execute, reconciliation invariant) — all ImportError pre-fix → green
- [x] Golden FULL (6c): 96 passed, swift known-env; cost invariants 33 passed
- [x] Lead independently re-ran 11 tests + push diff=0

Known heuristic limits (documented): multi-line import nested-paren edge cases, mixed string+call on one line keeps the match (pinned).

### DIFF
--- a/tests/unit/mcp/test_guard_caller_count_excludes_non_calls.py
+++ b/tests/unit/mcp/test_guard_caller_count_excludes_non_calls.py
@@ -303,3 +303,118 @@ class TestTraceImpactExcludesImportAndStringMatches:
             f"guard counted {guard_count}, nav callers counted {nav_caller_count}. "
             "Import-line phantom inflated the guard count."
         )
+
+
+class TestFStringCallSitesNotFiltered:
+    """Codex P2 on #661: a call inside an f-string replacement field is real
+    code — must NOT be filtered (filtering it undercounts callers and makes
+    unsafe edits look safe)."""
+
+    @staticmethod
+    def _only_in_strings(line, symbol):
+        from tree_sitter_analyzer.mcp.tools.trace_impact_tool import (
+            _is_symbol_only_in_strings,
+        )
+
+        return _is_symbol_only_in_strings(line, symbol)
+
+    def test_fstring_replacement_field_call_is_code(self):
+        assert self._only_in_strings('msg = f"{my_func(x)}"', "my_func") is False
+
+    def test_fstring_literal_text_mention_is_string(self):
+        assert self._only_in_strings('msg = f"my_func failed"', "my_func") is True
+
+    def test_fstring_mixed_literal_and_field(self):
+        assert self._only_in_strings('f"my_func: {my_func(x)}"', "my_func") is False
+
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

Issue #658, opened 2026-06-13
[P2][dogfood] transitive_caller_count reports the _MAX_TRANSITIVE=200 cap as if it were an exact total (no is_capped label)


aimasteracc:
Dogfood round 2: `nav action=impact mode=function_impact` returns transitive_caller_count=200 which is the `_MAX_TRANSITIVE=200` cap (`if len(result) >= _MAX_TRANSITIVE: break`), not a real count. The field name implies an exact total; an agent reports '200 transitive callers' as precise. Fix per the honest-truncation convention (#444/#448): add `transitive_count_is_capped: true` or a pre-cap total, mirroring callers' truncated/listed_cap pattern.

aimasteracc:
Fixed by #662 — transitive_count_is_capped:true (+ symmetric callee flag) emitted only when the _MAX_TRANSITIVE=200 BFS cap fires; absent below, per the honest-truncation convention.

Issue #656, opened 2026-06-13
[P2][dogfood] nav impact blast_radius mode returns only the seed (FunctionRef dict-key identity miss)


aimasteracc:
Dogfood round 2: `nav action=impact mode=blast_radius function_names=['_walk_for_symbols']` returns total_affected_functions=1, total_files_at_risk=0 — only the seed, no propagation, for a function with 4 real callers. Root cause traced: CallGraph `caller_refs_of(func)` uses the FunctionRef OBJECT as a dict key, but `resolve_targets` produces a DIFFERENT FunctionRef instance than the one used to build `_callers` → key miss → empty. blast_radius mode is effectively non-functional. Fix: key _callers on a stable identity (qualified name + file + line) not object identity, or add __hash__/__eq__ to FunctionRef.

aimasteracc:
Investigated in #662: FunctionRef.__hash__/__eq__ (file+name+line) are already present (call_graph.py:76-86), so blast_radius propagation works — CallGraph('.') returns 25,302 affected, not 1. The dogfood's total_affected=1 was a stale/edgeless index (BuildProjectIndexTool doesn't populate call edges; only ASTCache.index_project does). Added TestBlastRadiusPropagation regression guards so it can't silently break. Closing as not-a-code-bug + guarded.

Issue #444, opened 2026-06-10
[P3][dogfood] Hyphae select silently caps at 100 — no truncated flag, no total


aimasteracc:
## Summary
`search action=select` (Hyphae) silently truncates at the default cap: `count: 100` with **no `truncated` flag and no total**, so the agent cannot tell "exactly 100" from "100 of 400".

## Repro (develop @ 32412d6e)
```
search action=select selector=".function:calls(#apply_toon_format_to_response)"
→ count: 100, symbols: [100 rows], no truncated field, no total_count
```
`apply_toon_format_to_response` has 80 downstream files per `edit action=safe`, and dozens of per-tool `execute` callers + tests — the true caller count is almost certainly >100.

## Impact
A blast-radius query that under-reports without saying so is worse than one that errors: the agent plans a refactor against "100 callers" and misses the rest. This also violates the project's own exact-count discipline (a count that can silently mean ">=count" is the API equivalent of a `>=` assertion).

## Suggested fix
Return `truncated: true` + `total_count` (cheap COUNT(*) alongside the LIMIT query) whenever the cap is hit, and have `next_step` say "N more — narrow with :in(path)". Apply the same contract to every capped facade action.

aimasteracc:
Fixed by #489 (merged)。Hyphae select 响应新增 total_matches(截断前总数)+ truncated 旗 + 截断时 next_step 给出收窄选择器指引(:in(path)/[file=]/[language=]/:not)。Codex 两个 P2 加固:① 恰好 100 个唯一匹配不再误报截断(truncated = total > listed 末尾判定);② :not(...) 重入 eval 不再清零计数器(计数局部化,最外层末尾一次性赋值)。精确钉测试覆盖 150/100/边界/重入四种形态。

aimasteracc:
**Retested on develop @ 465ea828 (#489) with a full fresh index — CONFIRMED FIXED ✓**

`.function:calls(#apply_toon_format_to_response)` → `count: 100, total_matches: 119, truncated: true`, next_step: *'Results truncated at 100 of 119 matches. Narrow the selector with :in(path), [file=], [language=], or :not(...) … or raise max_results.'* Exactly the contract this issue asked for (and the true caller count was indeed >100). Good to close.

PR #662, opened 2026-06-13
fix(nav): blast_radius accepts singular param + labels transitive cap; FunctionRef propagation regression-guarded

### LINKED ISSUE #657: [P2][dogfood] nav impact blast_radius: schema says function_name (singular), validator requires function_names (plural)
Dogfood round 2: `nav action=impact mode=blast_radius function_name=X` → hard ERROR 'function_names is required for blast_radius mode'. The schema description documents `function_name` (singular, used by function_impact mode); blast_radius needs `function_names` (plural, undocumented). An agent following the schema fails with no hint. Fix: document function_names in the schema for blast_radius mode (or accept both); the facade-actions doc (#519) should show the per-mode param.

aimasteracc:
Closes #657 and #658; documents #656 (dogfood round 2).

## #657 — singular function_name alias for blast_radius
`validate_arguments` rejected blast_radius calls with only `function_name` (singular) — required `function_names` (plural, undocumented). Now accepts either (singular → wrapped in list); both schema descriptions mention blast_radius mode.

## #658 — transitive cap labeled
`transitive_caller_count=200` was the `_MAX_TRANSITIVE=200` BFS cap presented as exact. Now adds `transitive_count_is_capped: true` (+ symmetric callee flag) ONLY when the cap fires — absent below, per the honest-truncation convention (#444/#448).

## #656 — blast_radius propagation: already correct, regression-guarded
Investigated: `FunctionRef.__hash__/__eq__` (on file_path+name+start_line) are already present (call_graph.py:76-86), so `caller_refs_of` dict lookup works — `CallGraph('.')` propagates to 25,302 affected (not 1). The dogfood's total_affected=1 was almost certainly a stale/edgeless index (BuildProjectIndexTool doesn't populate call edges — only ASTCache.index_project does). Added `TestBlastRadiusPropagation` (transitive chain → 3, four direct callers → 5) as a regression guard so this can't silently break.

## Test plan
- [x] RED-first: 4 failed (singular accept, singular-wraps-list, schema describes, capped-flag-at-limit) → 44 in impact tool + 238 regression passed (-n 0)
- [x] facade-actions.md no regen needed (verified)
- [x] Lead independently re-ran 44 tests + push diff=0

### DIFF
--- a/tests/unit/test_codegraph_impact_tool.py
+++ b/tests/unit/test_codegraph_impact_tool.py
@@ -470,3 +470,200 @@ class TestImpactTestPartition:
         names = {r["name"] for r in result}
         assert "real_dep" in names
         assert "fake_dep" in names
+
+
+# ---------------------------------------------------------------------------
+# #656 — blast_radius transitive propagation (FunctionRef dict-key identity)
+# ---------------------------------------------------------------------------
+
+
+class TestBlastRadiusPropagation:
+    """#656: blast_radius must propagate transitively, not just return the seed.
+
+    The graph walks caller_refs_of(current) where *current* may come from a
+    previous iteration's caller list.  With __hash__/__eq__ keyed on
+    (file_path, name, start_line), a FunctionRef VALUE retrieved from _callers
+    can be used as a KEY for the next-level lookup even if it is a different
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #663, opened 2026-06-13
fix(structure): class_tree subclasses drops same-name parentless ghost

### LINKED ISSUE #659: [P2][dogfood] class_tree subclasses: MockLanguagePlugin (no parent) falsely listed as LanguagePlugin subclass
Dogfood round 2: `structure action=class_tree mode=subclasses class_name=LanguagePlugin` lists `MockLanguagePlugin` (analysis_engine.py:157, `class MockLanguagePlugin:` with NO inheritance) as a subclass with parents:null. A same-named `class LanguagePlugin(Protocol)` at line 47 in the same file likely triggers a spurious extends edge resolving by bare name to the wrong same-named class. Agent counts 26 subclasses, one is a ghost. Fix: extends-edge resolution must respect the actual base in the class's bases tuple, not same-name collision (related to the bare-name inheritance resolution family).

aimasteracc:
Closes #659 (dogfood patrol round 2).

## Diagnosis
Extraction is correct: `_ast_cache_write.py:341-359` only creates extends edges when `parents` is truthy, so `MockLanguagePlugin` (analysis_engine.py:157, no inheritance) produces NO edge. The real edge comes from a DIFFERENT same-named `MockLanguagePlugin` (test_manager.py:17) that legitimately inherits LanguagePlugin.

**Root cause (query side)**: `ClassHierarchy.subclasses_of` looks up `_classes['MockLanguagePlugin']` and returns ALL ClassInfo entries for any class with that name — including the parentless analysis_engine.py one. Both got emitted → ghost subclass.

## Fix (query-time, class_hierarchy.py)
`_build_edges_from_edge_store` records confirmed `(file_path, class_name)` pairs from edge source node IDs; `subclasses_of` emits a ClassInfo only if its `parents` is non-empty OR `(file, name)` is in the confirmed set. Extraction untouched — `_AST_CACHE_EXTRACTOR_VERSION` stays 11.

## Test plan
- [x] RED-first: `test_same_name_class_without_parents_not_listed_as_subclass` failed → green
- [x] Regression: 26 class-hierarchy + 46 hierarchy/unresolved/CLI passed (-n 0)
- [x] Lead independently re-ran 26 tests + push diff=0

### DIFF
--- a/tests/unit/test_codegraph_class_hierarchy_tool.py
+++ b/tests/unit/test_codegraph_class_hierarchy_tool.py
@@ -418,3 +418,56 @@ class TestClassHierarchyEdgeStore:
             )
         finally:
             cache.close()
+
+
+class TestSameNameDifferentBaseCollision:
+    """Codex P2 on #659: two same-named children inheriting DIFFERENT bases
+    must each only appear under their actual base."""
+
+    def test_same_name_children_different_bases_isolated(self, tmp_path):
+        from tree_sitter_analyzer.class_hierarchy import ClassHierarchy, ClassInfo
+
+        h = ClassHierarchy.__new__(ClassHierarchy)
+        # Minimal hand-built state (bypass index): two Worker classes, one
+        # inherits Base, the other inherits Other.
+        from collections import defaultdict
+
+        h._built = True
+        h._parent_map = defaultdict(list)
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

Issue #669, opened 2026-06-13
[P3][dogfood] nav xref count semantics undocumented: caller_count=0 vs import_dependent_count=5 confuses agents


aimasteracc:
Dogfood round 3: `nav action=xref mode=file` for _ast_extraction.py returns `caller_count:0, callee_count:50, import_dependent_count:5, file_dependent_count:0, callers:[]` — 5 files clearly import from it, and nav callers shows 6 function-level callers, yet caller_count/file_dependent_count read 0. The three count fields (callers / import_dependents / file_dependents) have distinct meanings that aren't explained in the response, so an agent reads '0 callers' as 'no inbound deps' (wrong). Fix (low-risk): add a one-line semantics note or an `agent_summary` that disambiguates the three counts. NOT a code bug — clarity/honest-envelope gap.

aimasteracc:
Fixed in #671 — nav xref file mode now computes real inbound caller_count/file_dependent_count (resolved-only via callee_resolved_file), with a complete agent_summary (verdict + summary_line + count_semantics) on both MCP and direct paths, chunked under the SQLite variable cap.

PR #671, opened 2026-06-13
fix(nav): impact file-level recovery hint + xref count semantics

### LINKED ISSUE #668: [P3][dogfood] nav impact file-level query has no path + no recovery hint to nav xref
Dogfood round 3: an agent asking 'what does _ast_extraction.py affect' via `nav action=impact mode=blast_radius file_path=...` hits `ValueError: function_names is required for blast_radius mode` with recovery_hint 'A required parameter is missing'. All three impact sub-modes (blast_radius/function_impact/risk_score) are function-only; the file-level alternative (`nav action=xref mode=file`) exists but the error doesn't guide the agent there. Fix (low-risk): the impact validation error for a file_path-only call should recovery_hint → 'for file-level dependents use nav action=xref mode=file'. No code-path bug — pure UX.

aimasteracc:
Closes #668 and #669 (dogfood patrol round 3 — UX/clarity, response-text only, no extraction change).

## #668 — impact file-level recovery hint
`nav action=impact mode=blast_radius file_path=X` (no function_name) raised `function_names is required for blast_radius mode` with a generic hint. The message now appends the actionable alternative: `...; for file-level dependents use nav action=xref mode=file` — agents get the file-level path instead of guessing. Error preserved (still invalid call).

## #669 — xref count semantics
`nav action=xref mode=file` returned caller_count/import_dependent_count/file_dependent_count with no explanation (agents misread caller_count:0 as 'no inbound deps'). Added `agent_summary.count_semantics` disambiguating the three: callers = function-level call sites; import_dependents = files importing this module; file_dependents = files with call-graph edges in.

## Test plan
- [x] RED-first: 8 tests (hint-mentions-xref, count_semantics present + content) → green; 161 regression (impact/xref/nav facade/contracts) + 22 error-recovery + 3 boundary passed (-n 0)
- [x] Golden FULL (6c): 96 passed, swift known-env — response-text fields not golden-snapshotted (zero drift); ratchet exit=0
- [x] Lead independently re-ran 66 tests + push diff=0

### DIFF
--- a/tests/unit/mcp/test_error_recovery_coverage.py
+++ b/tests/unit/mcp/test_error_recovery_coverage.py
@@ -81,6 +81,23 @@ class TestBuildAgentFriendlyError:
         assert result["error_type"] == "validation"
         assert "suggested_tool" not in result
 
+    def test_blast_radius_required_keeps_xref_guidance(self):
+        # #668: the blast_radius "function_names is required ..." message also
+        # contains "required", so the blast_radius-specific rule MUST win over
+        # the generic "required" rule — otherwise the agent-facing recovery_hint
+        # / next_step lose the file-level xref guidance and #668 isn't fixed on
+        # the real MCP error path.
+        err = ValueError(
+            "function_names is required for blast_radius mode; "
+            "for file-level dependents use nav action=xref mode=file"
+        )
+        result = build_agent_friendly_error("codegraph_impact", err)
+        _assert_canonical(result)
+        assert result["error_type"] == "validation"
+        assert result["suggested_tool"] == "nav action=xref mode=file"
+        assert "nav action=xref mode=file" in result["recovery_hint"]
+        assert result["agent_summary"]["next_step"] == result["recovery_hint"]
+
     def test_validation_error_pattern(self):
         err = ValueError("format must be one of: full, compact")
         result = build_agent_friendly_error("analyze_file", err)
--- a/tests/unit/test_codegraph_impact_tool.py
+++ b/tests/unit/test_codegraph_impact_tool.py
@@ -667,3 +667,46 @@ class TestTransitiveCapFlag:
         result = tool._function_impact(graph, "hub", None, 10)
         assert result["transitive_callee_count"] == _MAX_TRANSITIVE
         assert result["transitive_callee_count_is_capped"] is True
+
+
+# ---------------------------------------------------------------------------
+# #668 — blast_radius file-only call: recovery_hint guides to nav xref
+# ---------------------------------------------------------------------------
+
+
+class TestBlastRadiusFileOnlyRecoveryHint:
+    """#668: when blast_radius is called with file_path only (no function_names),
+    the error message must contain actionable guidance to nav action=xref mode=file.
+
+    The recovery_hint is derived from the ValueError message by
+    error_recovery._classify().  A specific rule matching 'blast_radius' must
+    be installed BEFORE the generic 'required' rule so the hint is contextual.
+    """
+
+    def test_blast_radius_error_message_contains_xref_hint(self):
+        """validate_arguments raises; error message mentions 'xref'."""
+        tool = CodeGraphImpactTool()
+        with pytest.raises(ValueError) as exc_info:
+            tool.validate_arguments(
+                {"mode": "blast_radius", "file_path": "src/_ast_extraction.py"}
+            )
+        msg = str(exc_info.value)
+        assert "xref" in msg
+
+    def test_blast_radius_error_message_mentions_file_mode(self):
+        """Error message must reference mode=file so the agent knows which xref mode."""
+        tool = CodeGraphImpactTool()
+        with pytest.raises(ValueError) as exc_info:
+            tool.validate_arguments(
+                {"mode": "blast_radius", "file_path": "src/module.py"}
+            )
+        msg = str(exc_info.value)
+        assert "mode=file" in msg
+
+    def test_blast_radius_with_no_args_error_still_mentions_xref(self):
+        """Even with no file_path provided, the blast_radius error hints at xref."""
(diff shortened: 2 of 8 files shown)


aimasteracc:
@codex review

aimasteracc:
## Codex P2 triage (both REAL — fixed in ddd1069f)

|---|---------|---------|-----|
| 1 | impact_tool:423 — blast_radius recovery hint never reaches agent (`_classify` matches generic `required` first) | ✅ REAL | Added a `blast_radius mode`-specific recovery rule **ahead of** the generic `required` rule in `error_recovery.py`, so `recovery_hint`/`next_step` carry the `nav action=xref mode=file` guidance. #668 now fixed on the real MCP error path. |
| 2 | xref_tool:148 — count_semantics documents inbound counts the engine never computes | ✅ REAL | `_file_callers` bound `file_path = ? AND file_path != ?` (same value → always empty); `_find_file_dependents` selected edges where THIS file is the caller then discarded all as same-file. Both rewritten to resolve inbound callers/file-dependents by the file's defined symbol names. caller_count/file_dependent_count are now real (the shared fix corrects symbol mode too). |

RED-first: 3 new tests (file-mode inbound counts `== 2` on the a/b/c fixture, symbol-mode file_dependents, blast_radius recovery-hint routing) all fail when the production fix is reverted. Bandit B608 handled via the project's named-prefix/suffix + `?`-placeholder convention (`_route_cache.py`).

@codex review

aimasteracc:
## Codex P2 round-2 triage (REAL — fixed in 59fd5b4d)

**xref.py:397 — bare callee_name over-counts same-named callables across files** → ✅ REAL, reproduced:
- a.py and b.py both define `helper`; c.py calls resolved `b.helper` (`callee_resolved_file='b.py'`). Name-only matching wrongly credited a.py with c.py.
- Fix: gate both inbound queries by `callee_resolved_file = <this file>`, with a name-only fallback only for genuinely unresolved rows (`callee_resolved_file = ''`). Edges the resolver tied to a *different* file are now excluded.
- Verified: same-name fixture → a.py caller_count `0`, b.py `1` (c.py); the original a/b/c fixture is unaffected (caller_count stays `2`). RED-first test `test_file_xref_excludes_same_name_resolved_to_other_file` fails without the gate. The `idx_edges_resolved` index already covers the new predicate.

@codex review

aimasteracc:
## Codex P2 round-3 triage (REAL — fixed in 9374d1f3)

**codegraph_xref_tool.py:145 — partial agent_summary on the direct path** → ✅ REAL:
- The #669 `agent_summary` carried only `count_semantics`. On the CLI-bridged direct `execute()` path (`mcp_commands/_helpers.py`) the MCP success post-hook never runs, so `agent_summary.verdict`/`summary_line` were missing — breaking the mirror contract for consumers that branch on `agent_summary`.
- Fix: populate `verdict` + `summary_line` + `next_step` alongside `count_semantics` at build time, so both MCP and direct paths emit a complete agent_summary. (On the MCP path the post-hook's `setdefault` leaves my fields intact and mirrors my summary_line to top-level — no divergence.)
- RED-first: `test_file_mode_agent_summary_carries_canonical_fields` asserts `agent_summary.verdict == result.verdict` + non-empty `summary_line` on the direct path; fails without the fix.

This is the 4th real finding on this PR across the dev↔review cycle — all reproduced, none dismissed.

@codex review

aimasteracc:
## Codex P2 round-4 triage (both REAL — fixed in 3891f37d)

**A — xref.py: unresolved-fallback over-counts common method names** → ✅ REAL:
- a.py defining `format` + another file's unresolved `"x".format()` (callee_resolved_file='') was credited to a.py via my round-2 `OR callee_resolved_file=''` fallback. Common names (format/get/items) would inflate blast radius.
- Fix: **drop the unresolved fallback** — count ONLY edges resolved to this file. Under-count beats over-count for a blast-radius signal. Verified: the a/b/c fixture (caller_count stays 2) and same-name fixture (a=0/b=1) are unaffected because their edges ARE resolved; `count_semantics` now says 'resolved-only'. RED test: a.py defining `format` gets 0 callers from a `str.format` call.

**B — top-level summary_line still missing on direct path** → ✅ REAL:
- Round-3 set summary_line only inside `agent_summary`; `--codegraph-xref --format json` reads the **top-level** `summary_line`, absent on the direct path. Fix: set it at top level too (and assert top-level == agent_summary.summary_line). RED test added.

Round 4 of dev↔review. Every finding reproduced + fixed RED-first; none dismissed.

@codex review

aimasteracc:
## Codex P2 round-5 triage (REAL — fixed in 19f7e0c5)

**xref.py:321 — IN() queries exceed SQLite bound-variable cap** → ✅ REAL:
- A file with ~1000+ symbols binds `len(names)+2` variables per inbound IN(...) query; on SQLite builds capped at 999 `SQLITE_MAX_VARIABLE_NUMBER` (the repo already chunks `_route_cache.py` for exactly this) it raises `OperationalError: too many SQL variables` instead of returning data.
- Fix: route both inbound queries through a shared `_inbound_edge_rows` helper that chunks the name list (`_XREF_NAME_CHUNK=900`); callers dedup + order + cap 50 across chunks. Single-chunk (common) behavior unchanged.
- RED-first: `test_file_xref_chunks_large_symbol_list` shrinks the chunk to 2 over a 5-symbol file with the caller targeting a 2nd-chunk symbol — verified it fails when the loop is mutated to first-chunk-only. (Modern SQLite defaults to 32766 vars so the raw limit isn't portably reproducible in a test; the chunk-correctness guard is the deterministic equivalent.)

Round 5. All 7 P2s reproduced + fixed RED-first. The obvious failure modes (resolution, precision, contract completeness, summary mirror, SQLite cap) are now covered.

@codex review

aimasteracc:
@codex review

PR #674, opened 2026-06-13
test: ratchet batch-34 — pin 5 loose assertions in test_logging.py (439→434)

aimasteracc:
Pins 5 handler-count assertions in `test_logging.py` to exact measured values (`== 1`/`== 2`), all verified by the green run (66 passed).

**Deliberately small (5, not ~20):** this round's next-ranked files were poor targets and are deferred to batch-35:
- `integration_tests.py` has **6 pre-existing failures** (CSV TOON response missing `table_output` — filed as a separate bug) → its loose asserts are unreachable, can't be measured by a green run.
- The CLI-output files (`test_cli_queries`, `test_cli_table_partial_query`) need **content-invariant** assertions, not byte counts that vary by platform temp-stem length.

Quality over count — per the exact-assertion lock, every pin here is a measured fact from a passing test, not a guess.


aimasteracc:
@codex review

PR #676, opened 2026-06-13
test: ratchet batch-35 — pin 15 loose assertions in top-4 AST-ranked files (434→418)

aimasteracc:
Pins 15 loose assertions to exact measured values across 4 pre-vetted deterministic files (all measured from green runs), + 1 nondeterministic exemption.

## Files
- `test_queries_ruby.py` — query-string lengths (==119/==274/==69) + ALL_QUERIES count (==24)
- `test_language_detector_helpers.py` — extension map (==54), content weights (==9), per-language pattern dict (restructured loop → exact dict)
- `test_parser_readiness_records.py` — next_steps (==8/==8/==2), verification_commands (==3)
- `test_parser.py` — supported languages (==22), cache stat_hits (==1), misses (==2), **cache maxsize == 2000**

## Exemption (1, justified)
- `mtime_ns > 0` — genuinely nondeterministic filesystem timestamp; `# ratchet: nondeterministic` marker.

**Lead note:** the pod initially exempted `cache_maxsize` too ("env-configurable"), but the test doesn't set `TSA_PARSER_CACHE_SIZE`, so it's the deterministic CI default 2000 — re-pinned `== 2000` so a default change goes red (the >=1000 design floor stays in the docstring). Every other value measured from a passing test, never guessed.

### DIFF
--- a/scripts/loose_assertion_baseline.txt
+++ b/scripts/loose_assertion_baseline.txt
@@ -74,7 +74,13 @@
 #   missing from CSV TOON, filed separately) so its asserts are unreachable,
 #   and the CLI-output files need content-invariant pins, not byte counts that
 #   vary by platform temp-stem length — both deferred to batch-35).
+#   batch-35 (AST): 434 -> 418, 2026-06-13 (15 exact pins + 1 nondeterministic
+#   exemption in 4 files: test_parser_readiness_records.py/test_language_detector_helpers.py/
+#   test_parser.py/test_queries_ruby.py; loop assert in test_key_languages_have_patterns
+#   restructured to per-language exact dict; cache_maxsize pinned == 2000 (the
+#   deterministic CI default; env override is out-of-test); only mtime_ns exempt
+#   as a genuinely nondeterministic filesystem timestamp).
 
-BASELINE_COUNT=434
+BASELINE_COUNT=418
 MEASUREMENT_DATE=2026-06-13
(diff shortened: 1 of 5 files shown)


aimasteracc:
@codex review

aimasteracc:
Force-pushed c91df12f: fixed two lead-corrected pins (cache_maxsize == 2000 instead of an unjustified exemption; get_supported_languages == 23 to match CI's full grammar set — the pod's local == 22 was missing the optional tree-sitter-swift wheel, verified == 23 after installing it). @codex review

aimasteracc:
## Codex P1 triage: already fixed (stale-commit review)

Codex reviewed the older commit `5f1d57f7` (`== 22`) and correctly predicted that CI's `uv sync --all-extras` installs `tree-sitter-swift` → 23 languages, so `== 22` fails. **This was already diagnosed and fixed before the review landed:** the current HEAD `c91df12f` pins `== 23` (verified by installing the swift extra locally to match CI), and all CI axes are now green (13 pass).

On the suggested alternative — *"derive the expected value from the installed loader mapping"*: rejected. Deriving the expected count from `LANGUAGE_MODULES` would make the assertion tautological (it would pass regardless of the actual registered count), defeating the locked exact-assertion rule whose whole purpose is that a grammar add/remove goes **red** and forces a conscious re-pin. The hardcoded `== 23` with the docstring note is the rule-compliant choice.

PR #678, opened 2026-06-13
fix(mcp): verdict-aware summary_line synthesis (#672)

### LINKED ISSUE #672: [P3][dogfood] Canonical envelope says "<tool>: ok" verdict-blind — WARN/ERROR summary_line lies, inner recovery hint dropped
## Summary

The canonical-envelope summary-line synthesizer says `"<tool>: ok"` **regardless of verdict**. On a WARN/ERROR/NOT_FOUND response the synthesized one-liner literally says **"ok"**, contradicting the verdict, and the inner tool's descriptive `summary_line` + recovery `next_step` are dropped.

Reproduces deterministically on the **live MCP server** (real consumer path, not `execute()`).

## Reproduction

`index action=status` against a project whose `.ast-cache` is empty/missing:

```jsonc
{
  "verdict": "WARN",
  "indexed": false,
  "hint": "Index missing or empty. Run the `index` tool with action=auto to build the cache.",
  "summary_line": "index: ok",          // ← lies: verdict is WARN
  "agent_summary": {
    "summary_line": "index: ok",        // ← lies
    "next_step": "",                    // ← inner tool's recovery hint was dropped
    "verdict": "WARN"
  }
}
```

The inner `CodeGraphStatusTool` correctly builds:
- `agent_summary.summary_line = "codegraph_status: index missing or empty"`
- `next_step = "Index missing or empty. Run the index tool with action=auto…"`

…but both are lost by the time the envelope reaches the agent. Healthy index is also affected (less harmfully): `verdict: INFO` + `"index: ok"` + `next_step: ""` — the inner's `"Index is healthy — proceed with nav or search tools"` next_step is also gone.

## Impact (P3, agent-UX)

An agent gating on `agent_summary.summary_line` / `next_step` (the documented place agents read) to decide *"is the index ready, what next?"* sees **"ok"** with no next step on the WARN path → proceeds to `nav`/`search` and gets phantom-empty results. The `verdict` field is correct, so a careful agent reading `verdict` is fine — that bounds severity to P3.

## Root cause (actionable)

`tree_sitter_analyzer/mcp/server_utils/error_recovery.py:428-429`:

```python
synthesized = (
    f"{tool_name}: {identifier_value}" if identifier_value else f"{tool_name}: ok"
)
```

The synthesizer is **verdict-blind** — it appends `": ok"` even when `response["verdict"]` is `WARN`/`ERROR`/`NOT_FOUND`. Two compounding issues:

1. **Verdict-blind "ok"**: when falling back to synthesis, the literal should reflect the verdict (e.g. `": index missing"` / `": warn"` / `": error"`), never an unconditional `"ok"`.
2. **Inner agent_summary lost through the facade**: the `index` facade delegates to `CodeGraphStatusTool` verbatim, yet the inner's rich `agent_summary.summary_line` + `next_step` don't survive to the boundary — forcing the synthesizer's step-2 fallback in the first place. The inner hint should be preserved (step-1 mirror should have used `"codegraph_status: index missing or empty"`).

## Suggested fix

- Make the synthesizer verdict-aware: only append `": ok"` when `verdict in (None, "INFO", "OK", "SUCCESS")`; otherwise derive the suffix from verdict/hint.
- Trace why the inner `agent_summary` is dropped before `_resolve_summary_line` runs for facade-routed status (so step-1 mirror fires instead of step-2 synthesis).

## RED-first test

`tests/unit/mcp/server_utils/` (or the envelope test that owns `_resolve_summary_line`): assert that a `{verdict: "WARN", ...}` response with no pre-set `summary_line` synthesizes a line that does **not** contain `"ok"` and that a verdict-appropriate token is present — exact-string pin, not substring-absence-only.

aimasteracc:
Closes #672.

## Problem
The canonical-envelope summary-line synthesizer (`error_recovery.py`) appended `": ok"` **regardless of verdict**. A WARN/ERROR/NOT_FOUND response with no `summary_line` / `agent_summary.summary_line` / identifier — e.g. `index action=status` on an empty cache — got a synthesized `"<tool>: ok"` that contradicts its own `verdict: WARN` and misleads an agent that gates on `summary_line`.

## Fix
Add a `_verdict_suffix` helper mapping `WARN`/`ERROR`/`NOT_FOUND` → `warning`/`error`/`not found`; success-like or absent verdicts keep `"ok"`. Reads the top-level `verdict` first, then `agent_summary.verdict`. Identifier-based summary lines (`<tool>: <file_path>`) are unchanged and still take precedence.

## RED-first
`TestSynthesizedSummaryLineIsVerdictAware`: WARN/ERROR/NOT_FOUND honest suffixes, INFO-stays-ok regression, agent_summary-verdict read, identifier-precedence, and no-verdict→ok fallback. The 4 verdict tests fail without the fix; new lines fully covered.

## Scope note
This fixes the verdict-blind "ok" (the visible lie). The secondary observation in #672 — that a facade can strip the *inner* tool's rich agent_summary before this synthesizer runs (forcing the fallback at all) — is a separate, deeper investigation left for a follow-up.

### DIFF
--- a/tests/unit/mcp/test_error_recovery_coverage.py
+++ b/tests/unit/mcp/test_error_recovery_coverage.py
@@ -281,3 +281,66 @@ class TestSuccessEnvelopeNextStepMirror:
         result = ensure_canonical_success_envelope("nav", response)
         # Contract unchanged when there is nothing to mirror.
         assert result["agent_summary"]["next_step"] == ""
+
+
+class TestSynthesizedSummaryLineIsVerdictAware:
+    """#672/#678: when no summary_line / agent_summary.summary_line / identifier
+    is present, the synthesized fallback must reflect the verdict — never claim
+    ``"ok"`` on any non-success verdict (WARN/ERROR/NOT_FOUND/CAUTION/REVIEW/
+    UNSAFE; e.g. index action=status on an empty cache, or viz action=similarity
+    returning CAUTION), which lies to an agent that gates on summary_line. Only
+    the success-like allowlist (INFO/SAFE) keeps ``"ok"``."""
+
+    def test_warn_verdict_does_not_say_ok(self):
+        response = {"success": True, "verdict": "WARN", "indexed": False}
+        result = ensure_canonical_success_envelope("index", response)
+        assert result["summary_line"] == "index: warn"
+        assert result["agent_summary"]["summary_line"] == "index: warn"
+
+    def test_not_found_verdict_suffix(self):
+        response = {"success": True, "verdict": "NOT_FOUND"}
+        result = ensure_canonical_success_envelope("nav", response)
+        assert result["summary_line"] == "nav: not found"
+
+    def test_error_verdict_suffix(self):
+        response = {"success": True, "verdict": "ERROR"}
+        result = ensure_canonical_success_envelope("search", response)
+        assert result["summary_line"] == "search: error"
+
+    def test_caution_review_unsafe_verdicts_are_not_ok(self):
+        # Codex #678 P2: the security/quality verdicts must not fall through to
+        # "ok" — they're the exact cases an agent must NOT read as success.
+        for verdict, expected in (
+            ("CAUTION", "viz: caution"),
+            ("REVIEW", "viz: review"),
+            ("UNSAFE", "viz: unsafe"),
+        ):
+            response = {"success": True, "verdict": verdict}
+            result = ensure_canonical_success_envelope("viz", response)
+            assert result["summary_line"] == expected
+
+    def test_info_and_safe_verdicts_still_ok(self):
+        # Regression: the success-like allowlist keeps the "ok" suffix.
+        for verdict in ("INFO", "SAFE"):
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

aimasteracc:
## Codex P2 triage: REAL — fixed in f80269ad

Good catch. My initial denylist map (`WARN`/`ERROR`/`NOT_FOUND`) left `CAUTION`/`REVIEW`/`UNSAFE` falling through to `"ok"` — exactly the security/quality verdicts (e.g. `viz action=similarity` duplicate findings) an agent must NOT read as success.

Switched to a success-like **allowlist** (`{INFO, SAFE, OK, SUCCESS}`): those keep `"ok"`; every other canonical verdict synthesizes its own lowercased suffix (`caution`/`review`/`unsafe`/`warn`/`error`/`not found`). This also future-proofs it — a newly-added non-success verdict can never silently become `"ok"` again. Added RED tests for CAUTION/REVIEW/UNSAFE + an INFO/SAFE regression.

@codex review

Issue #515, opened 2026-06-11
[P2][dogfood] structure explore rejects its documented 'symbols' param — multi-symbol contract unusable as documented


aimasteracc:
## Summary
`structure action=explore` is documented as the **multi-symbol** source explorer — facade schema: "show source of several related symbols grouped in one capped response. **Params: symbols**, language" — but the inner tool **rejects the documented param**:

## Repro (develop @ af493be1)
```
explore symbols=["legacy_to_facade","is_facade_name"]
→ ValueError: unknown parameter 'symbols' for tool 'codegraph_explore' — did you mean 'symbol'?

explore symbol=["legacy_to_facade","is_facade_name"]      (list into the suggested param)
→ "query is required (or pass it as `symbol`)"            (confusing: symbol WAS passed)

explore symbol="legacy_to_facade"                          (single string)
→ works ✓ (source inlined correctly)
explore symbol="legacy_to_facade,is_facade_name"           (comma string)
→ returns something, but this form is documented nowhere
```

## Impact
The whole pitch of explore (and the server instructions: "don't loop codegraph_node — ONE explore call returns several symbols") depends on multi-symbol input. An agent following the facade schema fails on the first call; the strict-param hint sends it to a param that then rejects lists with an unrelated error message. Doc/schema drift class — same family as the skills (#437) and guidance (#440) sweeps, but in the facade description itself.

## Suggested fix
Accept `symbols: [..]` (list) at the facade/inner schema and keep `symbol` as single-value alias; document the comma form if it stays; fix the list-rejection error to say "expected string". Contract test: facade-doc params ⊆ inner accepted params for every action (this drift is mechanically detectable).

aimasteracc:
已由 #526 修复并合入 develop(合默认分支前手动关闭)。修复:validate_arguments 归一 symbols=[...] 与误传 list 的 symbol;公开 schema 宣告 symbols(array)+ symbol(string-or-array,Codex P2 二轮——严格客户端不再 pre-execute 拒绝);facade 描述同步纠正(删除从不存在的 language)。3 个 RED→GREEN 新测试 + 契约 53 + callability 75 全绿。

aimasteracc:
[...]

PR #679, opened 2026-06-13
fix(mcp): accept output_format as a universal envelope param (#651)

### LINKED ISSUE #651: [P3][dogfood] output_format rejected by modification_guard + batch_search — universal envelope param not wired everywhere
## `output_format` is a universal envelope param but two tools reject it — inconsistent contract
Most facade actions accept `output_format` (it's how TOON-vs-JSON is selected, the envelope's universal control). But the strict-param guard on at least two tools rejects it outright:

```
edit action=guard (modification_guard) {symbol, modification_type, file_path, output_format}
→ ValueError: unknown parameter 'output_format' for tool 'modification_guard'
search action=batch  → same (tracked partly in #569)
```
(Confirmed on develop @ b2bee8d2.) search/nav/structure/health/index/viz actions all accept it across hundreds of prior probes.

## Why it matters
An agent that sets `output_format` uniformly across calls (the reasonable default for a token-budget-aware client) fails *only* on these tools, with no hint that the param is simply not wired here. This is the mirror image of the #515/#529/#540 facade-doc drift family: there the doc named params the inner lacked; here the inner lacks a param the whole rest of the surface honors. The #515 contract test (facade-doc ⊆ inner) won't catch it because output_format isn't in the facade doc per-action — it's cross-cutting.

## Suggested fix
Make `output_format` a universally-accepted param at the facade/base layer (accept-and-ignore where a tool has only one output shape), or whitelist it in `enforce_strict_params` for every action. Add a contract test: every facade action accepts `output_format` without raising. Concrete instances today: modification_guard, batch_search.

aimasteracc:
Closes #651.

## Problem
`output_format` is the universal TOON-vs-JSON envelope control — accepted by search/nav/structure/health/index/viz actions across hundreds of probes. But the strict-param guard (`enforce_strict_params`) rejected it on tools whose schema omits it (`modification_guard`, `batch_search`):

```
edit action=guard {..., output_format: json} → ValueError: unknown parameter 'output_format'
```

An agent that sets `output_format` uniformly (the reasonable default for a token-budget-aware client) fails *only* on these tools, with no hint the param is simply unwired there — the mirror image of the #515/#529/#540 facade-doc-drift family.

## Fix
Whitelist `output_format` at the single choke point (`enforce_strict_params`) via a `_UNIVERSAL_ENVELOPE_PARAMS` set. Every tool now accepts it — formatting tools honour it, single-shape tools accept-and-ignore — while genuinely-unknown params still raise with the did-you-mean hint. One central fix covers all 25+ tools consistently.

## RED-first
`TestUniversalOutputFormatParam`: `output_format` accepted when the schema omits it (the modification_guard case), a bogus param still rejected, and a per-tool contract asserting every BaseMCPTool accepts `output_format` without raising. The first test fails without the fix; verified `modification_guard` accepts it end-to-end.

### DIFF
--- a/tests/unit/mcp/test_base_mcp_tool_contract.py
+++ b/tests/unit/mcp/test_base_mcp_tool_contract.py
@@ -141,3 +141,44 @@ class TestBaseMCPToolContract:
             f"{_tool_id(tool)}: output_format.default must be 'toon', got {default!r}. "
             "See CLAUDE.md §1: MCP defaults to TOON — LOCKED."
         )
+
+
+class TestUniversalOutputFormatParam:
+    """#651: output_format is a universal envelope param. enforce_strict_params
+    must accept it even when a tool's schema omits it (modification_guard and
+    batch_search rejected it with ValueError), so an agent can set it uniformly
+    across every call — while still rejecting genuinely-unknown keys."""
+
+    def test_output_format_accepted_when_schema_omits_it(self) -> None:
+        from tree_sitter_analyzer.mcp.utils.schema_strictness import (
+            enforce_strict_params,
+        )
+
+        # modification_guard's schema (no output_format) — must NOT raise.
+        schema = {
+            "properties": {"symbol": {}, "modification_type": {}, "file_path": {}}
+        }
+        enforce_strict_params(
+            "modification_guard",
+            schema,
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #680, opened 2026-06-13
fix(toon): inline-key fallback for divergent nested-dict cells (#643)

### LINKED ISSUE #643: [P3] TOON nested-dict cell values positionally ambiguous when subkeys diverge from the annotation sample
Residual noted in #637's fix (pre-existing, applies to homogeneous tables too): nested-dict CELL values encode values-only `(v1,v2)` against the header's subkey annotation sampled from one row — rows whose nested dict has different subkeys than the sample are positionally ambiguous. The #637 losslessness invariant covers row keys + leaf values; a sub-schema (union at the nested level or inline fallback for divergent cells) would close this. Low priority: requires divergent nested shapes, rarer than the row-level case.

aimasteracc:
Closes #643.

## Problem
A nested-dict table column annotates its header from one sampled row (`meta{x,y}`), then encoded **every** row's dict cell as values-only `(v1,v2)`. A row whose dict had different subkeys than the sample was positionally mis-read:

```
rows: [{meta:{x:1,y:2}}, {meta:{x:3,z:9}}]
header: meta{x,y}   (sampled from row 0)
row 1:  (3,9)       ← reads as y=9 under meta{x,y}; the 'z' key is lost, 9 mis-attributed
```

The #637 losslessness invariant covered row keys + leaf values but not this nested sub-schema divergence (noted as a residual in #637).

## Fix
Encode a dict cell compactly (values-only `(v1,v2)`) **only** when its keys match the column's header sample; cells with divergent / extra / missing subkeys fall back to self-describing inline `(k1:v1,k2:v2)`:

```
row 1: (x:3,z:9)    ← self-describing, lossless
```

The compact common case (and its token savings) is unchanged — only genuinely-divergent cells pay the inline cost. Fully contained in `_toon_encoder_table_helpers.py`.

## RED-first
- `test_issue_643_divergent_nested_subkeys_encode_inline` — divergent `{x,z}` → inline (fails without the fix).
- `test_issue_643_matching_nested_subkeys_stay_compact` — matching `{x,y}` → compact (regression guard).
- 100 passed across the TOON encoder + #637 losslessness suites (no regression).

### DIFF
--- a/tests/unit/formatters/test_toon_heterogeneous_637.py
+++ b/tests/unit/formatters/test_toon_heterogeneous_637.py
@@ -56,6 +56,24 @@ class TestHeterogeneousArrayLosslessness:
         )
         assert out == "rows:\n  [2]{a,meta{x,y}}:\n    1,\n    2,(1,2)"
 
+    def test_issue_643_divergent_nested_subkeys_encode_inline(self) -> None:
+        """#643: a dict cell whose subkeys DIVERGE from the header sample must
+        encode self-describing ``(k:v)`` — not values-only ``(v1,v2)`` that
+        would be positionally mis-read. Header samples ``meta{x,y}`` from row 0;
+        row 1's ``{x,z}`` cannot reuse that annotation without losing ``z``."""
+        out = ToonEncoder().encode(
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #681, opened 2026-06-13
fix(project): smart action doc matches inner schema — file_path not task (#573)

### LINKED ISSUE #573: [P1][QC-battle] project smart documented param 'task' doesn't exist — doc-following calls fail hard
## [QC-battle] project smart: documented param `task` doesn't exist — every doc-following call fails hard
Develop @ 9fffb7d7. Facade doc: "action=smart — S2 task-focused context... Params: **task**, limit". Inner SmartContextTool: `required: ["file_path"]`, no `task` property. `_project_args` strips `task` → `ValueError('file_path is required')` on every documented invocation. With file_path it runs (task silently dropped — so the "task-focused" pitch isn't implemented on this path at all). Worst of the #515/#529/#540 facade-drift family because the documented call CANNOT work. Fix: either implement task→relevance (the documented behavior) or redocument to file_path and update the smart pitch. The #515 contract test (facade-doc params ⊆ inner schema) catches this mechanically — please ship it; this is the 4th instance.

aimasteracc:
Closes #573.

## Problem (P1 — documented call cannot work)
The `project` facade documented `action=smart` as *"S2 task-focused context… Params: **task**, limit"*, but the inner `SmartContextTool` requires `file_path` and has neither `task` nor `limit`. So every doc-following invocation hard-failed:

```
project action=smart task='…' → ValueError: file_path is required
```

The worst of the facade-doc-drift family (#515/#529/#540) because the documented call **cannot** succeed.

## Fix
`SmartContextTool` is a one-shot **file** orientation (file_health + exported API + upstream/downstream deps + test files + edit-risk for a `file_path`), not task-focused. Re-document the action to match what the code does — `Params: file_path`. (Implementing `task`→relevance would be a separate feature, not a doc fix.)

## RED-first
`test_smart_action_documented_params_subset_of_inner_schema` parses the documented `Params:` and asserts they're a subset of the inner schema (+ universal `output_format`). Fails on the old `task`/`limit` doc; verified the `file_path` call now succeeds end-to-end.

## Follow-up noted
An audit of all 8 facades' `Params:` prose vs inner schemas surfaced a few more candidates beyond regex noise (e.g. project actions documenting `format` where the inner has `output_format`; `index` auto/sync/build param names). A systematic all-facade param-doc contract test needs the freeform `Params:` prose normalized first — recommended as a focused follow-up rather than bundled here.


aimasteracc:
@codex review

PR #682, opened 2026-06-13
fix(structure): surface tree-sitter parse errors in analyze (#572)

### LINKED ISSUE #572: [P1][QC-battle] root.has_error never surfaced — wrong-language files yield success:true + phantom symbols
## [QC-battle] tree-sitter root.has_error is never surfaced — wrong-language/corrupt files return success:true with phantom symbols
Develop @ 9fffb7d7, four manifestations of one root cause (TSA never reads `root.has_error` / never emits parse_errors):
1. C++ source in a `.py` file → `success: true`, phantom class named **"public"**, zero warnings. Agent builds inferences on garbage.
2. Non-UTF8 bytes (\xff\xfe\x80) → silently processed, no encoding flag.
3. NUL byte mid-function → end_line wrong ([1,1] vs [1,3]), no warning.
4. Invalid-syntax file (emoji-only identifiers) → `success: true, methods: [], verdict: INFO` — indistinguishable from a legitimately empty file.
Fix: after parse, if `root.has_error`, add `parse_errors: true` + verdict WARN + error-node count; same for decode-replacement. The audit memory's "MCP掩盖success:False" lesson, one level deeper: now success itself is the mask.

aimasteracc:
Closes #572 (manifestations #1 + #4).

## Problem (P1 — trust)
A wrong-language or corrupt file parsed *clean*: `structure analyze` on C++ source in a `.py` returned `success:true, verdict:INFO`, a **phantom class named `public`**, and zero warnings — so an agent builds inferences on garbage. tree-sitter's `root.has_error` was never surfaced; **success itself was the mask** (the audit memory's 'MCP掩盖success:False' lesson, one level deeper).

## Fix
After a successful structure analysis, check parse validity via the existing `is_file_parse_broken` helper — which reuses the **cached** `Parser` (LRU), so this is a cache hit, not a fresh parse. When the tree has ERROR nodes: set `parse_errors: true` + `verdict: WARN` + an `agent_summary.next_step` warning the symbols may be phantom. A clean parse leaves the INFO envelope untouched (no new key).

## Scope
Covers #572 manifestations **#1 (wrong-language)** and **#4 (invalid syntax)** — the core 'phantom symbols look valid' trust bug. Manifestations **#2 (non-UTF8 encoding)** and **#3 (NUL-byte end_line)** have distinct nuances (the null-byte false-positive guard *intentionally* excludes #3 — null bytes are legal in string literals) and are noted as follow-ups rather than forced into this fix.

## RED-first
`TestAnalyzeSurfacesParseErrors`: C++-in-.py → `parse_errors:true` + `verdict:WARN` (fails without the fix); a valid file → `INFO`, no `parse_errors` key (regression). 82 passed across the analyze + agent-contracts suites (no regression).

### DIFF
--- a/tests/unit/mcp/test_analyze_code_structure_tool_core.py
+++ b/tests/unit/mcp/test_analyze_code_structure_tool_core.py
@@ -214,3 +214,40 @@ class TestAnalyzeCodeStructureToolValidateArguments:
         arguments = {"file_path": "test.py", "suppress_output": "true"}
         with pytest.raises(ValueError, match="suppress_output must be a boolean"):
             tool.validate_arguments(arguments)
+
+
+@pytest.mark.asyncio
+class TestAnalyzeSurfacesParseErrors:
+    """#572: a wrong-language / corrupt file must not be reported as a clean
+    INFO with phantom symbols. tree-sitter's parse error is surfaced as
+    parse_errors + verdict WARN so an agent doesn't trust the garbage."""
+
+    async def test_wrong_language_file_flags_parse_errors(self, tmp_path):
+        # C++ source in a .py file parses with errors (phantom 'public' class).
+        f = tmp_path / "evil.py"
+        f.write_text("class Foo { public: int x; void bar() {} };\n")
+        tool = AnalyzeCodeStructureTool(project_root=str(tmp_path))
+        result = await tool.execute({"file_path": str(f), "output_format": "json"})
+        assert result["parse_errors"] is True
+        assert result["verdict"] == "WARN"
+        assert result["agent_summary"]["verdict"] == "WARN"
+
+    async def test_valid_file_has_no_parse_errors_flag(self, tmp_path):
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

aimasteracc:
## Codex P1 triage: REAL — fixed in 4d2d4ed1

Correct and important catch. The check used `options.file_path` (raw) while the analysis uses `options.resolved_path`. With an MCP `project_root` + a project-relative `file_path`, `is_file_parse_broken`'s `Path(file_path).is_file()` returns False from the server cwd → the check silently skipped and #572 still reported phantom symbols as clean — exactly the common MCP shape.

Switched to `options.resolved_path`. Added `test_relative_file_path_with_project_root_still_flags` (relative `file_path` + project_root) and **mutation-verified** the resolved-path line specifically fixes it: reverting to `file_path` fails the relative test while the absolute test still passes.

@codex review

PR #581, opened 2026-06-12
fix(params): default-valued parameters extracted in JS/Python

aimasteracc:
Closes #533 (P2, QC-sweep — Theme E residual)

## Root causes (AST-dump confirmed)

- **Python**: `breed: str = \"Mixed\"` parses as `typed_default_parameter` — missing from `_PARAMETER_NODE_TYPES` (untyped `x=5` → `default_parameter` was already there, hence the typed-only drops)
- **JS**: `limit = 10` parses as `assignment_pattern` in formal_parameters — unhandled branch in `_extract_parameters` (destructuring/rest had branches; plain defaults didn't)
- **Bash**: the grammar has NO formal-param nodes (functions use $1/$2) — `parameters=[]` is structurally correct; contract pinned in 2 tests, documented n/a

Convention: emit full node text (the C# precedent) — `limit = 10`, `breed: str = \"Mixed\"`.

## Verification

- 10 RED→GREEN + 4 Bash contract pins; 120 JS + 118 Python plugin tests green
- Python goldens (full/compact/csv) updated — diffs are exactly the two `__init__` signatures gaining their defaults; JS goldens unaffected (fixture has no defaults; verified by run)
- Lead independently verified Sample.py table shows `breed:str = \"Mixed\"` / `indoor:bool = True`
- Coverage: `_function_mixin` 95.60%; ratchet exit 0

PR #684, opened 2026-06-13
fix(structure): dedup param parsing so compound generics aren't inverted (#576)

### LINKED ISSUE #576: [P2][QC-battle] structure analyze inverts param name/type on dict[str, Any]-style annotations
## [QC-battle] structure analyze inverts param name/type on compound annotations: dict[str, Any] → name="Any]"
Develop @ 9fffb7d7. `get_method_parameters()` (analyze_code_structure_helpers.py:78-80) whitespace-splits the param string: `result: dict[str, Any]` → parts ['result:','dict[str,','Any]'] → name=`Any]`, type=`result: dict[str,`. Every Python method using a generic with a space (dict[...], Optional[...], Callable[...]) is misrepresented — that's most modern typed Python. Observed live on reduce_to_control_surface. Fix: split on the first ':' instead; test with dict[str, Any], Callable[[int], str], default values.

aimasteracc:
Closes #576.

## Problem (P2)
`structure analyze` carried a **duplicate** `_parse_string_param` that whitespace-split the parameter string:

```
'result: dict[str, Any]' → ['result:', 'dict[str,', 'Any]']
                         → name='Any]', type='result: dict[str,'
```

Every modern typed-Python param with a space in its generic (`dict[...]`, `Callable[...]`, `Optional[...]`) was inverted.

## Root cause — duplicate-logic drift
The shared helper `_parse_string_parameter` was **already fixed** (splits on the first `:`, handles defaults — the #581 work). The bug survived only in this **dead duplicate** in `analyze_code_structure_tool.py` (reachable via `_get_method_parameters`, which production no longer calls — the live path uses the helper's `get_method_parameters`). So #576's live impact was resolved, but the duplicate was a latent landmine.

## Fix
Delegate the duplicate to the helper — single source of truth, no more drift. (A fuller refactor-cleaner pass could delete the vestigial `_get_method_parameters` cluster entirely; noted as follow-up.)

## RED-first
`test_get_method_parameters_compound_generic_types` pins `dict[str, Any]` / `Callable[[int], str]` / `list[tuple[int, str]]` / default-valued forms; fails against the old whitespace split.

### DIFF
--- a/tests/unit/mcp/test_analyze_code_structure_tool_helpers.py
+++ b/tests/unit/mcp/test_analyze_code_structure_tool_helpers.py
@@ -395,3 +395,22 @@ class TestAnalyzeCodeStructureToolGetMethodParameters:
         assert result[0]["type"] == "str"
         assert result[1]["name"] == "param2"
         assert result[1]["type"] == "int"
+
+    def test_get_method_parameters_compound_generic_types(self):
+        """#576: ``name: <generic with spaces>`` must split on the FIRST ':',
+        not whitespace — otherwise ``result: dict[str, Any]`` parsed as
+        name='Any]', type='result: dict[str,'. Covers the modern typed-Python
+        forms (generics, nested brackets, defaults) that whitespace-splitting
+        mangled."""
+        mock_method = MagicMock()
+        mock_method.parameters = [
+            "result: dict[str, Any]",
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

Issue #529, opened 2026-06-12
[P3][dogfood] ast_diff param contract: facade doc names nonexistent before/after; mode required-with-default; error hides diff_git


aimasteracc:
## Summary
`edit action=ast_diff` has a triple param-contract problem (same drift family as #515, error-quality family as #449):

1. **Facade doc names params that don't exist**: the facade describes "Params: file_path, **before, after** or git ref params" — the inner schema has no `before`/`after`; it has old_file/new_file (diff_files), old_source/new_source (diff_strings), old_ref/new_ref+file_path (diff_git).
2. **`mode` is `required` yet documented with a default**: schema says `"required": ["mode"]` while the description says "(default: diff_files)" — and in practice omitting mode validated as diff_files and rejected my valid diff_git args.
3. **Error doesn't enumerate the contract**: passing `file_path+old_ref+new_ref` (a perfectly sensible git-diff request) fails with "old_file and new_file are required for diff_files mode" — no mention that `mode=diff_git` exists and matches exactly the args I passed.

## Repro (develop @ 9fb13638)
```
ast_diff file_path=...py old_ref=HEAD~3 new_ref=HEAD
→ ValueError: old_file and new_file are required for diff_files mode
ast_diff mode=diff_git file_path=...py old_ref=HEAD~3 new_ref=HEAD   (works, presumably)
```

## Suggested fix
Infer mode from the args provided (old_ref/new_ref present → diff_git; old_source → diff_strings), keep `mode` optional; fix the facade description to the real param names; on mismatch, error with the three mode signatures. Contract test from #515 (facade-doc params ⊆ inner schema) would have caught leg 1 mechanically.

aimasteracc:
已由 #551 修复并合入 develop(手动关闭)。三腿:mode 从参数形状推断(sources > files > refs+file_path 判别子,CLI bridge 的默认 refs 不会偷推断——Codex P2 加固);schema required 摘除 mode;facade 描述列三模式真签名;空签名错误枚举全部模式。

aimasteracc:
**Retested on develop @ 46d212eb (#551) — CONFIRMED FIXED ✓** `file_path + old_ref/new_ref` without `mode` now infers diff_git and succeeds. Contract legs all addressed. (Size observation moved to a separate issue — ast_diff returns 155KB for the same small diff; the #542/#543 budget pattern hasn't reached it.) Good to close.

PR #686, opened 2026-06-13
fix(mcp): stop silent wrong-mode fallbacks in cache + imports (#575)

### LINKED ISSUE #575: [P2][QC-battle] Silent wrong-mode fallbacks: cache symbol=→stats; imports ignores file_path
## [QC-battle] Silent wrong-mode fallbacks: index cache symbol=→stats; health imports ignores file_path when mode omitted
Develop @ 9fffb7d7:
1. `index action=cache symbol="X"` — `symbol` is the facade's own core param but the inner ASTCacheTool only knows `query`; symbol is stripped, mode falls to `stats`, and the agent gets index statistics instead of symbol results. No diagnostic. (R3 normalization maps symbol→function_name only — needs symbol→query for this action.)
2. `health action=imports file_path=...` without mode= — inner declares `required:["mode"]` yet execute() defaults to summary, silently ignoring file_path. Misleading result + a required-with-default contradiction (cf. #529's mode lesson).
Both are "answered a different question than asked" failures — worse than errors. Fix: map symbol→query for cache; infer file_deps mode when file_path present (or error); reconcile required vs default.

aimasteracc:
Closes #575.

Two *'answered a different question than asked'* failures — worse than errors because the agent gets a confident wrong answer with no diagnostic.

## 1. `index action=cache symbol=X` → stats instead of symbol search
The facade whitelisted `symbol` away (the inner ASTCacheTool only declared `query`), so `mode` fell to `stats` and the agent got **index statistics** instead of symbol results. Fix: ASTCacheTool declares `symbol` and aliases it to `query` (the facade's canonical id) — `cache symbol=X` now searches (`mode=search`); an explicit `query` still wins.

## 2. `health action=imports file_path=X` (no mode) → summary, ignoring file_path
`execute()` defaulted to `summary` (which ignores file_path), and `required:['mode']` contradicted the `default:'summary'`. Fix: `_effective_mode` infers `deps` (what the file imports) when `file_path` is present, else `summary`; explicit mode wins; `mode` removed from `required` (it has a default + inference — the #529 required-with-default lesson).

## RED-first
- `TestSymbolAliasesQuery` — cache `symbol`→search, explicit `query` wins.
- `TestModeInferenceFromFilePath` — file_path→deps, none→summary, explicit wins, mode not required.

Both fail without the fix; 222 tests green across the cache/import_graph/contracts suites (the `required` change is safe).

### DIFF
--- a/docs/api/facade-actions.md
+++ b/docs/api/facade-actions.md
@@ -67,7 +67,7 @@ Reading the tables:
 | `deps` | `file_path`, `mode`, `output_format` | `success`*, `verdict`*, `agent_summary`, `error` + action payload | `--dependencies` |
 | `file` | `file_path`*, `compact_only`, `language`, `output_format` | `success`*, `verdict`*, `agent_summary`, `error` + action payload | `--file-health` |
 | `heatmap` | `directory`, `file_path`, `function_name` (`symbol` aliases `function_name`), `language`, `max_files`, `mode`, `output_format` | `success`*, `verdict`*, `agent_summary`, `error` + action payload | `--codegraph-complexity-heatmap` |
-| `imports` | `mode`*, `file_path`, `max_depth`, `output_format` | `success`*, `verdict`*, `agent_summary`, `error` + action payload | `--import-graph` |
+| `imports` | `file_path`, `max_depth`, `mode`, `output_format` | `success`*, `verdict`*, `agent_summary`, `error` + action payload | `--import-graph` |
 | `matrix` | `mode`*, `file_path`, `output_format`, `threshold`, `top_k` | `success`*, `verdict`*, `agent_summary`, `error` + action payload | `--dependency-matrix` |
 | `overview` | `max_coupled_files`, `max_dead`, `max_entry_points`, `max_hubs`, `output_format` | `success`*, `verdict`*, `agent_summary`, `error` + action payload | `--codegraph-overview` |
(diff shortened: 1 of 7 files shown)


aimasteracc:
@codex review

aimasteracc:
## Codex P2 triage: REAL — fixed in 8d854f4a

Correct — the CLI spec (`_specs_extended.py`) passed `mode: getattr(args, 'import_graph_mode', 'summary') or 'summary'`, forcing 'summary' so `--import-graph-file` without `--import-graph-mode` ignored the file (the exact scenario the MCP fix addresses) — breaking parity. Now it passes `getattr(args, 'import_graph_mode', None)` (the user's mode, else None), so the tool's `_effective_mode` infers 'deps' from file_path on the CLI path too. Verified: `mode=None + file_path → deps`.

Also regenerated `docs/api/facade-actions.md` (it drifted from the schema changes — the 5 CI failures; mechanical `generate_facade_actions_doc.py`).

@codex review

aimasteracc:
## Codex P2 round-3 triage: REAL — fixed in 49844134

Deeper than my round-2 fix caught: argparse defaulted `--import-graph-mode` to `'summary'` (`_analysis_graph_nav.py:406`), so `args.import_graph_mode` was `'summary'` even when omitted — my `getattr(..., None)` never saw None. Changed the parser default to `None`, so an omitted flag now yields None → the spec forwards None → `_effective_mode` infers 'deps' from file_path. Verified end-to-end through `create_argument_parser`: omitted → None, explicit → 'cycles'. No test pinned the old default.

@codex review

PR #687, opened 2026-06-13
fix(search): chain DSL rejects '|' loudly + doc params match inner (#574)

### LINKED ISSUE #574: [P2][QC-battle] chain DSL: pipe separator silently mis-parses; facade params all drift
## [QC-battle] chain DSL: `|` pipe silently mis-parses (correct separator is `.`); facade params all drift
Develop @ 9fffb7d7:
1. Facade sells action=chain as "jQuery-style compose search → explore → callers" with NO syntax example. The natural pipe `A | B` doesn't error — it falls to the no-paren fallback and runs `explore("A | B").related()`: wrong result, zero diagnostics. Real separator: `.`.
2. Facade doc params `chain/program, default_limit, include_source` — inner accepts `query, max_symbols, max_files, include_code, compact`. None of the documented names exist; the required `query` is undocumented.
Fix: syntax example in the description, reject unparseable chains loudly (or warn when falling back), align param docs. (#515 drift family, 5th instance.)

aimasteracc:
Closes #574.

Two facade-drift failures on `action=chain`:

## 1. `|` pipe separator silently mis-parsed
`query="A | B"` has no parens, so it fell to the no-paren fallback `explore("A | B").related()` — a **wrong result with success:true and zero diagnostics**. `parse_chain` now rejects `|` with the real syntax (`steps are separated by '.', e.g. search('X').callers()`). Identifiers never contain `|`, so this is unambiguous; the legitimate plain-string fallback is preserved.

## 2. Doc params all wrong (#515 family, 5th instance)
The doc advertised `chain/program, default_limit, include_source` — **none** of which `CodeGraphQueryTool` accepts (`query` (required), `max_symbols`, `max_files`, `include_code`, `compact`). Aligned the doc to the inner **and added a `.`-separator syntax example** — the missing example was the root cause of agents reaching for `|`.

## RED-first
`test_pipe_separator_rejected_loudly` + `test_chain_documented_params_subset_of_inner_schema`; both fail without the fix. 62 DSL/query tests green (the early `|` check breaks no existing chain).


aimasteracc:
@codex review

aimasteracc:
## Codex P2 triage: REAL — fixed in 45377c27

Correct — my first cut rejected EVERY '|', which would regress legitimate searches for code containing a pipe (`search('string | number')`, TS unions, type hints). Replaced the unconditional check with `_has_unquoted_pipe` — a quote-aware scan that flags a '|' only OUTSIDE single/double-quoted spans (separator position). Verified: `.function | .method` → rejected; `search('string | number')` → allowed (args == ['string | number']). Added `test_quoted_pipe_is_legitimate_search_text`; error message now points to quoting for literal pipes.

@codex review

PR #688, opened 2026-06-13
fix(nav): lineage returns real inheritance hierarchy for classes (#568)

### LINKED ISSUE #568: [P2][dogfood] nav lineage: no inheritance content (impact profile instead) + refs=0 for 74-subclass base class
## nav action=lineage returns an impact profile, not lineage — zero inheritance content for a 74-subclass base class
Develop @ f443c246, full fresh index:
```
nav action=lineage symbol=BaseMCPTool
→ keys: definitions, references(0!), downstream_files(449), test_files_to_run, risk … 
→ NO subclasses / superclasses / overrides / hierarchy key at all
```
The facade schema sells this action as "class/function **inheritance and override lineage**". `BaseMCPTool` has ~74 subclasses in the index (extends edges exist — index reports 746). Two defects:
1. The advertised content is absent — an agent asking "what's the lineage of this class" gets a change-impact profile instead. Either rename/redocument the action (impact_profile?) or join the extends edges into a real lineage tree (structure action=class_tree already proves the data exists).
2. `reference_count: 0` next to `downstream_file_count: 449` is internally contradictory for the most-inherited class in the repo — refs=0 will mislead any consumer that checks references before renaming.

aimasteracc:
Closes #568 (defect 1; defect 2 is a follow-up).

## Problem (P2)
`nav action=lineage` advertises *'class/function inheritance and override lineage'* but returned only an impact profile (definitions, references, downstream files) — **zero inheritance content**, even for a base class with ~74 subclasses whose extends edges exist in the index (`structure action=class_tree` already proves the data is there).

## Fix
When the symbol is a class, join the extends edges via `ClassHierarchy` and add a `hierarchy` section (`subclasses` + `subclass_count` + `subclasses_truncated` + `superclasses` + `superclass_count`), reusing the same computation `class_tree` uses. **Guarded on a class definition** so non-class symbols skip the hierarchy build entirely and carry no `hierarchy` key. The impact-profile fields are unchanged — additive, closing the doc-vs-impl gap.

## RED-first
`TestInheritanceLineage` — a base class with 2 subclasses returns `subclass_count == 2` / `subclasses == [Sub1, Sub2]`; a function symbol returns no `hierarchy` key. Fails without the fix (KeyError). 31 lineage tests green.

## Follow-up (defect 2)
`reference_count: 0` next to `downstream_file_count: 449` for the most-inherited class is a separate references-computation issue (refs likely miss inheritance edges) — noted for a focused follow-up rather than bundled here.

### DIFF
--- a/tests/unit/mcp/test_symbol_lineage_tool.py
+++ b/tests/unit/mcp/test_symbol_lineage_tool.py
@@ -169,6 +169,94 @@ class TestExecute:
             asyncio.run(t.execute({"symbol": "foo"}))
 
 
+class TestInheritanceLineage:
+    """#568: lineage of a class returns the advertised inheritance/override
+    content (subclasses/superclasses via extends edges), not only an impact
+    profile. Non-class symbols carry no hierarchy section."""
+
+    def test_class_symbol_returns_inheritance_hierarchy(self, tool, tmp_path):
+        from tree_sitter_analyzer.ast_cache import ASTCache
+
+        _write_py(
+            tmp_path, "base.py", "class Base:\n    def run(self):\n        pass\n"
+        )
+        _write_py(
+            tmp_path,
+            "sub.py",
+            "from base import Base\n\nclass Sub1(Base):\n    pass\n\n"
+            "class Sub2(Base):\n    pass\n",
+        )
+        ASTCache(str(tmp_path)).index_project(max_files=50)
+        result = asyncio.run(tool.execute({"symbol": "Base", "output_format": "json"}))
+        hier = result["hierarchy"]
+        assert hier["subclass_count"] == 2
+        assert sorted(s["name"] for s in hier["subclasses"]) == ["Sub1", "Sub2"]
+        # Second call with the index unchanged: result is stable (the cache is
+        # not spuriously invalidated when the AST index mtime is the same).
+        again = asyncio.run(tool.execute({"symbol": "Base", "output_format": "json"}))
+        assert again["hierarchy"]["subclass_count"] == 2
+
+    def test_function_symbol_has_no_hierarchy(self, tool, tmp_path):
+        from tree_sitter_analyzer.ast_cache import ASTCache
+
+        _write_py(tmp_path, "m.py", "def standalone():\n    return 1\n")
+        ASTCache(str(tmp_path)).index_project(max_files=50)
+        result = asyncio.run(
+            tool.execute({"symbol": "standalone", "output_format": "json"})
+        )
+        assert "hierarchy" not in result
+
+    def test_qualified_symbol_name_resolves_hierarchy(self, tool, tmp_path):
+        # Codex P2: a qualified symbol (pkg.Base) must resolve via its bare name
+        # — ClassHierarchy stores classes by bare name from the AST cache.
+        from tree_sitter_analyzer.ast_cache import ASTCache
+
+        _write_py(tmp_path, "base.py", "class Base:\n    pass\n")
+        _write_py(
+            tmp_path, "sub.py", "from base import Base\n\nclass Sub(Base):\n    pass\n"
+        )
+        ASTCache(str(tmp_path)).index_project(max_files=50)
+        result = asyncio.run(
+            tool.execute({"symbol": "base.Base", "output_format": "json"})
+        )
+        assert result["hierarchy"]["subclass_count"] == 1
+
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

aimasteracc:
## Codex triage — all 3 P2s REAL, fixed in 80aec3ed (+ codecov)

All three shared one root: my guard relied on the definition's kind/type instead of asking ClassHierarchy authoritatively. Redesigned `_hierarchy_for` to use `has_class(bare_name)` as the gate:

1. **Qualified name (line 447)** ✅ — strip to bare name (`pkg.Base`→`Base`) before `has_class`; ClassHierarchy stores by bare name. Test: `base.Base` resolves the hierarchy.
2. **grep-fallback class defs (line 438)** ✅ — detection no longer depends on definition `kind`/`type`; `has_class` checks the AST cache directly, so a class found only via the K3 grep fallback still gets its hierarchy.
3. **cache invalidation (line 326)** ✅ — added `_invalidate_symbol_cache_on_index_change`: the per-symbol cache now clears when `.ast-cache/index.db` mtime changes, so a lineage call made before indexing no longer serves a stale no-hierarchy hit after the index is built. Test: call before index → no hierarchy; build index → re-query returns it.

codecov/patch: covered the new branches (qualified name, index-change invalidation, error-degrades-to-None); removed dead `project_root` guards (caller validates).

@codex review

aimasteracc:
## Round-3 triage — Codex WAL/SHM P2 + a self-inflicted cache regression

**Codex P2 (index.db mtime misses SQLite WAL writes) — VALID, fixed.**
`_invalidate_symbol_cache_on_index_change` statted only `.ast-cache/index.db`. Confirmed empirically that `index_project()` leaves `index.db-wal` with a *later* mtime than `index.db` (`...294105602` > `...279178509`), so a WAL-only write was invisible to the stale check. Fixed via new `_index_signature()` = max `st_mtime_ns` across `index.db`, `index.db-wal`, `index.db-shm`.

**Self-inflicted regression also fixed (root cause of the 5 red CI checks):** the previous commit recorded the index signature at the *start* of `execute`, but `find_references`/`_hierarchy_for` *create* `.ast-cache` as a side effect during the call. So the cold call recorded mtime=0, and the very next warm call saw the freshly-created index and spuriously wiped the per-symbol cache → `from_cache` came back False, breaking `test_symbol_lineage_cache_invalidates_on_file_change`. Fixed by recording the signature in `_finalize_and_cache_response` (AFTER this call's own index reads/writes), so a call that creates the index no longer evicts the next warm hit.

Both root causes have green coverage: `test_cache_invalidates_when_index_built_after_first_call` (index built post-first-call → refresh) and `test_symbol_lineage_cache_invalidates_on_file_change` (warm hit survives, source-modify still invalidates). Full lineage + graph-cache suites: 43 passed; ruff + mypy clean.

@codex review

aimasteracc:
## Post-merge Codex triage (4 P2s across iterations)

| # | commit | finding | verdict |
|---|--------|---------|---------|
| 1 | 17e3a4f0 | invalidate cache when hierarchy built after a no-hierarchy call | **Fixed** — `_invalidate_symbol_cache_on_index_change` + index-signature (fdf17cf0) |
| 2 | 17e3a4f0 | grep-fallback class defs (`type=definition`, no `kind`) skip hierarchy | **Already handled** — `_hierarchy_for` gates on `ClassHierarchy.has_class(bare)`, never on `kind`/`type`; a grep-only def still gets its hierarchy. Covered by the docstring + `test_hierarchy_degrades_to_none_on_error`. |
| 3 | 80aec3ed | include WAL/SHM mtimes in invalidation | **Fixed** — `_index_signature` takes max `st_mtime_ns` across `index.db`/`-wal`/`-shm` (f9e043af); empirically the WAL sidecar carries a later mtime than index.db. |
| 4 | fdf17cf0 | stale hierarchy rows after source edits without re-index | **Real-but-minor → follow-up filed.** The in-process response cache IS correctly invalidated, but the `hierarchy` block reads persisted `ast_index` rows while `references` re-scans live source — so a deleted subclass can still appear under the base until re-index, an inconsistency *within one envelope*. This matches the freshness semantics of every index-derived query (`structure action=class_tree` has the same lag), so it is not a regression, but the live-vs-index mix is a genuine footgun. Tracking as a focused enhancement (flag hierarchy staleness when max source mtime > index mtime). |

#1/#3 fixed in this PR; #2 already correct by design; #4 → follow-up issue.

PR #543, opened 2026-06-12
fix(mcp): classify returns verdict+summary by default, bulk payload opt-in with budget

aimasteracc:
Closes #528

## Diagnosis (per-key byte breakdown)

The 1.29MB response was 99.98% one field: `classifications` — each `ClassifiedHunk.to_dict()` embedded full `ASTNodeInfo` subtrees (up to 267 nodes, depth 12) for BOTH sides of every hunk. A 10KB diff serialized to 1.95MB (182×).

## Fix (honest-truncation convention)

- Default: verdict + category + per-hunk classification summary (labels + line ranges), **no AST children** — `include_ast_nodes=false`
- `hunk_cap` (default 50) caps the classifications list with truncated/listed_cap/next_step honesty fields
- Both params declared in the inner schema, neither in `required` (facade trap avoided)
- **Differential invariant tests** (rule 11.5): string-mode response ≤ 10× raw diff (documented — semantic labels add per-hunk overhead; pre-fix was 120×, post-fix 5.4×); git-mode test on a real 2-commit fixture asserts response ≤ raw diff bytes

## Measurements

| Repro | Before | After |
|---|---|---|
| 10KB diff (outline tool, HEAD~3) | 1,954,775 B | 24,172 B (**−98.8%**) |
| Lead independent repro (explore tool, HEAD~5) | 1.29MB class | 24,054 B default / 904KB opt-in |

## Test plan

- [x] 6 RED→GREEN new tests + 32 total green (`-n 0` narrow)
- [x] Coverage 94.67% on the tool (remaining misses are pre-existing defensive branches)
- [x] Loose-assertion ratchet exits 0

PR #542, opened 2026-06-12
feat(mcp): byte budget + honest truncation for structure/outline surface

aimasteracc:
Part of #513 (leg 3 of 3, per the [lead triage](https://github.com/aimasteracc/tree-sitter-analyzer/issues/513#issuecomment-4686371597)).

## Problem (dogfood P2)

`structure action=outline` on giant files returned unbounded payloads (250KB on vscode.d.ts; ~2.3MB at 500-class synthetic scale) with no cap or sizing note — blowing agent context windows. The DF-1/DF-13 budget work never reached this surface.

## Fix (honest-truncation convention, #486/#489/#500/#505)

`GetCodeOutlineTool` now applies a default `listed_cap=50` to the classes + top-level-functions lists, with: pre-cap totals (`classes_total`, `top_level_functions_total`), `truncated`, `listed_cap` echo, and a `next_step` narrowing hint (steering giant-file consumers to the signatures view, −80% on .d.ts per #541). Aggregate totals are computed over ALL elements before truncation. Explicit `listed_cap` raises the limit. Deterministic ordering under the cap. CLI path unchanged (its full-output+jq workflow is user-locked); TOON default untouched.

## Measurements (lead-verified)

- Synthetic 60-class .d.ts: default response 112KB JSON (50/60 classes listed, truncated=true, totals correct); per-class payload 1,090 B — lean shape, no inline source
- Pod's 500-class scale test: ~2.39MB → ~237KB (10×)
- Schema: `listed_cap` declared in inner schema (facade whitelist passes it; NOT in `required` — the 6×-recurred facade trap avoided)

## Test plan

- [x] RED (`ImportError` on the new cap constant) → GREEN: 30 new tests (cap default, explicit raise, totals-before-truncation, determinism)
- [x] 46 existing outline + 35 structure facade + 33 cost-invariant tests green
- [x] Loose-assertion ratchet exits 0

Closes #513 (legs 1=#527, 2=#541, 3=this).

PR #694, opened 2026-06-14
fix(edit): ast_diff defaults to node summaries; full bodies opt-in with byte budget (#552)

### LINKED ISSUE #552: [P2][dogfood] ast_diff: 155KB for a small diff — needs the #542/#543 budget treatment
## ast_diff: 155KB response for a 5-commit single-file diff — last unbudgeted tool in the flood family
Develop @ 46d212eb (right after #551's contract fix): `ast_diff file_path=codegraph_explore_tool.py old_ref=HEAD~5 new_ref=HEAD` → **154,984 B**. Sibling tools just got the treatment — outline (#542), classify (#543, 1.29MB→15.6KB), similarity (#481), heatmap pending (#516). ast_diff presumably inlines full node payloads for every changed subtree. Same fix shape: added/removed/changed node summaries (kind, name, line span) by default; node bodies behind opt-in with a byte budget.

aimasteracc:
## Summary
Fixes #552. `edit action=ast_diff` returned **154,984 B** for a 5-commit single-file diff — 99.8% in `hunks`. Root cause: `ASTNodeInfo.to_dict()` (ast_diff.py:154-166) recursively inlined the full `children` subtree; one added/removed module-level literal dumped its entire AST (a single hunk's `children` was 60,897 B).

Mirrors the #542 (outline) / #543 (classify) budget treatment.

## Fix
- **Default**: node dicts no longer inline `children`; they carry the cheap summary fields + a new `child_count` int.
- **Opt-in**: `include_node_bodies` (default False) inlines `children` but enforces `NODE_BODIES_BUDGET = 32_768` bytes; over budget → compact form + `children_truncated=True`, `bytes_omitted=<int>` (no silent drop).
- Threaded through `to_dict(include_children=...)` on all three dataclasses; `execute()` enforces the budget.
- **CLI parity**: `--ast-diff-include-bodies` → `include_node_bodies`.
- **Facade**: param is in the inner schema, so the `edit` facade whitelist passes it through — no facade change.

## Numbers (verified by lead)
- Before: 154,984 B · After (default): **10,954 B (-93%)** · opt-in over budget: compact + `bytes_omitted=144030`.

## Tests
6 new in `TestASTDiffNodeBudget` (RED-first, exact `==`): no `children` key by default, `child_count == 5` (grammar-pinned), `include_node_bodies` inlines (`len == 5`), over-budget sets `children_truncated`, documented relationship `default_bytes < include_bodies_bytes`. 36 file tests green; no ast_diff golden masters exist.

@codex review


aimasteracc:
## Post-merge Codex triage (1 P2 — REAL, follow-up opened)
**P2 'Preserve semantic classify body opt-in' — VALID, regression introduced by this PR.** Making the shared `ASTNodeInfo.to_dict()` default `include_children=False` means `ClassifiedHunk.to_dict()` (semantic_change_classifier.py:118, calls `hunk.to_dict()` with no args) now drops children, so `semantic_classify(include_ast_nodes=True)` returns hunks WITHOUT AST nodes — the opt-in is broken. The byte-budget test only covers the default (`include_ast_nodes=False`) path, so it didn't catch this. This is exactly the 'shared serializer changed, untested consumer path broke' pattern. Fix: `ClassifiedHunk.to_dict()` → `hunk.to_dict(include_children=True)` (default path still strips via `_compact_hunk`); + RED test for the `include_ast_nodes=True` path. Follow-up PR incoming.

Issue #691, opened 2026-06-14
[P2][QC-battle] E2E gap: 0.16% of tests are e2e, 0 MCP stdio round-trips — envelope bugs ship green (273 execute() vs 8 handle_call_tool)


aimasteracc:
## Viewpoint
The suite is overwhelmingly **unit + in-process integration**; true **end-to-end** coverage of the product's primary interface — an LLM agent talking to the MCP server over stdio JSON-RPC — is effectively absent. This is structural, and it's *why* a ~18k-green suite keeps shipping agent-facing envelope bugs that the dogfood arm finds by hand.

## Evidence (develop @ 55cac0f9)
Test-function inventory:
| layer | test funcs | share |
|---|---|---|
| `tests/unit` | 16,940 | 93.6% |
| `tests/integration` | 1,136 | 6.3% |
| `tests/e2e` | **30** (3 files) | **0.16%** |

MCP testing boundary:
- **273 files** call `.execute()` **directly** — in-process, bypassing the MCP JSON-RPC envelope, dispatch, and serialization.
- only **8 files** exercise `handle_call_tool` (the real dispatch boundary).
- **0 files** drive a real `stdio_client` / `ClientSession` round-trip. (`grep -rlE 'stdio_client|ClientSession|mcp.client' tests` → empty.)
- `tests/e2e/test_mcp_smoke.py`'s "subprocess" usage is testing **tool-internal subprocess encoding**, not spawning the server.

## Why this matters (not academic)
Every recent dogfood finding lives in the layer `.execute()` tests skip and no e2e test covers:
- #689 — `health dead` summary/stat/cap incoherence (envelope-shaping layer)
- #690 — `Invalid mode` error omits valid modes (validation→envelope layer)
- the historical TOON-vs-JSON 1.96× regression (serialization layer) — found by a human, never by the suite (see CLAUDE.md §11)

These all reproduce only when you go through the *whole* path: client → server dispatch → tool → envelope/serialization → wire. The conformance suite asserts in-process `execute()` shapes; it cannot see what the agent actually receives.

## Proposal (not asking to write code now — capturing the gap)
1. A small **MCP stdio round-trip e2e harness**: spawn the real server as a subprocess, do the `initialize` handshake, `tools/list`, then `tools/call` each of the 8 facades once, and assert on the **wire envelope** the agent receives (verdict, summary_line, agent_summary, TOON/JSON shape, byte budget).
[...]

PR #701, opened 2026-06-14
docs(rfc): RFC-0017 — test effectiveness (mutation scoring + value invariants)

aimasteracc:
## Summary
RFC-0017 (status: **draft**) — answers, with a plan, the question 'why does ~2× test volume catch no real defects?'

**Diagnosis (first principles):** the suite tests **conformance** ('does the code match its spec?') and is structurally blind to **value** ('is the spec good / does the tool deliver what it advertises?'). 2× conformance = 0× discovery. The incentive (reward green + coverage%) mass-produced green conformance tests and avoided red value tests.

**Evidence (this session):** a 7-PR cascade produced **9 real P1/P2 defects** — all caught by dogfood or the Codex reviewer, **zero by the ~18k-test suite** (table in the RFC). The one value test we had (#543 byte budget) was the only suite test that caught a regression.

**Proposal (4 mechanical countermeasures):**
1. **Mutation scoring** (mutmut) on core modules — quantify whether tests *can* catch bugs (line coverage can't); ratchet phase 2.
2. **Per-tool value invariants** (output size / latency / advertised-capability / envelope *content* not just presence) as a required CI gate.
3. **Outside-the-loop authorship** — RED-first against an independent spec, test author ≠ implementer (extends dev≠review to test authorship).
4. **Boundary + all-consumer testing** for shared components (the #552→#694→#696→#698 serializer chain) + cross-surface CLI↔MCP differential.

Plus: every escaped defect lands the value invariant that would have caught it (ratchet from conformance → value).

No MCP/CLI surface change. Adds dev-deps + CI gates. Per RFC process — review (incl. Codex) before any implementation. Acceptance criteria as checkboxes in the doc.

@codex review

### DIFF
--- a/rfcs/0017-test-effectiveness-mutation-and-value-invariants.md
+++ b/rfcs/0017-test-effectiveness-mutation-and-value-invariants.md
@@ -1,0 +1,237 @@
+# RFC-0017: Test effectiveness — mutation scoring, value invariants, and outside-the-loop authorship
+
+- **Status**: accepted (owner approved 2026-06-14; phase 1 = mutation baseline first)
+- **Author(s)**: @aimasteracc
+- **Created**: 2026-06-14
+- **Last updated**: 2026-06-14
+- **Tracking issue**: TBD
+- **Affected source paths** (pin them — reviewers watch for drift here):
+  - `tests/unit/mcp/test_output_cost_invariants.py` (new + extended)
+  - `tests/effectiveness/` (new — mutation harness config + reports)
+  - `pyproject.toml` (dev-deps: `mutmut` or `cosmic-ray`)
+  - `.github/workflows/` (new gate: value-invariants required; mutation score reported)
+  - `CLAUDE.md` (codify "test value not just conformance" + "test author ≠ code author")
+
+## Summary
+
+Our test suite is ~2× the size of the source it covers, yet in a single
+session it failed to catch **nine real P1/P2 defects** — every one was found by
+an outside-the-loop signal (a human dogfooding, or the Codex reviewer), not by
+the suite. This RFC diagnoses why (the suite tests **conformance** — "does the
+code match its spec?" — and is structurally blind to **value** — "is the spec
+good / does the tool deliver what it advertises?") and proposes four mechanical
+countermeasures: (1) **mutation scoring** to quantify whether tests *can* catch
+bugs; (2) **per-tool value invariants** (output size / latency / advertised
+capability) as CI gates; (3) **outside-the-loop test authorship** (RED-first
+against an independent spec, author ≠ implementer); (4) **boundary + all-consumer
+testing** for shared components. None of these is "write more tests" — they
+change *what* tests assert and *how* discovery happens.
+
+## Motivation
+
+### The evidence (this session, 2026-06-14)
+
+A seven-PR fix cascade produced nine real defects. The suite (≈18k tests, all
+green throughout) caught **zero**:
+
+| Defect | Found by | Why the suite missed it |
+|---|---|---|
+| `ast_diff` returns 155 KB for a small diff (#552) | human dogfood | tests asserted hunks are *correct*, never that output is *small* |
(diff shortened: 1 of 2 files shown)


aimasteracc:
## QA review (adversarial) — strong RFC; 1 factual inconsistency + 3 substantive gaps

Direction is right and the diagnosis (2× conformance volume = 0× discovery) is correct. Blocking-quality nits below; none are reasons to abandon, all are reasons to tighten before `accepted`.

### 1. Factual inconsistency in the evidence — `#552` is cited as two different defects
- Evidence table row 1: **"#552 — `ast_diff` returns 155 KB — found by human dogfood — suite missed it."**
- Counter-example paragraph: **"the #543 byte-budget was the only suite test that caught a real regression this session (the #552 child_count leak)."**

So `#552` is simultaneously "suite missed it" and "suite caught it," and is described once as an ast_diff *size* defect and once as a *child_count leak*. `#552` is the ast_diff-155KB issue; the child_count/children leak is a different issue (#698 / the #543-classify family). In an RFC whose whole thesis is rigor about *which test catches which defect*, a mis-cited counter-example undercuts the headline argument. Fix the issue number and make the "one value test that worked" example unambiguous.

### 2. Gap: mechanism 4 stops at `handle_call_tool`, but there is **zero** real stdio round-trip coverage (#691)
"Contract tests run through the real boundary — `handle_call_tool`, not `execute()`" is correct but incomplete. Per #691, the repo has **0** tests that spawn the server and drive a real `stdio_client`/`ClientSession` (`initialize`→`tools/call`) — `handle_call_tool` is still in-process and skips JSON-RPC framing + wire serialization, the exact layer the TOON 1.96× and envelope defects live in. Mechanism 4 should name a **stdio round-trip golden** as the outermost boundary tier and cross-link #691, not stop at in-process dispatch.

### 3. Gap: cost invariants should **ratchet down**, not merely exact-pin
An exact-pin byte budget (per the exact-assertion rule) forces a *conscious* re-pin — good — but it still permits re-pinning **upward** every time output grows, so bloat creeps in one "conscious" bump at a time (this is how 1.96× would have survived a pinned-but-rising ceiling). For *cost* specifically, make the budget a **monotonic non-increasing ratchet** (like the mutation score and loose-assertion ratchets): raising it requires explicit justification in the PR, lowering is free. Exact-pin is right for *counts*; ratchet-down is stronger for *cost*.

### 4. Risk: mutation harness vs the known pytest OOM / xdist landmine
Mutation testing reruns the suite per mutant. This repo has a documented Mock-parent-chain infinite-loop OOM (140 GB incident) and an xdist-worker memory profile. Before the CI job lands, the harness must pin **single-instance pytest** for mutant runs (or a tightly-scoped test subset per module) and cap memory, or the nightly mutation job will reproduce the OOM at N× scale. Add this to the test plan / drawbacks.

### Minor
- Acceptance criteria leaves the mutation baseline number as the headline deliverable but the RFC is `draft` without it — fine for draft, but the number is what actually answers the investor's "why don't 2× tests catch bugs?"; prioritize landing the baseline so the RFC carries data, per §11 ("a claim is a BELIEF until measured").

— perpetual dogfood QA arm (review-only).

aimasteracc:
## Codex triage (2 P2 — both VALID, RFC updated `ff5296bf`)
1. **Pin mutmut to fork-capable runners** — correct. `mutmut` relies on `os.fork` (absent on Windows). RFC now specifies the mutation job runs on **`ubuntu-latest` only**, as a separate workflow NOT part of the cross-platform reusable test matrix. Same Linux pin if we ever swap to cosmic-ray.
2. **Per-run reports as artifacts, not committed** — correct; committing generated reports every nightly/on-label run would dirty the repo / need bot commits. RFC now: per-run report = **CI artifact** (uploaded); only a curated `tests/effectiveness/BASELINE.md` (surviving-mutant count per module) is tracked, refreshed deliberately. Acceptance criteria updated to match.

Both were spec-level catches before any code — exactly the value of RFC-first review. RFC stays **draft** pending investor accept → then implementation.

PR #702, opened 2026-06-14
feat(nav): flag stale lineage hierarchy when source is newer than the index (#692)

### LINKED ISSUE #692: [P3][nav] lineage hierarchy can report stale rows (live refs vs index-derived hierarchy in one envelope)
## Context — Codex P2 on #688 (fdf17cf0)

`nav action=lineage` (#688) added a `hierarchy` block sourced from persisted `ast_index` extends-edge rows. The rest of the lineage envelope (`references`, `definition_count`) comes from a **live** re-scan via `execute_find_references`. The in-process per-symbol response cache is correctly invalidated on source/index change.

**The gap:** the on-disk AST index is not re-synced per query (by design — re-indexing is 5-15s; the tool's docstring says "Cache via project_index"). So when source is edited *without* re-indexing:
- `references` reflect the edit (live scan)
- `hierarchy` still reports the **old** extends edges from the stale index

→ e.g. delete `sub.py`; `nav lineage Base` still lists `Sub` as a subclass while references show it gone — an inconsistency *within a single response*.

## Why this is P3, not P2
This is the same freshness lag every index-derived query has (`structure action=class_tree`, `codegraph`). It is **not a regression** — it's inherent to the index model. What makes it worth tracking is the *mixed* freshness in one envelope (live + indexed), which is a footgun for an agent that trusts both equally.

## Proposed cheap fix (surgical, no re-index)
`_hierarchy_for` already has access to the source fingerprint (max source mtime) and `_index_signature()` (index mtime). When `max_source_mtime > index_mtime`, annotate the block: `hierarchy["index_stale"] = True` (+ a one-line hint to run `project_index`/sync). Honest signal, no expensive auto-reindex, no behavior change when the index is fresh.

## Acceptance
- RED test: build index with `Sub(Base)`, delete `sub.py` (no re-index), assert `hierarchy["index_stale"] is True`.
- GREEN test: fresh index → `index_stale` absent/False.
- Exact `==` assertions on counts; no `>=`.
- CLI↔MCP parity preserved.

Follow-up to #688 / #568 defect-1.

aimasteracc:
## Summary
Closes #692 (P3, follow-up to #688/#568). `nav action=lineage` builds its `hierarchy` block from the persisted AST index, while `references` come from a LIVE re-scan — so if source is edited without re-indexing, the hierarchy can list stale rows (e.g. a deleted subclass) with no signal. This adds an honest freshness flag (no expensive auto-reindex).

## Fix
In `_hierarchy_for`: compare source max mtime (reuses the already-computed `_dep_graph_fingerprint.max_mtime_ns` — no second tree walk; falls back to `compute_graph_fingerprint` if the graph build failed) against `_index_signature()` (index db+wal+shm mtime). When source is newer:
- `hierarchy["index_stale"] = True` (always present as a bool — mirrors `subclasses_truncated` for contract stability)
- `hierarchy["index_hint"]` = run project_index to refresh (present only when stale, keeps the fresh path lean)

## Tests (RED-first, exact ==, deterministic mtimes via os.utime)
- `test_hierarchy_flags_stale_after_source_edit_without_reindex` — index built, `sub.py` mtime bumped +10s, no re-index → `index_stale is True`, `subclass_count == 1` (the stale row is still counted — that's the point).
- `test_hierarchy_not_stale_when_index_fresh` — source mtimes −60s, fresh index → `index_stale is False`, `index_hint` absent, `subclass_count == 1`.
36/36 lineage + 9/9 graph-cache tests pass; ruff + mypy clean. CLI parity: response fields flow through `--symbol-lineage` unfiltered.

@codex review

### DIFF
--- a/tests/unit/mcp/test_symbol_lineage_tool.py
+++ b/tests/unit/mcp/test_symbol_lineage_tool.py
@@ -256,6 +256,107 @@ class TestInheritanceLineage:
         assert result["success"] is True
         assert "hierarchy" not in result
 
+    def test_hierarchy_flags_stale_after_source_edit_without_reindex(
+        self, tool, tmp_path
+    ):
+        """#692: index_stale=True when source is newer than the AST index."""
+        import os
+        import time
+
+        from tree_sitter_analyzer.ast_cache import ASTCache
+
+        _write_py(
+            tmp_path, "base.py", "class Base:\n    def run(self):\n        pass\n"
+        )
+        sub_path = tmp_path / "sub.py"
+        _write_py(
+            tmp_path,
+            "sub.py",
+            "from base import Base\nclass Sub(Base):\n    pass\n",
+        )
+        # Build the index so hierarchy is populated.
+        ASTCache(str(tmp_path)).index_project(max_files=50)
+
+        # Bump sub.py's mtime to AFTER the index was written, WITHOUT re-indexing.
+        # Use a deterministic future timestamp to avoid timing races.
+        future_ns = int(time.time() * 1e9) + 10_000_000_000  # 10 seconds ahead
+        future_s = future_ns / 1e9
+        os.utime(str(sub_path), (future_s, future_s))
+
+        # The tool must re-build the dep graph (source mtime changed) so
+        # _dep_graph_fingerprint.max_mtime_ns reflects the bumped mtime.
+        # Force cache invalidation by resetting the tool state.
+        tool._dep_graph = None
+        tool._dep_graph_fingerprint = None
+        tool._symbol_cache = {}
+
+        result = asyncio.run(tool.execute({"symbol": "Base", "output_format": "json"}))
+        hier = result["hierarchy"]
+        # Stale row still reported (the index still has Sub).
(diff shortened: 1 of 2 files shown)


aimasteracc:
## Post-merge Codex triage (1 P2 — REAL but bounded → follow-up filed)
**'Fingerprint all indexed sources for staleness'** — valid. The freshness check compares against `compute_graph_fingerprint`, whose `_SOURCE_EXTS` is a fixed subset (py/js/ts/java/go/rust/c/cpp/...). For indexed languages OUTSIDE that set (kotlin, php, ruby, swift, scala, c#, lua, …), those source files don't contribute to `max_mtime_ns`, so `index_stale` can under-report (miss a real edit). This is an *incompleteness of the new advisory signal*, NOT a regression — before #692 there was no signal at all, and the hierarchy was already index-derived. Filing a focused follow-up rather than rushing: the correct fix is to derive staleness from the AST index's own recorded source set (authoritative) or pass the index's extension set to `compute_graph_fingerprint(extensions=...)`.

Issue #548, opened 2026-06-12
[P2][UX-journey] --callers cold-start NOT_FOUND without index hint; README example uses removed click API


aimasteracc:
## [UX-journey] --callers cold-start: NOT_FOUND before --full-index with no hint; README example uses removed click API
Persona-1 (newcomer on a fresh clone of click):
- `--callers "BaseCommand.make_context"` → NOT_FOUND both before and after `--build-project-index` (which builds the OTHER cache, .tree-sitter-cache). Only `--full-index` populates the call graph — nothing in the failure or the README CLI section says so. (docs/index-lifecycle.md now explains it, but the README examples and the NOT_FOUND response don't point there.)
- The README example symbol `BaseCommand.make_context` no longer exists — click 8.2 renamed BaseCommand→Command; with `Command.make_context` (post full-index) it works.
Fix: (a) NOT_FOUND on callers/callees when the AST index is empty/partial should reuse #497's index-state hint; (b) one prerequisite line in the README CLI examples; (c) refresh the example symbol.

aimasteracc:
Fixed by #705: README CLI examples now note callers/callees need `--full-index` first; `--callers`/`--callees` on an empty call-graph index now surface a '--full-index first' hint (reuses #497 index-state gate); stale BaseCommand example already absent.

PR #497, opened 2026-06-11
fix(search): hyphae select reports index state — 0 from empty index is not 'no matches' (#491)

aimasteracc:
Closes #491.

## Problem
select on an empty/partial index returned silent `count: 0` + "Answer from these symbols" — an agent on an unindexed project read "no callers" instead of "index isn't built".

## Fix
- `index_state` field: missing / empty / ready (detection mirrors status tool's `truly_indexed` pattern — total_files > 0)
- State-routed next_step: not-ready → WARN "Index missing or empty — run index action=auto"; ready+0 → NOT_FOUND with selector hints (legit zero, no false alarm); ready+truncated/matches → existing hints

## Tests
6 RED-first exact-pin (empty→WARN, ready-zero→NOT_FOUND no false alarm, missing-cache exception path…); 55 hyphae + 53 contracts green; coverage 92.4%; ruff clean.

PR #705, opened 2026-06-14
fix(docs,mcp): first-contact onboarding — py3.10 prereq, callers/callees empty-index hint (#545, #548)

### LINKED ISSUE #545: [P1][UX-journey] README quick-start fails on macOS system Python 3.9 — no 'requires 3.10+' in the path
## [UX-journey] README quick-start fails silently on macOS system Python (3.9) — no "requires 3.10+" anywhere in the path
Persona-1 walkthrough: `/usr/bin/python3` (3.9, macOS default) + the README's exact one-liner `pip install "tree-sitter-analyzer[mcp]"` →
```
ERROR: Could not find a version that satisfies the requirement ... (from versions: none)
```
preceded by ~90 lines of "Ignored the following versions that require a different python version". The Get Started section never states Python ≥3.10 (only a PyPI badge shows it). First-contact failure with maximal noise and zero guidance. Fix: one line in Get Started — *Requires Python 3.10+ (`python3 --version`)* — in all 3 README languages.

aimasteracc:
## Summary
Closes #545 (P1) and #548 (P2) — one cohesive line: **a brand-new user's first contact, zero friction.** (面 E in the backlog map.) This is the highest-ROI fix for the 'downloads climbing, stars flat' signal — newcomers were hitting silent failures on step one and not converting.

### #545 (P1) — README quick-start failed on Python 3.9 with no guidance
The Get Started one-liner failed on macOS system Python (3.9) with 90 lines of pip noise and no statement of the requirement. Added a prereq line right under Get Started in **all 3 READMEs** (verified `requires-python = ">=3.10"` from pyproject):
> **Requires Python 3.10+** (check: `python3 --version`).

### #548 (P2) — callers/callees cold-start
- README CLI examples now state the call-graph needs `--full-index` first (NOT `--build-project-index`, a different cache) — 3 READMEs.
- **Code:** `--callers`/`--callees` returning NOT_FOUND on an empty/unbuilt call-graph index now surface a hint (reuses the #497 index-state gate; fires only when the graph has zero edges, so a genuine not-found in a populated graph stays clean). Before: `next_step: ""`. After: `"...Run tree-sitter-analyzer --full-index first, then retry."`
- Stale example symbol `BaseCommand.make_context`: already absent from READMEs (removed earlier) — no change needed.

## Tests (RED-first, exact substrings)
- `TestOnboardingPrerequisites` — 6 params (en/zh/ja × py-version + full-index), exact substrings `3.10`/`python3 --version`/`--full-index`.
- `TestEmptyIndexHint` — callers + callees empty-index hint mentions `--full-index`; + a no-spurious-hint test when the graph is populated.
54 passed; README under 500 lines; `test_readme_counts_match_registry` green.

@codex review


aimasteracc:
## Post-merge Codex triage (1 P2 — REAL, follow-up incoming)
**'Don't treat zero-edge projects as unindexed'** — valid edge case. The hint fires on `not has_any_call_edges`, which conflates *index-not-built* with *index-built-but-genuinely-zero-edges* (e.g. a single `def solo(): return 1`). Such a project would wrongly be told to run `--full-index`. Fix: gate the hint on index EXISTENCE (real index-state signal / index.db presence), not edge-count==0. Follow-up PR incoming.

PR #706, opened 2026-06-14
fix(mcp): callers/callees full-index hint fires only when index is unbuilt, not on zero-edge projects (#705 follow-up)

aimasteracc:
## Summary
Codex P2 on merged #705. The #705 empty-index hint used `has_any_call_edges = len(graph.call_edges()) > 0`, which conflated *index-not-built* with *index-built-but-zero-edges* (e.g. a single `def solo(): return 1`). A legitimately indexed, edge-free project would be wrongly told to run `--full-index`.

## Fix
Gate the hint on the authoritative built-state signal `_data_source` (set to `"cache"` by `_get_call_graph()` when a live `ASTCache` with `total_files > 0` exists — i.e. the user ran `--full-index`):
```python
has_any_call_edges = (self._data_source == "cache" or len(graph.call_edges()) > 0)
```
The edge-count arm is retained so an unindexed project whose fresh parse DOES find edges still isn't treated as unbuilt (keeps `test_callers_non_empty_index_no_spurious_hint` green). Applied to BOTH callers_tool.py and callees_tool.py.

## Tests (RED-first, exact)
- `test_callers_built_index_zero_edges_no_hint` / `test_callees_built_index_zero_edges_no_hint` — after `ASTCache.index_project()` on a solo-function project (total_files==1, zero edges), `--full-index` NOT in next_step.
- #705's `TestEmptyIndexHint` (unbuilt → hint present) all still green. 44/44 pass; ruff+mypy clean.

@codex review

### DIFF
--- a/tests/unit/test_callers_callees_tools.py
+++ b/tests/unit/test_callers_callees_tools.py
@@ -513,3 +513,52 @@ class TestEmptyIndexHint:
             f"--full-index hint must NOT appear when index has edges; "
             f"got: {next_step!r}"
         )
+
+    @pytest.mark.asyncio
+    async def test_callers_built_index_zero_edges_no_hint(self, tmp_path) -> None:
+        """#705 follow-up: index IS built (total_files > 0) but the project has
+        no call edges (e.g. a single ``def solo(): return 1``).  NOT_FOUND must
+        NOT carry a --full-index hint — the user already indexed; they just have
+        a project with no calls."""
+        from tree_sitter_analyzer.ast_cache import ASTCache
+
+        (tmp_path / "solo.py").write_text(
+            "def solo():\n    return 1\n", encoding="utf-8"
+        )
+        cache = ASTCache(str(tmp_path))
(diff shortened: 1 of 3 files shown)


aimasteracc:
## Post-merge Codex triage (1 P2 — REAL, but I'm stopping the oscillation, not point-fixing again)
**'Preserve the hint for partial zero-edge caches'** — valid. A user who ran `--build-project-index` (AST cache, `_data_source=='cache'`) but NOT `--full-index` (call graph) now gets NO hint — re-breaking #548's partial-cache scenario. This hint has oscillated: #705 over-hinted (zero edges → always hint), #706 under-hints (cache exists → never hint). **Root cause: there's no authoritative 'call-graph-built' signal distinct from 'AST cache exists'**, so any binary heuristic flips one error mode into the other. A third reactive point-fix would just oscillate again. Filing a follow-up to do it right with a real signal (index metadata flag) — deliberate, not reactive.

PR #710, opened 2026-06-14
fix(queries): Python FUNCTIONS no longer double-captures every function (#557)

### LINKED ISSUE #557: Python FUNCTIONS query contains a verbatim duplicate pattern — every function double-captured
## Found by

opencode adversarial review on #556 (the exact-pin batch made the smell visible: a 4-function fixture pins `functions == 32` = 4 defs × 4 captures × **2 identical patterns**).

## Bug

`tree_sitter_analyzer/queries/python.py:14-24` — `FUNCTIONS` contains two VERBATIM-identical `(function_definition ...)` patterns; the second (`@function.async`) was presumably meant to anchor `async` functions (the description says "including async") but lacks any async discriminator, so every function matches twice.

```
(function_definition name: ... params: ... body: ...) @function.definition
(function_definition name: ... params: ... body: ...) @function.async   ← identical, no async anchor
```

## Impact

Every consumer of the `functions` query key sees 2× captures for Python. (In tree-sitter, `async def` is a `function_definition` with an `async` keyword child — if separate counting is wanted the second pattern needs the anchor; otherwise delete it.)

## Acceptance

- [ ] Fix the query (anchor or delete the duplicate)
- [ ] Re-pin `tests/unit/core/test_query_code_cross_language_compatibility.py` `functions == 32` → `== 16` (and the intersection pins if affected) — the ratchet pin from #556 will go red and force this consciously
- [ ] Check golden masters / other consumers for the same double-count

aimasteracc:
## Summary
Fixes #557 (face A — extraction correctness). The Python `FUNCTIONS` query had a verbatim-duplicate pattern: `@function.async` was byte-identical to `@function.definition` (same `function_definition name/params/body`) and did NOT filter for async — so it matched **every** function, double-capturing all of them in the `functions` query output.

**Proof (3-function fixture incl. 1 async):** anchor captures 6→3, total (node,capture) tuples 24→12.

## Fix
`grep` confirmed **zero consumers** of `@function.async` anywhere → it was a copy-paste duplicate. Removed the second pattern (simplest correct fix). Verified no other `queries/*.py` module carries the same duplicate (Python-only).

## Tests (RED-first, exact ==)
`TestFunctionsQueryNoDuplicate`:
- `test_functions_query_no_double_capture` — anchor count `== 3` (was 6).
- `test_functions_query_no_async_capture_name` — `@function.async` absent.
FUNCTIONS stripped-length pin re-pinned 315→159 (the query shrank). 51 query tests + 1048 verification suite green; Python golden unaffected; ruff+mypy clean. (Swift golden failure is a pre-existing env flake — missing tree-sitter-swift wheel, identical on base.)

@codex review

### DIFF
--- a/tests/integration/mcp/test_user_story_3_query_integration.py
+++ b/tests/integration/mcp/test_user_story_3_query_integration.py
@@ -719,7 +719,9 @@ asyncio.run(main())
         with open(output_file, encoding="utf-8") as f:
             file_content = json.load(f)
             assert "results" in file_content
-            assert len(file_content["results"]) == 104
+            assert (
+                len(file_content["results"]) == 52
+            )  # was 104 (double-captured before #557); 52 is the correct single-capture count
 
         print(
             f"✓ ファイル出力最適化テスト成功: {result['count']}個の結果を{output_file.name}に保存"
--- a/tests/unit/core/test_queries_python.py
+++ b/tests/unit/core/test_queries_python.py
@@ -85,7 +85,7 @@ class TestPythonQueries:
         """Test that query constants are properly defined"""
         # Exact stripped lengths (update when query strings change)
         constant_lens = [
-            (FUNCTIONS, 315),
+            (FUNCTIONS, 159),
             (CLASSES, 156),
             (VARIABLES, 322),
             (IMPORTS, 445),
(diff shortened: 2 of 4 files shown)


aimasteracc:
## Post-merge Codex triage (1 P1 — FALSE POSITIVE, no action)
**'Add tracking to the new skip'** — the skip at `test_queries_python.py:399` already reads `pytest.skip("...; tracked: optional tree-sitter grammar dependency")`. AGENTS.md Anti-Patterns §1 explicitly accepts `tracked: ...` as a valid tracking reference (alongside `#123`/`POSTMORTEM`), and the enforcing contract `test_skips_have_tracking_references` **passed in this PR's green CI** (16/0). Codex flagged it without accounting for the `tracked:` prefix already present. No change needed.

PR #712, opened 2026-06-14
fix(mcp): enum-mode rejections enumerate valid values (#690)

### LINKED ISSUE #690: [P3][dogfood] facade 'Invalid mode: X' error never lists valid modes; recovery_hint is boilerplate (viz/health)
## Summary
When an action rejects a `mode` value, the error names the bad value but never lists the valid ones, and `recovery_hint` is generic boilerplate. An agent cannot self-correct — it has no way to discover the legal mode set. Worse: `mode=summary` is *valid* for some actions and *invalid* for others, so guessing fails inconsistently.

## Repro
```
viz action=similarity mode=summary
health action=dead     mode=summary
```
Both return:
```json
{
  "success": false,
  "error": "ValueError: Invalid mode: summary",
  "recovery_hint": "Review the error message and adjust your request.",
  "verdict": "ERROR"
}
```
The `recovery_hint` says "review the error message" — but the error message contains no valid modes to review.

## Why it's confusing
`health action=deps` documents `mode=summary` as a **valid** value. So "summary" works for `deps` but is rejected by `dead` and `similarity`. An agent that learned `mode=summary` on one action gets an opaque rejection on another, with no enumerated alternative.

## Suggested fix
Echo the valid set in the error and/or recovery_hint, e.g.:
```
"error": "Invalid mode 'summary' for action=dead. Valid modes: all",
"recovery_hint": "Retry with one of: all"
```
Cheap, deterministic, and turns a dead-end into a self-correctable error. Applies anywhere a `mode`/enum param is validated in the facades.

## Environment
develop @ 55cac0f9. MCP `viz` + `health` facades.
Found via perpetual dogfood QA round (file-only, no code change).

aimasteracc:
## What

Closes #690. Five facade-reachable MCP tools rejected an invalid `mode` with a bare `ValueError("Invalid mode: X")` — the agent saw the bad value but never the legal set, so it could not self-correct, and `recovery_hint` fell back to the dead-end boilerplate ("Review the error message…") pointing at a message with nothing to review.

This routes every bare enum-mode raise through the existing `invalid_enum_error()` helper, so each rejection carries the canonical `Invalid mode: 'X'. Valid values: …` enumeration. As a side effect the canonical enum `recovery_hint` now fires (the central classifier keys on the `valid values:` substring), fixing #690's second half too.

## Plane-B line, not a point

This is one coherent line under **面 B (envelope / agent contract)** — *"every enum rejection is self-correctable"* — not a one-off patch of the reported repro. Tools touched:

- `code_similarity` (`viz action=similarity` — the #690 repro)
- `incremental_sync`, `auto_index`, `ast_cache` (`index` facade)
- `decision_journal` (`project action=journal`)

`ast_cache` intentionally omits the deprecated `fts_search` alias (J1) from the enumerated guidance so agents are steered to the supported `search` name (the alias is still accepted at the boundary).

Already-enumerating forms (`Must be 'symbol' or 'file'`, `expected one of: …`) were left as-is — an agent can already self-correct from those; force-normalizing them would be churn risking exact-string test breakage for no agent-facing gain.

## Mechanical invariant (so it can't silently regress)

`tests/unit/mcp/test_enum_error_enumeration.py`:
1. Parametrized `validate_arguments` check — every touched tool's enum error contains `Valid values:` and lists its real valid modes. A regression to a bare raise goes red.
2. A real `handle_call_tool` **boundary** test of the `viz action=similarity` repro — asserts the final envelope's `error` enumerates the values AND `recovery_hint` is the canonical enum hint, not the boilerplate. (Boundary, not `execute`, per the audit lesson that execute-level tests miss facade arg-projection seams.)

## Test plan

- [x] `test_enum_error_enumeration.py` — 6 passed (incl. boundary)
- [x] 1187 passed across the 5 tool suites + `tests/unit/mcp/tools/` + `test_agent_contracts.py`
- [x] ruff / ruff format / mypy clean
- [ ] CI green on all axes
- [ ] Codex review triaged

### DIFF
--- a/tests/unit/mcp/test_enum_error_enumeration.py
+++ b/tests/unit/mcp/test_enum_error_enumeration.py
@@ -1,0 +1,121 @@
+"""Contract: every enum-mode rejection enumerates the valid values (#690).
+
+Background (#690): facade actions that reject a ``mode`` value used to raise a
+bare ``ValueError("Invalid mode: <got>")`` — the agent saw the bad value but
+never the legal set, so it could not self-correct, and ``recovery_hint`` fell
+back to generic boilerplate ("Review the error message and adjust your
+request.") that pointed at a message with nothing to review.
+
+This test pins the fix as a mechanical invariant on two surfaces:
+
+1. **Inner tool ``validate_arguments``** — a bad ``mode`` raises a ``ValueError``
+   whose message carries the canonical ``"Valid values: ..."`` enumeration and
+   actually lists the real valid modes. A regression to a bare raise goes red.
+2. **The real ``handle_call_tool`` boundary** (the #690 headline repro,
+   ``viz action=similarity mode=<bad>``) — the final envelope's ``error`` lists
+   the valid values AND ``recovery_hint`` is the canonical enum hint, not the
+   boilerplate. Asserted through the boundary (not ``execute``) per the audit
+   lesson that ``execute``-level tests miss facade arg-projection seams.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+from unittest.mock import Mock, patch
+
+import pytest
+
+from tree_sitter_analyzer.mcp.server import TreeSitterAnalyzerMCPServer
+from tree_sitter_analyzer.mcp.tools.ast_cache_tool import ASTCacheTool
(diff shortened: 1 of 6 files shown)


aimasteracc:
@codex review

Issue #708, opened 2026-06-14
[P3][mcp] callers/callees full-index hint needs an authoritative 'call-graph-built' signal (stop the #705/#706 oscillation)


aimasteracc:
## Context
The `--callers`/`--callees` empty-index hint has oscillated across #705→#706:
- #705: `hint if len(call_edges)==0` → over-hints a legitimately edge-free but fully-indexed project (`def solo(): return 1`).
- #706: `hint if not (_data_source=='cache' or edges>0)` → under-hints a PARTIAL cache (AST-cached via `--build-project-index` but call-graph not built via `--full-index`) — re-breaking #548's target scenario.

**Root cause:** no authoritative signal distinguishes 'call-graph was built (`--full-index` ran)' from 'an AST cache exists' or 'zero edges'. Any binary heuristic flips one error into the other.

## Proper fix (deliberate, not another reactive point-fix)
Record a 'call_graph_built' marker in the index metadata when `--full-index` populates the call-edges table (or query whether that table was populated by a full-index run vs absent). Gate the hint on THAT signal: fire only when the call graph was never built; suppress for a fully-indexed project regardless of edge count. (Relates to face D — index/cache state honesty, see #578.)

## Acceptance (RED-first, exact ==)
- fully-indexed edge-free project → NO hint.
- AST-cache-only / call-graph-never-built → hint present.
- both callers + callees; CLI↔MCP parity.

Follow-up to #705/#706/#548.

aimasteracc:
Also covers #709 Codex P2: the callers/callees `agent_summary.next_step` in the NOT_FOUND-but-index-built (zero-edge) case gives the same ambiguous re-index advice. The authoritative 'call-graph-built' signal fix should drive BOTH the empty-index hint AND the agent_summary next_step from one source, so they can't disagree or mislead.

aimasteracc:
Fixed by PR #928 (squash-merged). `callers_tool` now reads an authoritative `_graph_built` marker file written by the indexer.

Issue #578, opened 2026-06-12
[P2][QC-battle] Reads during full-index rebuild return success:true with phantom-empty data — no rebuilding signal


aimasteracc:
## [QC-battle] Concurrent reads during full-index rebuild return success:true with wrong data — no rebuilding signal
Develop @ 9fffb7d7. During `--full-index --full-index-mode full` (~70s on this repo), concurrent reads serve the half-rebuilt table as truth:
```
t=1s:  --callers "HealthHistory.append" → success: true, verdict: NOT_FOUND, caller_count: 0   (pre/post-build: 50)
t=10s: --class-hierarchy LanguagePlugin → subclass_count: 3                                    (pre/post-build: 26)
```
No `index_building` flag, no warning, data_source still says "cache". An agent (or CI) querying during a rebuild acts on phantom-empty results — the #491 trust family at its worst, because here the index EXISTS and lies. Fix: write a build-in-progress marker (single sqlite meta row) and have readers either report `verdict: WARN index_rebuilding` or serve the pre-build snapshot (WAL makes snapshot reads feasible).
Positives from the same battle worth recording: 6-way concurrent reads deterministic, parallel incremental syncs clean, 20-thread subscribe/unsubscribe storm leak-free.

aimasteracc:
**Triage (lead): this P2 is RFC-gated, not a direct fix.** The proposed remedy adds (a) a build-in-progress marker as a SQLite meta row in the AST cache → an **ast_cache schema change**, and (b) a new reader contract (`index_building` flag / `verdict: WARN index_rebuilding` / pre-build snapshot reads) across CLI + MCP → a **cross-surface parity + envelope contract change**. Per CLAUDE.md's RFC process, both classes require an RFC before implementation (spec-first; the trust-family semantics — "the index EXISTS and lies" — deserve a designed contract, not an ad-hoc patch across N reader paths).

Next step: draft `rfcs/NNNN-index-rebuild-visibility.md` from the template covering: the meta-row schema, reader behavior (WARN vs WAL-snapshot pre-build read), the three-surface (CLI↔MCP) signal, and a RED-first test plan (concurrent read during `--full-index-mode full` must NOT return phantom-empty as truth). Keeping #578 open as the motivating issue; implementation waits on the accepted RFC.

[...]

Issue #491, opened 2026-06-11
[P3][dogfood] Hyphae select on empty/partial index: silent count:0 + 'answer from these symbols' — unindexed reads as no-callers


aimasteracc:
## Summary
`search action=select` on an **empty or partial index returns a silent success with count: 0** — no warning that the index is missing/incomplete — and its `next_step` still says "Answer from these symbols". An agent on a not-yet-indexed project reads "this function has no callers" instead of "the index isn't built".

## Repro (develop @ 465ea828, fresh worktree)
```
# worktree with only an on-demand partial cache (192KB, built by a --callees call):
select ".function"                                      → count: 0, total_matches: 0, no hint
select "#apply_toon_format_to_response"                 → count: 0   (the symbol definitely exists)
next_step: "Answer from these symbols, or refine the selector…"     ← answering from zero symbols
# after --full-index (219MB):
same selector → count: 100, total_matches: 119 ✓
```

## Why it matters
This is the #434 (DF-8 honest index status) trust family: a wrong "no results" is worse than an error. Selector queries are exactly what agents use for blast-radius decisions — a silent 0 on a cold project turns "unindexed" into "safe to delete".

## Suggested fix
When the symbol table is empty (or below a sanity floor) for the queried scope, return verdict WARN with "index empty/partial — run index action=build" instead of a bare 0; and make the zero-result next_step say "0 matches — check the selector or index status", not "answer from these symbols".

aimasteracc:
Fixed by #497 (merged)。① 三态 `index_state`(missing/empty/ready,与 status 工具同款检测,单次廉价 sqlite count);② missing/empty → verdict WARN + 「Index missing, empty, or unreadable — run index action=auto」(0 = not indexed ≠ no matches);③ opencode 兜底评审抓出部分索引盲点(原始复现场景:按需单文件 cache 归类 ready)→ 新增 `indexed_files` 标量 + ready+0 的 next_step 明示「across N indexed file(s)… if the project should contain more files, run index action=auto」;④ 工具描述文档化新字段与 WARN 语义。

aimasteracc:
**Retested on develop @ 99376f86 (#497) — CONFIRMED FIXED ✓** Select on an empty index now returns next_step: *'Index missing, empty, or unreadable. Run the `index` tool with action=auto to build the cache (if this persists, check .ast-cache permissions).'* — unindexed no longer reads as no-callers. Good to close.

PR #723, opened 2026-06-14
fix(index): persist rebuild-in-progress marker so reads stop lying mid-rebuild (#578)

aimasteracc:
## Summary
Closes the read-side of #578 (P2): during a full `--full-index` rebuild, concurrent reads on another connection/process saw the committed-but-empty / half-filled `ast_index` table and returned `success: true` with phantom-empty data (`caller_count: 0` when the real answer is 50). The index exists and lies — #491 trust family at its worst.

## Root cause
`ASTCache.index_project(force=True)` does `DELETE FROM ast_index` + `commit()` up front, then re-populates in **bounded 500-file batches** (`_commit_index_results`, the "A4" RSS/OOM guard). The batched re-insert commits incrementally over the ~70 s rebuild, so the empty/partial state is visible to every other connection during that window.

## Why not the "obvious" atomic fix
Wrapping the whole rebuild in one transaction (so WAL readers see the old snapshot until a single commit) would remove the window — but revive the exact OOM that A4's bounded-batch commit was added to prevent (7.9k-file Java repos). So that path is out by design; the marker is the safe route.

## Change
- **New `_ast_cache_build_state.py`** — persisted single-row marker (mirrors the `ast_resolve_state` idiom). Cross-connection/cross-process visible, and **self-expiring via TTL** so a crashed rebuild never wedges readers into permanent "rebuilding" warnings.
- **`index_project` force path** — stamps the marker before the DELETE, clears it in `try/finally` after re-population (covers the error path too).
- **First consumer — `index action=status`** — reports `index_rebuilding: true` + verdict `WARN`, and distinguishes the transient mid-rebuild empty window from a genuinely missing index (which would otherwise tell an agent to start a *second* rebuild). Also gives #708 the authoritative "build in progress" signal it asked for.

## Scope / follow-up
Deliberately focused per PR-hygiene. The read-family rollout (callers / callees / class-hierarchy emitting `WARN index_rebuilding` instead of phantom `NOT_FOUND`) is a separate follow-up PR — those readers are decentralized and bundling them here would make this a kitchen sink. **#578 fully closes with both PRs.**

## Test plan
- [x] `tests/unit/test_ast_cache_build_state.py` — marker unit + missing-table degrade + TTL staleness (crash recovery) + mid-rebuild window contract (asserts marker set while table is empty, cleared after) + status surfacing for both partial and empty windows, via a **separate connection** to pin the cross-connection contract.
- [x] change-impact focused command (194) green; full suite `uv run pytest -q` = 20195 passed (1 unrelated startup-latency flake: `test_repeated_headless_initialize_stays_under_budget`, passes isolated in 1.16s — load-induced under 20k parallel tests, untouched by this change).
- [x] ruff + mypy clean; patch-coverage gate: no added executable misses; CLI parity smoke (`--codegraph-status`) OK, field correctly omitted on the non-rebuilding path.

### DIFF
--- a/tests/unit/test_ast_cache_build_state.py
+++ b/tests/unit/test_ast_cache_build_state.py
@@ -1,0 +1,293 @@
+"""Persisted "full rebuild in progress" marker (#578).
+
+A ``--full-index --full-index-mode full`` rebuild does ``DELETE FROM ast_index``
++ ``commit`` up front, then re-populates in bounded 500-file batches (the "A4"
+RSS guard in ``_commit_index_results``). Because the batched re-insert commits
+incrementally over the ~70 s rebuild, a concurrent reader on another connection
+(or process) sees the committed-but-empty / half-filled table and returns
+``success: true`` with phantom-empty data (``caller_count: 0`` when the real
+answer is 50).
+
+The single-big-transaction "atomic swap" cannot be used here — it would revive
+the exact OOM that A4's bounded-batch commit was added to prevent. So the fix is
+a persisted, cross-connection marker: the rebuild stamps a single sqlite meta
+row before the DELETE and clears it after re-population, and readers consult it
+to warn instead of trusting the half-built table.
+
+These tests pin the marker contract RED-first.
+"""
+
+from __future__ import annotations
+
+import sqlite3
+import time
+from pathlib import Path
+
+import pytest
+
+from tree_sitter_analyzer import _ast_cache_build_state as bs
+from tree_sitter_analyzer.ast_cache import ASTCache
+from tree_sitter_analyzer.mcp.tools.codegraph_status_tool import CodeGraphStatusTool
+
+
+def test_build_state_helpers_degrade_on_missing_table() -> None:
+    """No ast_build_state table → safe defaults, no raise; mark creates it."""
+    conn = sqlite3.connect(":memory:")
+    conn.row_factory = sqlite3.Row
+    try:
+        # Absent table → not building, clear is a no-op, neither raises.
+        assert bs.build_in_progress(conn) is False
+        bs.clear_build_in_progress(conn)  # must not raise
+
+        # mark creates the table and persists a building row.
+        bs.mark_build_in_progress(conn)
+        assert bs.build_in_progress(conn) is True
+
+        # clear flips it back off.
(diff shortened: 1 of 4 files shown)


aimasteracc:

Codex's review quota was exhausted on this PR, so per the project's review-arm fallback chain (codex → opencode → adversarial agent) this PR was reviewed by **opencode** (adversarial, falsify-not-endorse). Every finding triaged below; all addressed in the amend (`90c383ba`).

|---|---|---|---|
| **P2** | MARK+DELETE+commit lived *outside* the `try`; if DELETE/commit raised (e.g. SQLITE_FULL) the `finally` never cleared → marker stuck for the whole TTL on a healthy cache | ✅ Real — fixed | Moved MARK+DELETE+commit **inside** the try; corrected the misleading comment; added regression test `test_marker_cleared_when_delete_phase_raises` (proxy conn fails the DELETE, asserts marker cleared) |
| P3 | TTL 900s could be shorter than a real 20k-file rebuild → stale mid-rebuild | ✅ Real — fixed | Primary staleness is now **recorded-pid liveness** (live rebuilder stays fresh past any TTL); TTL bumped to 1 h backstop |
| P3 | `pid` stored but never read (dead metadata) | ✅ Real — resolved | `pid` is now the primary liveness signal (crashed rebuilder ⇒ stale immediately) |
| P3 | Wall-clock TTL vulnerable to NTP backward-step | ✅ Real — fixed | Added future-timestamp guard (`_CLOCK_SKEW_GRACE_SECONDS`) |
| P3 | mark-failure proceeds without a marker | ⚖️ Won't make fatal | Degrades to *pre-#578 behaviour* (not worse); aborting the rebuild would be a worse "can't rebuild under contention" mode — documented in `mark_build_in_progress` |
| P3 | `_is_rebuilding` degrades *open* on a 2 s SELECT timeout under contention | ✅ Correct by design | Added a clarifying comment (better to miss a WARN than wedge status behind a lock) |
| — | "creates db file?", cross-conn visibility, idempotent clear, row-factory handling | ✅ Verified non-issues | none |

New module is at **100%** coverage; ruff/mypy clean; patch-coverage gate clean. Full suite re-run after the `index_project` restructure.

Issue #798, opened 2026-06-14
[P1][dogfood] nav action=impact / --codegraph-impact: risk.level="critical" (score 100) maps to verdict=INFO + "Low risk change — proceed with edit" (missing "critical" branch in _impact_verdict inverts the edit-safety decision surface)


aimasteracc:
**Severity:** P1  |  **Surface:** MCP nav action=impact (mode=function_impact) and CLI --codegraph-impact --codegraph-impact-mode function_impact; root cause tree_sitter_analyzer/mcp/tools/codegraph_impact_tool.py:_impact_verdict (lines 583-594), with the next_step branch at lines 488-496 and the 4-level emitter at lines 180-187.
_Found by parallel dogfood audit round 3 on develop @ 471ac5f5; adversarially re-reproduced + dup-checked._

### Observed
Both CLI and MCP return risk.score=100, risk.level="critical", fan_in=114, cross_file_callers=64, but verdict=INFO and agent_summary.next_step="Low risk change — proceed with edit; run nearest test file afterwards." The summary_line is internally contradictory: "impact: 'apply_toon_format_to_response' risk=critical verdict=INFO". _impact_verdict has no branch for level=="critical" so it falls through `return "INFO"`; the next_step branch then treats INFO as the low-risk/proceed path. Net: every function in the highest-risk bucket (score>=60, biggest blast radius) is reported to agents as safe to edit. The medium-risk format_as_toon correctly yields verdict=REVIEW, confirming the two are inverted relative to actual risk.

### Expected
The verdict and next_step must be monotonic with risk.level: the most dangerous category (critical) must get the strongest warning, never the weakest. critical should map to a blocking/caution verdict (e.g. CAUTION or a dedicated DANGER) and next_step must warn ("trace callers and run tests before editing"), never "Low risk change — proceed with edit". Add a `if level == "critical": return "CAUTION"` (or stronger) branch to _impact_verdict (codegraph_impact_tool.py ~line 588) so the decision surface is monotonic with risk.score. A RED-first test should assert level=critical -> verdict in {CAUTION, DANGER} and that next_step does not contain "proceed with edit".

### Repro
```bash
cd /Users/aisheng.yu/git-private/tsa-dogfood
# CLI:
[...]

PR #817, opened 2026-06-14
fix(mcp): smart_context applies the syntax gate — no grade=A/SAFE on unparseable files (#754)

### LINKED ISSUE #754: [P1][dogfood] --smart-context returns grade=A/score=100/verdict=SAFE on a file that fails to parse (false green for edit-gating)
**Severity:** P1  |  **Surface:** CLI --smart-context on malformed/syntax-error input
_Found by the parallel multi-agent dogfood audit on develop @ 471ac5f5; adversarially re-reproduced and dup-checked by an independent verifier agent._

### Observed
--smart-context returns success=true with health.grade='A', health.score=100.0, health.signal='healthy', agent_summary.verdict='SAFE', edit_risk/risk='safe', recommendation='Grade A, risk: safe', next_step='Make a focused edit, then run the nearby tests and scoped change-impact.' The envelope has NO parse_errors / syntax_error / encoding_warning key. tree-sitter confirms the file does not parse (root_node.has_error=True). For the SAME file, --file-health reports signal='syntax_error', verdict='ERROR', grade='N/A', risk='high', recommendation='File fails to parse...'; --outline reports verdict='WARN' with 'Parse errors detected'. The non-UTF8 file via --smart-context returns grade='F', signal='no_data', verdict='CAUTION' — still never surfaces the syntax_error/encoding signal that file-health emits. Source confirms smart_context_tool.py performs no is_file_parse_broken() check (0 references), unlike file_health_tool.py which short-circuits on it before scoring.

### Expected
--smart-context should detect parse failures the way --file-health/--outline/--code-patterns do (call is_file_parse_broken / short-circuit), and on a non-parsing file downgrade the verdict and surface a parse_errors field or signal='syntax_error' — never report grade=A / score=100 / verdict=SAFE / risk=safe on a file tree-sitter cannot parse. The non-UTF8/binary case should likewise surface an encoding/parse signal rather than a bare no_data. An agent gating an edit on the smart-context verdict must not be told a corrupt file is SAFE to edit.

### Repro
```bash
cd into a TSA checkout (develop). Create a non-parsing Python file:
  mkdir -p _df_tmp && printf 'def broken(\n  x = \n  return\nclass {{{\n' > _df_tmp/syntax_err.py
Run:
  uv run python -m tree_sitter_analyzer --smart-context _df_tmp/syntax_err.py --output-format json
Compare with the sibling tools on the SAME file:
  uv run python -m tree_sitter_analyzer --file-health _df_tmp/syntax_err.py --output-format json
  uv run python -m tree_sitter_analyzer --outline _df_tmp/syntax_err.py --output-format json
Optional non-UTF8 case:
  printf '\xff\xfe\x00\x01broken\xc0\xc1' > _df_tmp/nonutf8.py
  uv run python -m tree_sitter_analyzer --smart-context _df_tmp/nonutf8.py --output-format json
```

### Verifier notes
Re-ran the exact repro in /Users/aisheng.yu/git-private/tsa-dogfood (develop). --smart-context on a file that tree-sitter cannot parse (direct check: root_node.has_error=True) returns success=true, health.grade='A', health.score=100.0, health.signal='healthy', verdict='SAFE', risk='safe', next_step='Make a focused edit...', with NO parse_errors/syntax_error/encoding key. The SAME file via --file-health correctly returns signal='syntax_error', verdict='ERROR', grade='N/A', risk='high'; --outline returns verdict='WARN' 'Parse errors detected'. Non-UTF8 file via --smart-context returns grade='F' signal='no_data' verdict='CAUTION' — never surfaces the syntax_error/encoding signal file-health does. Root cause confirmed in source: file_health_tool.py:205-207 short-circuits on is_file_parse_broken() (utils/parse_validity.py) before scoring; smart_context_tool.py imports only _build_signal and has ZERO references to is_file_parse_broken/parse_validity/syntax_error, calling scorer.score_file() with no parse-validity guard. Flag usage is correct (documented bare 'killer 1-call'), so not a user/flag error. No open issue covers this (only unrelated #733). P1 because it is a false-green on the exact tool an agent uses to gate an edit — an agent would proceed to edit a corrupt file on a SAFE/grade-A verdict.

aimasteracc:
Closes #754 (P1, edit-safety false-green family with #798/#780).

## Root cause
`smart_context` blends file_health + risk signals but calls `HealthScorer.score_file` **directly**, bypassing the shared syntax-validity gate that the sibling detection tools (`file_health` / `safe_to_edit` / `code_patterns`) already apply at their boundary. tree-sitter is permissive, so a file that fails to parse still yields a "clean" tree → `grade=A, score=100, verdict=SAFE, next_step="proceed with edit"`. That is a false green telling an agent it is safe to edit unparseable code.

## Fix (surgical, reuses existing infra)
Short-circuit `execute()` through the canonical `syntax_error_envelope` (verdict=ERROR, signal=syntax_error, "fix syntax before further analysis") via the existing `is_file_parse_broken` helper — before any grade is computed. Identical pattern to the sibling tools; clean files are unaffected.

## Verification
- Repro (RED): a syntax-error `.py` returned `grade=A / score=100 / verdict=SAFE`. After: `verdict=ERROR / signal=syntax_error`, no grade-A, no "proceed" advice; a clean file stays `SAFE`.
- New `TestSmartContextSyntaxGate` (broken→ERROR, clean→ungated). Logic verified directly (the shared local test guard was contended by a concurrent run; CI runs the full suite). ruff + mypy clean.
- Found by the parallel dogfood audit (round 3, #754); fixed by the QA arm joining the dev effort. Independent opencode review to follow.

### DIFF
--- a/tests/unit/mcp/test_smart_context_tool.py
+++ b/tests/unit/mcp/test_smart_context_tool.py
@@ -149,6 +149,37 @@ class TestSmartContextTool:
         assert "health" in result
 
 
+class TestSmartContextSyntaxGate:
+    """#754: smart_context blended file_health + risk but bypassed the shared
+    syntax gate (it called HealthScorer.score_file directly), so a file that
+    fails to parse scored grade=A / verdict=SAFE — a false green telling an agent
+    it was safe to edit unparseable code. It must short-circuit to the canonical
+    syntax_error envelope like the sibling detection tools.
+    """
+
+    def test_parse_broken_file_is_error_not_safe(self, tool, tmp_path):
+        bad = tmp_path / "broken.py"
+        bad.write_text("def broken(:\n    x =\nclass  :\n")
+        result = _run(tool.execute({"file_path": "broken.py", "output_format": "json"}))
+        assert result.get("verdict") == "ERROR"
+        assert result.get("signal") == "syntax_error"
+        next_step = (result.get("agent_summary") or {}).get("next_step", "").lower()
+        assert "proceed" not in next_step
+        # Shape parity: the happy-path keys are preserved (degraded), so a
+        # consumer reading result["health"]["grade"] etc. does not KeyError.
+        assert result["health"]["grade"] == "N/A"
+        for key in ("exports", "structure", "dependencies", "tests", "language"):
+            assert key in result, key
+
+    def test_clean_file_is_not_gated(self, tool, tmp_path):
+        good = tmp_path / "clean.py"
+        good.write_text("def ok():\n    return 1\n")
+        result = _run(tool.execute({"file_path": "clean.py", "output_format": "json"}))
+        assert result.get("verdict") != "ERROR"
+        assert result.get("signal") != "syntax_error"
+        assert "health" in result
+
+
 class TestExportExtraction:
     def test_extracts_class(self):
         # models.py was decomposed into models/ package; use base.py as the canonical
--- a/tree_sitter_analyzer/mcp/tools/smart_context_tool.py
+++ b/tree_sitter_analyzer/mcp/tools/smart_context_tool.py
@@ -25,6 +25,7 @@ from .utils.element_extractor import (
     get_all_exports,
     get_structure,
 )
+from .utils.parse_validity import is_file_parse_broken, syntax_error_envelope
 from .utils.test_discovery import find_test_files
 
 logger = setup_logger(__name__)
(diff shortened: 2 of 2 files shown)


aimasteracc:

|---|---|---|---|
| **3** | **Bare `syntax_error_envelope` dropped `health`/`exports`/`structure`/`language`/`line_count` — a consumer reading `result["health"]["grade"]` on a broken file now KeyErrors (traded false-green for a crash); `file_health` instead degrades to `grade="N/A"` and the helper docstring says the caller should overlay** | **P2 — valid, fixed** | The envelope now **overlays** smart_context's normal top-level keys degraded: `health.grade="N/A"`, empty `exports/structure/tests`, `dependencies={...}`, `line_count`, `language`, `recommendation`. Test strengthened to assert `health.grade=="N/A"` + key presence. |
| 2 | double `resolve_and_validate_file_path` (execute + _build_profile) | P3 — code smell | No fix: both calls hit the shared resolved-path/security caches (idempotent); cost ≈ two dict lookups. |
| 4 | unknown-language broken files still slip the gate (`is_file_parse_broken` returns False on unknown lang by design) | not this PR | Pre-existing, identical to every other gated tool; belongs on the shared gate, out of #754 scope. |
| — | "branch carries 5 unrelated commits" | not real | Artifact of a stale local `develop` on the reviewer side; `git log origin/develop..HEAD` shows exactly one commit. GitHub's PR diff is the single #754 change. |

Net: minimal syntax-gate wire-in that now preserves full shape on the error path. Re-verified the behavior directly (broken→ERROR/grade=N/A/no-KeyError, clean→ungated); full suite runs in CI (local guard contended by a concurrent run). Good P2 catch on the consumer-shape hazard.

Issue #780, opened 2026-06-14
[P1][dogfood] --check-constraints fabricates 14 phantom 'error' violations from bare method-name collision (execute/update), forcing a project-wide UNSAFE verdict with ZERO real forbidden edges


aimasteracc:
**Severity:** P1  |  **Surface:** constraint checking / edit-safety gating (--check-constraints, the gate that bubbles into edit action=safe & change_impact)
_Found by parallel dogfood audit round 2 on develop @ 471ac5f5; adversarially re-reproduced + dup-checked by an independent verifier._

### Observed
verdict=UNSAFE with 14 error-severity violations across 3 rules (mcp-must-not-depend-on-cli x10, languages-isolated-from-mcp x3, core-must-not-import-mcp x1). Grouped by callee: 10x 'execute' all resolved to cli/commands/base_command.py, 4x 'update' all resolved to mcp/tools/utils/file_health_blocks.py. Every one is a method-name collision: the callers import no forbidden module, the calls are stdlib (hashlib .update) or the caller's own facade.execute(). The constraint engine even flags its own file (constraint_violation_query.py:32). No mcp file imports cli anywhere in the tree, so zero real forbidden edges exist.

### Expected
0 violations and verdict=SAFE (no mcp->cli, core->mcp, or languages->mcp import exists in the tree). The callee resolver must bind a bare call like 'update'/'execute' only to a candidate actually reachable from the caller's import graph, not to any same-named method anywhere in the repo. A constraint gate that emits 14/14 false-positive UNSAFE verdicts is a false-green by desensitization — it trains agents to ignore the gate, the inverse failure mode of the #754 false-SAFE family.

### Repro
```bash
[...]

PR #818, opened 2026-06-14
fix(mcp): safe_to_edit summary_line carries the escalated verdict (#781)

### LINKED ISSUE #781: [P2][dogfood] --safe-to-edit summary_line reports verdict=CAUTION while every structured verdict field says UNSAFE (constraint-violation escalation patches verdict but not summary_line)
**Severity:** P2  |  **Surface:** edit-safety (--safe-to-edit agent_summary.summary_line vs verdict)
_Found by parallel dogfood audit round 2 on develop @ 471ac5f5; adversarially re-reproduced + dup-checked by an independent verifier._

### Observed
Top-level verdict='UNSAFE', risk_level='caution', recommendation starts 'CAUTION: health grade B, 133 downstream file(s)...', agent_summary.verdict='UNSAFE'. But agent_summary.summary_line='tree_sitter_analyzer/mcp/server.py risk=caution verdict=CAUTION health=B tests=yes' and the top-level summary_line is identical -- both say verdict=CAUTION. Root cause (tree_sitter_analyzer/mcp/tools/utils/safe_to_edit_helpers.py): two error-severity mcp-must-not-depend-on-cli constraint violations escalate base_verdict CAUTION -> UNSAFE via _max_verdict (lines 176-178); the patch at lines 188-189 updates summary['verdict'] but not summary['summary_line'], which was built from the un-escalated base_verdict inside build_agent_summary (lines 331-336).

### Expected
summary_line.verdict must equal agent_summary.verdict / top-level verdict on the same call. When a constraint violation (or fixture detection) escalates the verdict beyond the risk-level-derived base_verdict, the summary_line must be rebuilt with the escalated verdict so it reads 'verdict=UNSAFE'. An agent that reads only the one-line compact decision surface (summary_line) currently sees CAUTION (proceed-with-awareness) while the machine-readable verdict says UNSAFE -- a gating disagreement that can let an unsafe edit slip through. The file should be UNSAFE everywhere or CAUTION everywhere, never mixed.

### Repro
```bash
cd /Users/aisheng.yu/git-private/tsa-dogfood && uv run python -m tree_sitter_analyzer tree_sitter_analyzer/mcp/server.py --safe-to-edit --edit-type refactor
```

### Verifier notes
Re-ran the exact repro in /Users/aisheng.yu/git-private/tsa-dogfood: `uv run python -m tree_sitter_analyzer tree_sitter_analyzer/mcp/server.py --safe-to-edit --edit-type refactor`. Output reproduces the finding verbatim: top-level verdict='UNSAFE', risk_level='caution', recommendation='CAUTION: health grade B...', agent_summary.verdict='UNSAFE', but BOTH agent_summary.summary_line AND the top-level summary_line say 'verdict=CAUTION'. Located the exact root cause in tree_sitter_analyzer/mcp/tools/utils/safe_to_edit_helpers.py: at line 162 base_verdict=CAUTION (from risk_level), at lines 176-178 verdict is escalated to UNSAFE via _max_verdict because of two error-severity constraint violations (mcp-must-not-depend-on-cli at server.py:204 and :372). build_agent_summary (lines 331-336) builds summary_line using the UN-escalated base_verdict (CAUTION). The escalation patch at lines 188-189 (`if verdict != base_verdict: summary['verdict'] = verdict`) corrects summary['verdict'] to UNSAFE but does NOT rebuild summary['summary_line'], leaving it at the stale CAUTION. The top-level summary_line is then mirrored from this stale value. Not a flag/user error: --safe-to-edit --edit-type refactor are documented flags with valid args; the divergence is purely an internal escalation-patch omission. Not a CLI-vs-MCP asymmetry or locked design decision. No open issue matches; #754 is a different surface (smart-context false-green on parse failure), #743/#744 are about missing/wrong envelope structure on different commands, not a verdict/summary_line value disagreement.

aimasteracc:
Closes #781 (P2, edit-safety verdict-consistency).

## Root cause
When a constraint-violation/fixture escalation raised the verdict above the risk-derived base (CAUTION/SAFE → UNSAFE via `_max_verdict`), the caller patched `summary["verdict"]` but `build_agent_summary` had already baked the **un-escalated** base verdict into `summary_line`. So `agent_summary.summary_line` (and the mirrored top-level `summary_line`) read `verdict=CAUTION` while the structured `verdict` said `UNSAFE` — a one-line-vs-structured gating disagreement that can let an unsafe edit slip through.

## Fix (single source of truth)
`build_agent_summary` gains a `verdict_override` param so the final (possibly escalated) verdict flows into **both** `summary_line` and `summary["verdict"]`; the partial post-hoc patch is dropped. Non-escalated calls are unchanged (`override == base`).

## Verification
- Repro (RED): an error-severity violation escalates a low-risk file → `verdict=UNSAFE` but `summary_line` said `verdict=SAFE`. After: `summary_line` reads `verdict=UNSAFE`, matching `agent_summary.verdict` and the top-level verdict.
- New `test_summary_line_matches_escalated_verdict` (reuses the existing seed-violation scaffold). Verified end-to-end via the real tool; full suite in CI (local guard contended by a concurrent run). ruff + mypy clean.
- Found by the parallel dogfood audit (round 2, #781); fixed by the QA arm joining the dev effort. Independent opencode review to follow.

Note: the *reason* server.py escalates in the original repro is the phantom constraint violations of #780 — but this fix is orthogonal: after ANY escalation (real or not), summary_line must match the verdict.

### DIFF
--- a/tests/unit/mcp/tools/test_safe_to_edit_unsafe_verdict.py
+++ b/tests/unit/mcp/tools/test_safe_to_edit_unsafe_verdict.py
@@ -215,6 +215,42 @@ class TestSafeToEditConstraintIntegration:
             f"Got risk_factors: {risk_factors}"
         )
 
+    def test_summary_line_matches_escalated_verdict(self, tmp_path: Path) -> None:
+        """#781: an escalated verdict must reach the summary_line too.
+
+        Before the fix, a constraint escalation patched ``summary["verdict"]``
+        to UNSAFE but left ``summary_line`` built from the un-escalated base
+        verdict, so the one-line decision surface read verdict=CAUTION/SAFE
+        while the structured verdict said UNSAFE — a gating disagreement that
+        can let an unsafe edit slip through.
+        """
+        project = _scaffold_min_project(tmp_path)
+        _stage_dogfood_constraints(project)
+
+        db_path = project / ".ast-cache" / "index.db"
+        _init_violations_db(db_path)
+        _seed_violation(
+            db_path,
+            rule_id="mcp-no-cli",
+            caller_file=TARGET_FILE_REL,
+            severity="error",
+            caller_line=2,
+        )
+
+        tool = _make_safe_to_edit_tool(project)
+        result = _run(
+            tool.execute({"file_path": TARGET_FILE_REL, "output_format": "json"})
+        )
+
+        verdict = result["verdict"]
+        assert verdict == "UNSAFE", verdict
+        summary = result["agent_summary"]
+        assert summary["verdict"] == verdict
+        assert f"verdict={verdict}" in summary["summary_line"], summary["summary_line"]
+        # The mirrored top-level summary_line must be present and agree too.
+        assert "summary_line" in result, sorted(result)
+        assert f"verdict={verdict}" in result["summary_line"], result["summary_line"]
+
     def test_safe_to_edit_emits_CAUTION_on_warn_violation(self, tmp_path: Path) -> None:
         """Only warn-severity → CAUTION (not the legacy ``REVIEW``).
 
--- a/tree_sitter_analyzer/mcp/tools/utils/safe_to_edit_helpers.py
+++ b/tree_sitter_analyzer/mcp/tools/utils/safe_to_edit_helpers.py
@@ -182,11 +182,9 @@ def _format_safe_to_edit_result(
     if fixture_fact.is_fixture:
         risk_factors.append(_fixture_risk_factor(fixture_fact, context.file_path))
     recommendation = _format_recommendation(risk, facts, workflow)
-    summary = build_agent_summary(workflow_context, workflow)
-    # Promote agent_summary.verdict when constraint violations escalate
-    # beyond the risk-level-derived verdict.
-    if verdict != base_verdict:
-        summary["verdict"] = verdict
+    # #781: pass the (possibly escalated) verdict so summary_line AND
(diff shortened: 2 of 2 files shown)


aimasteracc:

opencode verdict: **"Nothing real to block merge — correct, minimal, single-source-of-truth, tested."** Both P3 nits applied:

|---|---|---|---|
| Q1 | other callers of `build_agent_summary`? | clean — exactly one caller; new param keyword-default, backward-compatible | none |
| Q2 | `_max_verdict` monotonic so no case lost; **P3 footgun**: `verdict_override` is contract-by-comment, a future caller could pass a lower verdict and downgrade | **applied** | `build_agent_summary` now computes `verdict = _max_verdict(_risk_to_verdict(context.risk), verdict_override)` — override can only ever *raise*, never downgrade (verified: a lower override stays at base). |
| Q3 | top-level `summary_line` mirrored correctly? | clean — `mirror_summary_line` always fires; both surfaces now agree | none |
| Q4 | codecov of new lines | covered by the new test (truthy path) + existing non-escalating tests (falsy path) | none |
| nit | test's top-level assertion was conditional (`if "summary_line" in result`) | **applied** | now unconditional `assert "summary_line" in result` — locks the mirror contract. |

Net: minimal single-source-of-truth fix, now hardened so the override is a floor-raiser only. Re-verified (escalate→UNSAFE, none→base, lower-override→no downgrade); full suite in CI.

PR #819, opened 2026-06-14
fix(mcp): change_impact reconciles top-level verdict with agent_summary (#782)

### LINKED ISSUE #782: [P2][dogfood] --change-impact top-level verdict (INFO) disagrees with agent_summary.verdict (REVIEW) on the same call
**Severity:** P2  |  **Surface:** impact gating (--change-impact verdict vs agent_summary.verdict)
_Found by parallel dogfood audit round 2 on develop @ 471ac5f5; adversarially re-reproduced + dup-checked by an independent verifier._

### Observed
Single --change-impact JSON response (full diff, changed_count=20): top-level verdict='INFO', risk_level='low'; agent_summary.verdict='REVIEW', agent_summary.risk='low'. Top-level summary_line ('change_impact changed=20 risk=low pytest_required=True') contains no verdict token; agent_summary carries verdict=REVIEW. Deterministic across repeated runs. (When scope resolves to 0 changed files — e.g. scoping to an unmodified file like server.py — both surfaces read INFO and agree; the contradiction appears for any changed_count>0.)

### Expected
The top-level verdict and agent_summary.verdict must agree (both INFO or both REVIEW) within a single change-impact response, and the top-level summary_line should carry a verdict token consistent with both. An agent gating on the top-level verdict must not see INFO (benign) while an agent gating on the agent_summary decision surface sees REVIEW (needs review) for the identical call — per the tool's own VERDICT INTEGRITY contract.

### Repro
```bash
cd /Users/aisheng.yu/git-private/tsa-dogfood && uv run python -m tree_sitter_analyzer --change-impact 2>/dev/null | python3 -c "import json,sys; d=json.load(sys.stdin); print('top verdict:',d.get('verdict')); print('top risk_level:',d.get('risk_level')); print('top summary_line:',d.get('summary_line')); a=d.get('agent_summary',{}); print('agent verdict:',a.get('verdict')); print('agent risk:',a.get('risk'))"
# Requires changed_count>0 in scope (workspace has 20 untracked/dirty files here).
# Scoped form from the finding also reproduces when the scoped file is itself changed:
#   --change-impact --change-impact-scope <a-modified-tracked-file>
```

### Verifier notes
Reproduced deterministically (2 runs identical). On the full diff (changed_count=20, risk=low) the same --change-impact response ships top-level verdict='INFO' (with risk_level='low') while agent_summary.verdict='REVIEW' (risk='low'). The top-level summary_line ('change_impact changed=20 risk=low pytest_required=True') carries no verdict token at all, matching the finding. Not a flag/user error: --change-impact is the documented flag invoked exactly as CLAUDE.md prescribes. Root cause is two independent verdict computations: agent_summary.verdict is content-aware (apply_scope_validation in change_impact_response.py:~490-500 sets REVIEW whenever changed_count>0), while the top-level verdict is risk-derived/defaulted to INFO (build_change_impact_response at change_impact_response.py:316 never sets a top-level 'verdict'; central canonicalization yields INFO for risk=low). The reconciler mirror_summary_line (base_tool.py:125) only copies a verdict when EXACTLY ONE surface is populated (top_is_real and not agent_is_real, or vice versa) and is a no-op when both surfaces are real-but-conflicting — so the contradiction ships. This contradicts the tool's own VERDICT INTEGRITY instruction ('agent_summary.verdict is the blast-radius gate ... report that verdict verbatim'): an agent gating on top-level INFO sees benign while one gating on the agent_summary decision surface sees REVIEW. Note: when scoped to a file with 0 scoped changes (e.g. server.py, which is unmodified) both surfaces read INFO and agree — the disagreement manifests for any case with scoped/workspace changed_count>0. Dedup: searched open issues (change-impact verdict / verdict agent_summary / change_impact); nearest are #754 (smart-context false green), #744/#743/#733 (agent_summary omissions on other tools) — none cover an intra-response top-vs-agent_summary verdict contradiction on change-impact; not in #734-#757.

aimasteracc:
Closes #782 (P2, edit-impact verdict-consistency).

## Root cause
The top-level `verdict` (risk-derived, possibly constraint-escalated) and `agent_summary["verdict"]` (content-aware: SAFE/REVIEW/WARN by `changed_count`) are computed by independent steps. `mirror_summary_line` only fills a **missing** verdict, not a divergent one — so a `changed_count>0` diff shipped top-level `verdict=INFO` while `agent_summary` said `REVIEW`. An agent gating on the top level sees "benign"; one gating on the decision surface sees "needs review" — for the identical call.

## Fix (scoped to change_impact)
Reconcile both to the **more severe** (by a verdict-severity rank) at the end of `apply_scope_validation`, which runs before `mirror_summary_line` in every change-impact path. Never downgrades: a constraint-escalated `UNSAFE` top level survives and lifts `agent_summary`; a content-aware `REVIEW` lifts a stale `INFO`. Ties keep the top-level value, so the `changed_count==0` benign case is unchanged.

Deliberately **not** changing the shared `base_tool.mirror_summary_line` (it has the same latent both-present-divergent gap, but that's high blast radius across every tool — a candidate follow-up). The rank is duplicated locally to avoid a circular import (`change_impact_tool` imports this module).

## Verification
- `TestVerdictReconciliation782`: changed>0 → both `REVIEW`; constraint `UNSAFE` not downgraded; changed==0 stays benign + agrees. Verified directly (3 scenarios); full suite in CI. ruff + mypy clean.
- Found by the parallel dogfood audit (round 2, #782). Independent opencode review to follow.

### DIFF
--- a/tests/unit/test_change_impact_response.py
+++ b/tests/unit/test_change_impact_response.py
@@ -3,6 +3,7 @@
 from tree_sitter_analyzer.mcp.tools.utils.change_impact_response import (
     AgentSummaryContext,
     ChangeImpactResponseContext,
+    apply_scope_validation,
     attach_queue_ledger,
     build_agent_summary,
     build_agent_summary_only_response,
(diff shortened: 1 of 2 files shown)


aimasteracc:
## Independent review triage (opencode adversarial arm) — all findings addressed

opencode caught a real **P1** (great catch): my first cut fixed the full-response path but the **compact `agent_summary_only` path** — the surface gating agents read most — recomputes the top-level verdict from `risk_level` *after* the reconciliation, re-opening #782.

|---|---|---|---|
| **P1** | compact `build_agent_summary_only_response` (line 176) recomputes verdict from risk → discards reconciliation; top=INFO / agent=REVIEW returns in compact mode | **Real, FIXED** | Re-run `_reconcile_envelope_verdict` at the end of `build_agent_summary_only_response`. Added `test_compact_response_preserves_reconciled_verdict` (+ the UNSAFE-no-downgrade variant). Verified: compact now REVIEW/REVIEW and UNSAFE/UNSAFE. |
| **P2** | `_VERDICT_SEVERITY` is an unguarded duplicate of `_JOURNAL_VERDICT_RANK` (drift risk) | **Addressed** | Added `test_verdict_severity_rank_matches_journal_rank` asserting byte-equality; comment documents the circular-import rationale. |
| P3 | early-return guard (missing verdict) uncovered | **Addressed** | Added `test_reconcile_no_ops_when_a_verdict_is_missing`. |
| P3 | tie-handling (top=INFO/agent=SAFE keeps INFO) | non-issue (both benign; no-changes path leaves top blank → early-returns) | none |
| Q1/Q4/Q5 | no existing test breaks; ordering correct; coverage | confirmed | covered by the added tests |

Net: both the full-response AND the compact path now keep top-level verdict == agent_summary.verdict (more-severe wins, never downgrade). The shared `mirror_summary_line` has the same latent both-present-divergent gap but is out of scope (high blast radius) — noted as a follow-up. Re-verified all paths directly; full suite in CI.

PR #878, opened 2026-06-15
fix(sql): regex fallback for CREATE TABLE absorbed into ERROR nodes (#808)

### LINKED ISSUE #808: [P2][dogfood] SQL: CREATE TABLE statements silently dropped depending on preceding statements (7 of 11 tables extracted, deterministic)
**Severity:** P2  |  **Surface:** CLI --advanced extraction (sql plugin)
_Found by parallel dogfood audit round 3 on develop @ 471ac5f5; adversarially re-reproduced + dup-checked._

### Observed
small.sql has 11 CREATE TABLE statements but extraction returns only 7 tables: ['Album','Artist','Employee','Genre','InvoiceLine','MediaType','Track']. Missing: Customer, Invoice, Playlist, PlaylistTrack. success:true throughout (no error, no warning) so the ~36% loss is silent. Deterministic across runs (7 both times). Each missing table extracts correctly in isolation (e.g. Customer alone -> 1 table), confirming a context-dependent parse-recovery cascade: an earlier ambiguous/wrapped clause (Customer's multi-line FOREIGN KEY ... ON DELETE/ON UPDATE) produces an ERROR node that absorbs the adjacent CREATE TABLE when preceded by other statements.

### Expected
All 11 CREATE TABLE statements should be extracted as table elements (count == 11), matching ground truth (grep -c 'CREATE TABLE' == 11). A common SQL DDL dump losing tables to parse-recovery should at minimum not silently report success:true with incomplete output.

### Repro
```bash
1. Download Chinook DDL: curl -sL https://raw.githubusercontent.com/lerocha/chinook-database/master/ChinookDatabase/DataSources/Chinook_Sqlite.sql -o chinook.sql
2. head -200 chinook.sql > small.sql  (contains exactly 11 CREATE TABLE; verify: grep -c 'CREATE TABLE' small.sql == 11)
3. Place small.sql inside the project working dir (CLI rejects absolute paths), then run:
   uv run python -m tree_sitter_analyzer small.sql --advanced --output-format json | python3 -c "import json,sys;d=json.load(sys.stdin);print(sorted(e['name'] for e in d['elements'] if e['type']=='table'))"
Minimal repro: sed -n '71,106p' small.sql > min.sql (Album, Artist, Customer) extracts only ['Album','Artist']; Customer alone (lines 88-106) extracts ['Customer']; Customer+Employee (88-128) extracts both.
```

### Verifier notes
Re-ran the exact repro inside the project dir (absolute paths are rejected by the CLI, so small.sql was copied into the repo and referenced relatively). Ground truth grep -c 'CREATE TABLE' small.sql = 11; --advanced --output-format json extraction returns exactly 7 table elements ['Album','Artist','Employee','Genre','InvoiceLine','MediaType','Track'], missing Customer, Invoice, Playlist, PlaylistTrack. success:true throughout, so the loss is silent. Deterministic across 2 runs (7 both times). Minimal cases reproduce exactly: Album+Artist+Customer (lines 71-106) -> only ['Album','Artist']; Customer alone -> ['Customer']; Customer+Employee -> ['Customer','Employee']. So a table that parses fine in isolation is dropped when preceded by earlier CREATE TABLEs -> context-dependent parse-recovery cascade in the tree-sitter SQL grammar (Customer's multi-line wrapped FOREIGN KEY ... ON DELETE/ON UPDATE clause is the likely ERROR-node trigger). Not a flag/user error: --advanced is the correct extraction flag, output-format json is valid, ground truth is unambiguous. Dedup: searched open issues for SQL/table-dropped/parse-recovery/ERROR-node terms; #775 is a different SQL bug (CREATE FUNCTION JSON serialization crash); all other 'dropped'/'silently' issues are non-SQL languages (TS/Rust/PHP/Scala/Bash/JS/Ruby). No overlap. Read-only respected: no commits, temp files removed.

aimasteracc:
## Problem

tree-sitter SQL parse-recovery creates ERROR nodes when a FOREIGN KEY clause with multi-line `ON DELETE` / `ON UPDATE` options triggers grammar ambiguity. The ERROR node absorbs the immediately following `CREATE TABLE` statement, causing it to be silently dropped during AST traversal.

**Reproduction (Chinook-style DDL):** Artist + Album (FK with ON DELETE) + Customer → only 7 of 11 Chinook tables extracted.

## Fix

Added `fill_missing_sql_tables_from_regex()` in `table_extractor.py`:
- Scans raw source text for `CREATE TABLE` patterns after the AST pass
- Appends any tables not already found by AST traversal
- Never removes existing tables — safe to call unconditionally
- Helpers `_find_statement_end_line` + `_extract_statement_text` use paren-depth tracking to bound statement text and line numbers

Called immediately after `_extract_sql_tables()` in `extract_sql_elements()`.

## Tests

Added `TestCreateTableErrorNodeCascade` in `tests/unit/languages/test_sql_fixes.py`:
- `test_all_three_tables_extracted` — Artist + Album + Customer all recovered
- `test_table_count_is_exactly_three` — no duplicates from regex + AST overlap
- `test_customer_table_name_correct` — recovered name is `Customer`, not a fragment
- `test_no_schema_name_on_simple_tables` — schema_name stays None for unqualified tables

All 17 SQL fix tests pass.

## Test plan
- [ ] `uv run pytest tests/unit/languages/test_sql_fixes.py -v` — 17 passed
- [ ] CI green on all platforms

Closes #808

### DIFF
--- a/tests/unit/languages/test_sql_fixes.py
+++ b/tests/unit/languages/test_sql_fixes.py
@@ -209,3 +209,73 @@ class TestCreateTableAsSelect:
         tables = [e for e in elements if isinstance(e, SQLTable)]
         assert len(tables) == 1
         assert tables[0].name == "archive_users"
+
+
+# ---------------------------------------------------------------------------
+# Bug #808 — CREATE TABLE silently dropped when FOREIGN KEY + ON DELETE/UPDATE
+#            triggers tree-sitter parse-recovery ERROR node cascade
+# ---------------------------------------------------------------------------
+
+# Minimal Chinook-style DDL that reproduces the cascade:
+# Artist → Album (FK with ON DELETE RESTRICT) → Customer
+# tree-sitter SQL grammar creates an ERROR node that absorbs Customer when
+# it encounters the multi-line ON DELETE/ON UPDATE clause.  The regex
+# fallback in fill_missing_sql_tables_from_regex must recover it.
+_SQL_FOREIGN_KEY_CASCADE = """
+CREATE TABLE Artist (
+    ArtistId INT NOT NULL,
+    Name VARCHAR(120),
+    CONSTRAINT PK_Artist PRIMARY KEY (ArtistId)
+);
+
+CREATE TABLE Album (
+    AlbumId INT NOT NULL,
+    Title VARCHAR(160) NOT NULL,
+    ArtistId INT NOT NULL,
+    CONSTRAINT PK_Album PRIMARY KEY (AlbumId),
+    FOREIGN KEY (ArtistId) REFERENCES Artist (ArtistId)
+        ON DELETE NO ACTION
+        ON UPDATE NO ACTION
+);
+
+CREATE TABLE Customer (
+    CustomerId INT NOT NULL,
+    FirstName VARCHAR(40) NOT NULL,
+    LastName VARCHAR(20) NOT NULL,
+    CONSTRAINT PK_Customer PRIMARY KEY (CustomerId)
+);
+"""
+
+
+class TestCreateTableErrorNodeCascade:
+    """#808: CREATE TABLE absorbed into ERROR node must be recovered via regex fallback."""
+
+    def test_all_three_tables_extracted(self):
+        """Artist + Album + Customer must all appear — Customer is the cascade victim."""
+        elements = _extract(_SQL_FOREIGN_KEY_CASCADE)
+        tables = [e for e in elements if isinstance(e, SQLTable)]
+        names = {t.name for t in tables}
+        assert names == {"Artist", "Album", "Customer"}
+
+    def test_table_count_is_exactly_three(self):
+        """Exact count — ensures no duplicates from regex + AST overlap."""
+        elements = _extract(_SQL_FOREIGN_KEY_CASCADE)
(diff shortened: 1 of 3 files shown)


aimasteracc:
## Codex P2 triage

**Fixed in latest commit (d99707cc):**
- **P2-1 (L187)** — Skip commented CREATE TABLE matches: added `_is_in_sql_comment()` helper; matches inside `--` or `/* */` comments are skipped.
- **P2-2 (L186)** — De-duplicate by `(schema, name)` pair: `found_keys` is now `set[tuple[str|None, str]]` so `tenant_a.users` and `tenant_b.users` are not collapsed.

**Not fixed in this PR — intentional follow-up scope:**
- **P2-3 (L199)** — Populate columns for recovered tables: Requires re-running `extract_table_columns` on recovered `raw_text`. This is a feature enhancement; the core bug (table silently dropped) is fixed. Follow-up issue to be opened.
- **P2-4 (L22)** — Recover quoted identifiers (`"Order Details"`, `[name]`, `` `name` ``): Requires extending the regex and `is_valid_identifier()`. Narrow use case; rare in practice. Follow-up issue to be opened.

Issue #748, opened 2026-06-14
[P2][dogfood] JS computed-property method [`post`] is extracted with an empty name ('')


aimasteracc:
**Severity:** P2  |  **Surface:** Cross-language extraction correctness (JavaScript) — methods
_Found by the parallel multi-agent dogfood audit on develop @ 471ac5f5; adversarially re-reproduced and dup-checked by an independent verifier agent._

### Observed
The computed-property method `[`post`](endpoint, data)` at lines 328-337 (inside the apiClient object literal in examples/ModernJavaScript.js) is extracted as a method with name: '' (empty string). The object is otherwise complete (parameters endpoint/data, complexity_score 2, visibility public) but the name is empty. No method named 'post' appears anywhere in the output. It is the only empty-named symbol among 32 extracted methods. An empty-name symbol is emitted into the methods list.

### Expected
The method should be named `post` (the computed-property key `[`post`]` is a template literal with no substitutions, so it resolves to the constant string 'post'), or — if computed names can't be statically resolved — it should carry an explicit placeholder label (e.g. '[computed]') rather than an empty string. An empty-name symbol breaks symbol search and navigate-by-name.

### Repro
```bash
cd /Users/aisheng.yu/git-private/tsa-dogfood && uv run python -m tree_sitter_analyzer examples/ModernJavaScript.js --advanced --table json | python3 -c "import json,sys; d=json.load(sys.stdin); print([(m['name'],m['line_range']) for m in d['methods'] if m['name']=='' ])"
```

### Verifier notes
Re-ran the exact repro command and got [('', {'start': 328, 'end': 337})] — confirming a method emitted with name=''. Read examples/ModernJavaScript.js lines 327-337 and verified line 328 is literally `[`post`](endpoint, data)`, a computed-property method with a template-literal key, inside the apiClient object. The emitted method object is fully populated (params endpoint/data, complexity_score 2, lines 328-337) but name is the empty string. No method named 'post' exists in the output (post present: False), so the computed name resolves to '' rather than to 'post'. This is the only empty-name method out of 32 total. Standard flags (--advanced --table json) with no custom config — not a flag/user error or misunderstanding. Dedup search across 'empty name method', 'computed property', 'JavaScript extraction', 'empty', 'JavaScript method name empty' on open issues found no matching issue (#578 is about index-rebuild phantom-empty data, a different surface).

aimasteracc:
Fixed by PR #839 (merged into develop).

Issue #747, opened 2026-06-14
[P2][dogfood] JS captures undeclared RHS identifiers from destructuring as phantom fields while dropping the actual bound names


aimasteracc:
**Severity:** P2  |  **Surface:** Cross-language extraction correctness (JavaScript) — fields
_Found by the parallel multi-agent dogfood audit on develop @ 471ac5f5; adversarially re-reproduced and dup-checked by an independent verifier agent._

### Observed
For `const { name, age, ...otherProps } = userProfile;` (line 27) and `const [first, second, ...rest] = arrayData;` (line 28), TSA captures `userProfile` and `arrayData` (the right-hand-side source identifiers, which are not declared anywhere in this file) as field names, while the actually-bound names `name`, `age`, `otherProps`, `first`, `second`, `rest` are NOT captured. Repro prints `userProfile(RHS)= True arrayData(RHS)= True name(bound)= False first(bound)= False`. Additionally, 5 in-function method-body locals (options@139, result@140, url@313, response@318, contentType@356) are listed as top-level 'fields'.

### Expected
Destructuring should capture the bound left-hand-side names (name, age, otherProps, first, second, rest), never the RHS source initializer expression. Method-body locals should not appear in the top-level fields list. Phantom symbols pointing at undeclared identifiers will mislead any agent doing symbol lookup or rename/impact analysis.

### Repro
```bash
cd /Users/aisheng.yu/git-private/tsa-dogfood && uv run python -m tree_sitter_analyzer examples/ModernJavaScript.js --advanced --table json | python3 -c "import json,sys; d=json.load(sys.stdin); fn=[f['name'] for f in d['fields']]; print('userProfile(RHS)=',('userProfile' in fn),'arrayData(RHS)=',('arrayData' in fn),'name(bound)=',('name' in fn),'first(bound)=',('first' in fn))"
```

### Verifier notes
[...]

PR #939, opened 2026-06-15
fix(js): skip destructuring phantom fields, name computed-property methods (#747 #748)

aimasteracc:
## Summary
- Destructuring patterns in function parameters no longer generate phantom field symbols (#747)
- Computed-property method names (e.g. `[Symbol.iterator]`) are extracted with a meaningful label (#748)

## Test plan
- [ ] `uv run pytest tests/unit/ -q` passes

🤖 Generated with Claude Code


aimasteracc:
Content already merged into develop via PR #839 (commit 73756369)

PR #955, opened 2026-06-15
fix(outline): expose top_level_functions as 'methods' fallback key

aimasteracc:
## Summary
- When `methods` key is absent from outline result, alias it from `top_level_functions` so consumers get consistent schema regardless of whether the file had classes

## Test plan
- [ ] `uv run pytest tests/unit/ -q` passes

🤖 Generated with Claude Code


aimasteracc:
Content already merged into develop via PRs #726 and #926

Issue #779, opened 2026-06-14
[P2][dogfood] Deeply nested functions silently truncated at nesting depth 25 (deeper symbols dropped, no warning)


aimasteracc:
**Severity:** P2  |  **Surface:** CLI --advanced (Python extractor)
_Found by parallel dogfood audit round 2 on develop @ 471ac5f5; adversarially re-reproduced + dup-checked by an independent verifier._

### Observed
Source defines 30 nested functions (L0..L29). Extractor returns only 25 functions (L0..L24); the 5 deepest (L25,L26,L27,L28,L29) are silently dropped. Threshold pinned exactly: defined=25 returns 25, defined=26/27/30/50 all return 25 (deepest are lost and it caps at 25). No warning/diagnostic on stderr (empty) or in the JSON envelope (no truncated/warning/error key); element_count/method_count report the truncated count as if complete. node_count (244) shows the tree IS walked, so this is a nesting-depth recursion guard at 25, not a node-count cap.

### Expected
All 30 nested functions extracted, OR — if a depth guard is intentional — an explicit signal in the envelope (e.g. a truncated=true / dropped_symbols field or a stderr warning) so an agent knows symbols were dropped. Silent truncation makes downstream 'what functions exist' answers wrong with no signal.

### Repro
```bash
cd /Users/aisheng.yu/git-private/tsa-dogfood && mkdir -p _probe_edge && python3 -c "lines=['def L0():']; ind='    '; [lines.append(ind*i+f'def L{i}():') for i in range(1,30)]; lines.append(ind*30+'return 1'); open('_probe_edge/deep_nest.py','w').write('\n'.join(lines)+'\n')" && uv run python -m tree_sitter_analyzer --advanced _probe_edge/deep_nest.py --output-format json | python3 -c "import json,sys;d=json.load(sys.stdin);print(len([e for e in d['elements'] if e['type']=='function']))"
# prints 25 (expected 30). Probe files MUST live under the repo root, not /tmp — absolute /tmp paths are rejected by the SecurityValidator ('Invalid file path: Absolute path must be within project directory').
```

### Verifier notes
[...]

PR #968, opened 2026-06-15
fix(call-path,extractor): BFS forward-meeting, backward dedup, extractor version (#951)

aimasteracc:
## Summary
- Record forward-side meeting nodes in bidirectional BFS so 3+ hop paths reconstruct in correct caller->callee order (previously returned a scrambled, non-executable chain)
- backward_visited already stores hops in forward order, so both reconstruction sites stop reversing the backward segment and dedup by hop signature — removing the duplicate path a shared meeting node produced
- Only run fallback backward search when forward found 0 paths (removes duplicate paths / inflated path_count)
- Bump AST cache extractor version 11 -> 12 so the #779 depth-cap fix applies to already-cached files

Addresses Codex P2s from PR #951.

## Test plan
- `tests/unit/test_call_path.py` — 31 passed (added TestBidirectionalMeetingOrder + TestFallbackBackwardGate regressions: 3+ hop executable order, no-duplicate, fallback gate)
- `test_symbols_json_enrichment.py` version-pin re-pinned 11 -> 12; `test_ast_cache.py` stale-version re-index test green
- nav facade / call_path_enrich / b1 edge-parity suites green (159 combined)

🤖 Generated with Claude Code

### DIFF
--- a/tests/unit/test_call_path.py
+++ b/tests/unit/test_call_path.py
@@ -10,6 +10,7 @@ from tree_sitter_analyzer.call_path import (
     CallPathFinder,
     CallPathResult,
     _files_in_chain,
+    _path_signature,
 )
 from tree_sitter_analyzer.graph.edge_store import EdgeKind, symbol_node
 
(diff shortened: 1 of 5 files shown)


aimasteracc:
@codex review

PR #970, opened 2026-06-15
fix(ast-cache): harden index freshness signals

aimasteracc:
## Summary
- Treat new, deleted, and modified supported source files as stale for AST-index freshness while preserving empty-index warm-cache behavior.
- Stop resolve-only and partial incremental runs from stamping the call-graph-built marker; refresh the marker from complete sweeps and valid single-file reindexes.
- Make callers/callees full-index hints depend on the authoritative built marker, and invalidate symbol-lineage cache before stale AST-index hits.
- Mark the real-git co_change performance fixture slow_ok; its own compute-path <2s assertion remains intact.

Fixes #931
Fixes #932
Fixes #933
Fixes #934
Fixes #935
Fixes #936
Fixes #937

## Verification
- uv run pytest tests/integration/test_ast_cache_lifecycle.py tests/unit/mcp/test_ast_cache_watch_modes.py tests/unit/mcp/test_symbol_lineage_tool.py tests/unit/mcp/tools/test_co_change.py tests/unit/test_ast_cache.py tests/unit/test_ast_cache_bfs.py tests/unit/test_ast_cache_build_state.py tests/unit/test_ast_cache_cascade_search.py tests/unit/test_ast_cache_maintenance.py tests/unit/test_ast_cache_mode_used.py tests/unit/test_ast_cache_tool.py tests/unit/test_ast_cache_utf8_bug.py tests/unit/test_callers_callees_tools.py tests/unit/test_codegraph_callees_tool.py -q
- uv run pytest tests/unit/mcp/test_symbol_lineage_tool.py tests/unit/test_call_graph_built_signal.py tests/unit/test_callers_callees_tools.py tests/unit/test_ast_cache.py --cov=tree_sitter_analyzer --cov-report=json -q
- uv run python scripts/check_patch_coverage.py --base origin/develop --coverage-json coverage.json
- uv run python scripts/generate_facade_actions_doc.py && uv run pytest tests/unit/docs/test_facade_actions_doc_drift.py -x
- uv run pytest -q

PR #980, opened 2026-06-15
fix(ast-cache): built marker on complete mixed incremental index + stale-walk drift guard (#978)

### LINKED ISSUE #978: fix(ast-cache): built-marker false-negative on mixed incremental index + is_ast_index_stale per-call os.walk (from #970 review)
Three P3 follow-ups surfaced during the independent (Opus) review of #970 — all conservative/safe-direction, none blocking, so #970 merged.

1. **`_completed_full_index_sweep` requires `skipped==0`** — a *mixed* incremental `index_project()` (some new + some already-cached files) won't re-stamp the call-graph built marker even when the graph is actually complete, because cached files count as `skipped`. This is a conservative false-*negative* (opposite of the false-positives #970 fixed — safe direction), but it diverges from the single-file path which uses `_indexed_source_files_are_complete()`. Consider using the completeness check in the `index_project` path too.

2. **`is_ast_index_stale` now does a full `os.walk` of the project on every call** (e.g. every lineage `execute()`). Potential perf cost on large repos / hot paths — consider caching the walk result or gating it.

3. **`_walk_supported_source_paths` must stay in sync with the indexer's actual file selection** (`_walk_source_files` + EXT_TO_LANG). If they drift, a file the indexer skips but this walk yields (or vice versa) → permanent false-stale. Add a test pinning the two walks agree.

P3. Codex was rate-limited; found during Opus fallback review.

aimasteracc:
## Summary
- `index_project` now stamps the call-graph built marker when the index is **actually complete** even on a fully-cached re-run (`indexed == 0`). Previously the stamp branch was gated on `indexed > 0`, so a project whose marker was cleared (e.g. predates #708) and is then re-run with everything cached never regained the marker and left callers/lineage permanently hinting `--full-index`.
- The new branch is guarded by `_indexed_source_files_are_complete()` (ast_index == current source set), which returns False for an empty, truncated, errored, or otherwise incomplete index — so #970's false-positive fixes stay intact (truncated/errored/empty runs still do NOT stamp; covered by tests).
- Drift-guard test pinning `_walk_supported_source_paths` (the stale-walk) to the indexer's file selection (`_walk_source_files` + `EXT_TO_LANG`), so a future divergence can't cause permanent false-stale.
- Fix 2 (perf) assessed and **declined**: the per-call `os.walk` in `is_ast_index_stale` is already single-per-call and short-circuits; cross-call memoisation would trade a correctness guarantee for a micro-optimisation. Documented in a code comment. Correctness over premature optimisation.

## Note on the original framing
The brief framed the false-negative as `indexed>0 AND skipped>0`. In the actual code `skipped` only increments for unsupported-language files, but the indexer pre-filters by `EXT_TO_LANG` so `skipped` is structurally 0 in real runs; a mixed run (`indexed=1, cached=1, skipped=0`) already stamps correctly. The genuine residual false-negative is the **fully-cached re-run (`indexed==0`)**, which this PR fixes. The mixed-incremental test is kept to lock that behavior.

## Tests (RED → GREEN)
- `test_fully_cached_rerun_stamps_marker_when_index_complete` — RED on current code (marker stayed False), GREEN after fix.
- `test_mixed_incremental_rerun_keeps_marker_true` — locks mixed-run behavior.
- `test_truncated_rerun_does_not_stamp_marker_when_incomplete` — #970 false-positive guard.
- `test_errored_file_does_not_stamp_marker_when_incomplete` — #970 false-positive guard.
- `test_stale_walk_agrees_with_indexer_file_selection` — Fix 3 drift guard.

29/29 in `test_call_graph_built_signal.py` + 45 in `test_symbol_lineage_tool.py` pass; 275 ast_cache-related tests green. ruff + mypy clean.

Closes #978.

🤖 Generated with Claude Code

### DIFF
--- a/tests/unit/test_call_graph_built_signal.py
+++ b/tests/unit/test_call_graph_built_signal.py
@@ -3,17 +3,26 @@
 from __future__ import annotations
 
 import json
+import os
 import sqlite3
 import subprocess
 import sys
 import time
 from pathlib import Path
 from typing import Any
+from unittest import mock
 
 import pytest
 
 from tree_sitter_analyzer import _ast_cache_callgraph_state as callgraph_state
-from tree_sitter_analyzer.ast_cache import ASTCache
+from tree_sitter_analyzer._graph_cache_fingerprint import (
+    _walk_supported_source_paths,
+)
+from tree_sitter_analyzer.ast_cache import (
+    ASTCache,
+    _language_from_ext,
+    _walk_source_files,
+)
 from tree_sitter_analyzer.mcp.tools.callees_tool import CodeGraphCalleesTool
 from tree_sitter_analyzer.mcp.tools.callers_tool import CodeGraphCallersTool
 from tree_sitter_analyzer.mcp.tools.codegraph_relation_tool import (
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

Issue #1004, opened 2026-06-15
bug(autoindex): --autoindex returns indexed:false + cache_stats:null despite 1955 files in cache (AGENTS.md entry-point lies to agents)


aimasteracc:
From dogfood session on `develop` @ 35d0dac0 (2026-06-16). **Critical dogfood regression** — the AGENTS.md-recommended first-run command lies about index state.

**Symptom**

`--autoindex` (the recommended idempotent cache-status / warm entry point per AGENTS.md) returns an empty envelope:

```
$ uv run python -m tree_sitter_analyzer --autoindex --format json
{
  "success": true,
  "verdict": "INFO",
  "project_root": "/Users/.../tree-sitter-analyzer",
  "indexed": false,         ← LIE: 1955 files are indexed
  "cache_stats": null       ← LIE: stats exist, just not surfaced
}
EXIT: 0
```

Meanwhile the actual cache is fully built:

```
$ uv run python -m tree_sitter_analyzer --ast-cache --ast-cache-mode stats --format json
{
  "indexed_files": 1955,
  "total_symbols": 49613,
  "edges_by_kind": {"calls": 125810, "contains": 21833, "extends": 728, "imports": 9136},
  "fts5_available": true,
  ...
}
```

**Repro**

```
uv run python -m tree_sitter_analyzer --autoindex --format json
# → indexed: false, cache_stats: null   (despite cache having 1955 files)
```

**Why this matters**

1. **AGENTS.md dogfood contract violation, at the front door.** AGENTS.md tells agents: "Run --autoindex to warm the cache." An agent's first call returns `indexed: false` with no stats — so it concludes the cache is missing and proceeds to run `--full-index`.
2. **`--full-index` hangs** (per #1001, >100 s on this same cache). The agent waits forever, concludes TSA is broken, and falls back to grep. The whole dogfood loop collapses at step 1.
3. **Compounds with the #978/#981 built-marker false-negative.** `indexed: false` is almost certainly reading the same false-negative built marker that breaks callers/callees (#981) and lineage determinism (#987). The marker is unreliable, and `--autoindex` trusts it without consulting actual indexed-file counts.
4. **Empty `cache_stats` field.** Even when the cache is genuinely empty, returning `cache_stats: null` gives the agent no signal about why (no project-root mismatch warning, no actionable next_step). Compare with `--ast-cache --ast-cache-mode stats` which always returns rich stats.

**Fix direction**

[...]

Issue #1001, opened 2026-06-15
bug(incremental-sync): --incremental-sync-mode sync/changes hang >100s with zero output on 187-pending cache (status works)


aimasteracc:
From dogfood session on `develop` @ 35d0dac0 (2026-06-16). **Same class as #990** — another warm-cache hang on the project's own codebase.

**Symptom**

`--incremental-sync --incremental-sync-mode sync` and `--incremental-sync-mode changes` both hang with **zero output** on a cache that reports 187 pending changes. The sibling `--incremental-sync-mode status` returns instantly:

```
$ timeout 100 uv run python -m tree_sitter_analyzer --incremental-sync --incremental-sync-mode status --format json
{
  "success": true,
  "verdict": "INFO",
  "mode": "status",
  "indexed_files": 1975,
  "total_symbols": 50300,
  "fts5_available": true,
  "pending_changes": 187,
  "up_to_date": false
}
EXIT: 0     ← returns instantly

$ timeout 100 uv run python -m tree_sitter_analyzer --incremental-sync --incremental-sync-mode sync --format json
EXIT: 124   ← killed at 100s, zero output

$ timeout 100 uv run python -m tree_sitter_analyzer --incremental-sync --incremental-sync-mode changes --format json
EXIT: 124   ← killed at 100s, zero output
```

So the system **knows** there are 187 pending changes (status mode), but actually syncing or even listing them hangs.

**Reproduction environment**

- Branch: `develop` @ `35d0dac0`
- `pending_changes: 187` per status mode
- AST cache: 1,975 indexed files

**Likely culprit (same family as #990)**

The incremental sync path is doing per-file work proportional to `pending_changes`, but with no progress output, no streaming, and no overall timeout. Each pending change may be triggering an inline re-parse + edge re-extraction, which compounds quickly.

The fact that `changes` mode (which should just LIST pending files, not write anything) also hangs suggests the bottleneck is in the **detection** path (computing content hashes for 187 files), not the write path.

**Why this matters**

1. **Agents cannot sync a stale cache.** An agent following a `next_step` hint to "run --incremental-sync" waits forever with no feedback. There's no progress bar, no streaming output, no partial result.
2. **AGENTS.md dogfood contract violation.** The recommended workflow ("develop → run incremental-sync") is broken on TSA's own dogfooded cache.
[...]

Issue #990, opened 2026-06-15
bug(codegraph-metrics): hangs >90s with zero output on warm cache (1955 files / 157K edges); regression test only covers cold cache


aimasteracc:
From dogfood session on `develop` @ 35d0dac0 (2026-06-16). **Critical regression — project dashboard unusable on TSA's own codebase.**

**Symptom**

`--codegraph-metrics` hangs indefinitely (>90s, killed by timeout) with **zero output** on a fully-warm cache. Reproducible via CLI and via direct MCP tool call:

```
$ timeout 90 uv run python -m tree_sitter_analyzer --codegraph-metrics --format json
EXIT: 124      ← timeout, no output

$ timeout 90 uv run python -c "
import asyncio, sys, time
from tree_sitter_analyzer.mcp.tools.codegraph_metrics_tool import CodeGraphMetricsTool
tool = CodeGraphMetricsTool(project_root='.')
async def go():
    sys.stderr.write('START\n'); sys.stderr.flush()
    r = await tool.execute({'sections': ['cache'], 'output_format': 'json'})
    sys.stderr.write('DONE\n'); sys.stderr.flush()
asyncio.run(go())
"
EXIT: 124
stderr: START       ← never reaches DONE
```

Same hang for `--codegraph-metrics-sections cache` (the supposedly cheapest section).

**Environment**

- Branch: `develop` @ `35d0dac0`
- AST cache: warm — 1,955 files indexed, 157,507 edges, 49,613 FTS symbols (per `--ast-cache --ast-cache-mode stats`)
- Project: TSA dogfooded on itself

So this is the **warm-cache path**, not the cold-cache path.

**Existing regression test does NOT cover this**

`tests/e2e/test_mcp_smoke.py:235 test_codegraph_metrics_cold_cache_under_5s` only covers the case where **no AST index exists** (tempdir, cold cache). PR #151 fixed the cold-cache 50 s hang by returning a `'run index first'` hint instead of waiting.

But the warm-cache path — where the index exists and `codegraph_metrics` actually has to compute metrics — was never load-tested, and it hangs on a 1,955-file / 157 K-edge codebase. The regression test passes (cold path is fast) while the production path is broken (warm path is hung).

**Likely culprit**

The `cache` section is supposed to be a fast stats read (the same data that `--ast-cache --ast-cache-mode stats` returns in <1 s). Something in `CodeGraphMetricsTool.execute()` is doing real work instead of reading the stats. Candidates:

[...]

Issue #987, opened 2026-06-15
bug(lineage): --symbol-lineage returns non-deterministic definition_count / reference_count across runs (25/25 → 18/32 → 28/22 for the same query)


aimasteracc:
From dogfood session on `develop` @ 35d0dac0 (2026-06-16). **Critical correctness bug.** Related to #978 / #981 (built-marker false-negative cascade).

**Symptom**

`--symbol-lineage <symbol>` returns **non-deterministic counts** across consecutive runs of the identical query. 5 back-to-back runs:

```
run 1 def: 25 ref: 25 risk_lvl: medium from_cache: False cache_inv: ast_index_stale
run 2 def: 18 ref: 32 risk_lvl: medium from_cache: False cache_inv: ast_index_stale
run 3 def: 28 ref: 22 risk_lvl: medium from_cache: False cache_inv: ast_index_stale
run 4 def: 28 ref: 22 risk_lvl: medium from_cache: False cache_inv: ast_index_stale
run 5 def: 28 ref: 22 risk_lvl: medium from_cache: False cache_inv: ast_index_stale
```

3 distinct count pairs (25/25, 18/32, 28/22) for the same query. Total def+ref is stable at 50, so it's the **classification that's flipping**, not the data set.

Repro (any project with built cache):

```
for i in 1 2 3 4 5; do
  uv run python -m tree_sitter_analyzer --symbol-lineage main --format json | \\
    python3 -c "import json,sys; d=json.loads(sys.stdin.read()); print(d['definition_count'], d['reference_count'])"
done
```

**Root cause (preliminary)**

`tree_sitter_analyzer/mcp/tools/symbol_lineage_tool.py:433-468` `_collect_definitions_and_refs` runs:

1. `execute_find_references` → `(definitions, references)`
2. `_reclassify_definition_like` → promotes definition-like references into `definitions`
3. `_enrich_references_with_callers` → adds call-graph callers to `references`

Steps 2 and 3 mutate the lists. The non-determinism is almost certainly in **step 2's content check** (`_line_looks_like_definition` at line 873 reads source files from disk) interleaved with **step 3's call-graph lookup**, which depends on the call-graph built marker. When the built marker is in a false-negative state (#978), the call-graph lookup falls through to in-memory construction whose iteration order is not stable.

[...]

PR #1006, opened 2026-06-15
fix(ast-cache): call_graph_built edges-table safety net — recover from missing marker (P0 #1005)

### LINKED ISSUE #1005: [P0 root cause] ast_call_graph_state marker missing while edges table has 159K rows — unifies #978/#981/#987/#990/#1001/#1004
From dogfood session on `develop` @ 35d0dac0 (2026-06-16). **Root-cause analysis** of the bug cluster filed today: #978, #981, #987, #990, #1001, #1004.

# The single-paragraph version

The cache at `.ast-cache/index.db` has **159,149 edges** in its `edges` table and **1,955 indexed files**, but the **`ast_call_graph_state` marker table does not exist**. `call_graph_built()` returns `False` on the missing-table path (`_ast_cache_callgraph_state.py:59-60` catches `OperationalError` and returns False). Every consumer that gates on this marker then concludes "the index is empty / missing / stale" and either lies to the agent, hangs trying to rebuild, or produces non-deterministic results.

# Proof

Direct probe of the cache:

```python
$ uv run python -c "
import sqlite3
conn = sqlite3.connect('.ast-cache/index.db')
cur = conn.cursor()
try:
    cur.execute('SELECT * FROM ast_call_graph_state')
    print('state:', cur.fetchall())
except sqlite3.OperationalError as e:
    print('TABLE MISSING:', e)
cur.execute('SELECT COUNT(*) FROM edges')
print('edges:', cur.fetchone())

from tree_sitter_analyzer._ast_cache_callgraph_state import call_graph_built
print('call_graph_built():', call_graph_built(conn))
"
TABLE MISSING: no such table: ast_call_graph_state
edges: (159149,)
call_graph_built(): False
```

Then I manually created the marker and re-ran the tools:

```python
$ uv run python -c "
import sqlite3
from tree_sitter_analyzer._ast_cache_callgraph_state import mark_call_graph_built, call_graph_built
conn = sqlite3.connect('.ast-cache/index.db')
mark_call_graph_built(conn)
print('after:', call_graph_built(conn))
"
after: True

$ uv run python -m tree_sitter_analyzer --callers "nonexistent_xyz_abc" --format json
{
  ...
  "verdict": "NOT_FOUND",
  "next_step": null       ← was: "Call-graph index is empty or has not been built yet. Run --full-index..."
}
```

**Fixing the marker made #981's misleading hint disappear.** The other symptoms (#987 non-determinism, #990 hang, #1001 hang, #1004 lie) follow the same gated path and will improve or disappear with the same fix.

# The broken contract

`tree_sitter_analyzer/_ast_cache_callgraph_state.py:53-65`:

```python
def call_graph_built(conn: sqlite3.Connection) -> bool:
    """Return True iff this cache explicitly records a built call graph."""
    try:
        row = conn.execute(
            "SELECT built FROM ast_call_graph_state WHERE id = 1"
        ).fetchone()
    except sqlite3.OperationalError:
        return False      # ← catches "table missing" and returns False
    if row is None:
        return False
    ...
```

The docstring is honest — it returns True **iff the marker is explicitly set**. But every consumer treats `False` as "the index is missing", which is wrong. `False` can mean any of:

1. The marker table exists but `built=0` (graph truly not built)
2. The marker table doesn't exist (cache built by an older TSA version, or upgraded mid-flight, or partial-write crash)
3. The DB itself is locked / corrupt
4. The graph WAS built but the marker stamp failed silently (per #978, which is about case 4)

All four collapse to `False`, and downstream code can't distinguish them. The result: a cache with 159 K real edges is treated as empty.

# Fix direction (root cause)

Make `call_graph_built()` consult the actual `edges` table as a fallback when the marker is False or missing:

```python
def call_graph_built(conn: sqlite3.Connection) -> bool:
    """Return True iff the cache has a usable call graph.

    A cache is usable when EITHER:
    (a) the built marker is explicitly set, OR
    (b) the edges table has at least one row.

    The marker is the fast path; the edges-table count is the safety net
    that recovers from missing / stale markers (#978, #981, #987, #990,
    #1001, #1004).
    """
    # Fast path: trust the marker if it's set.
    try:
        row = conn.execute(
            "SELECT built FROM ast_call_graph_state WHERE id = 1"
        ).fetchone()
        if row is not None and bool(row[0]):
            return True
    except sqlite3.OperationalError:
        pass  # table missing — fall through to edges-table check

    # Safety net: count edges. If there are any, the graph exists.
    try:
        edge_count = conn.execute("SELECT COUNT(*) FROM edges").fetchone()
        return bool(edge_count and edge_count[0] > 0)
    except sqlite3.OperationalError:
        return False
```

A similar fix is needed for the parallel `is_ast_index_stale()` path (referenced by #978 follow-up #2). The two functions should agree on what "freshly built" means.

# Why this is the highest-leverage fix

Today's dogfood session filed 14 issues. **Six of them** trace to this single root cause:

| Issue | Symptom | Fix trigger |
|---|---|---|
| #978 (open) | mixed incremental index doesn't stamp the marker | layer A: marker stamping fix |
| #981 | callers/callees "index empty" hint with 157K edges | **layer B (this fix): consult edges count** |
| #987 | symbol-lineage non-deterministic counts | layer B + determinism pass |
| #990 | codegraph-metrics hangs >90s warm | layer B (probably; needs verification) |
| #1001 | incremental-sync hangs >100s | layer B (probably) |
| #1004 | --autoindex reports indexed:false + cache_stats:null | layer B + independent stats-population fix |

Landing **layer B** (this fix) closes or significantly advances 5 of the 6. Layer A (#978 itself) prevents the marker from entering the false-negative state in the first place. Both layers are needed; layer B is the defensive must-have because layer A can never guarantee 100% marker integrity across crashes, version upgrades, and partial writes.

# Test gap

```python
def test_call_graph_built_recovers_from_missing_marker_table(tmp_cache_with_edges):
    """Drop the marker table; assert call_graph_built() still returns True
    when edges exist."""
    conn = sqlite3.connect(tmp_cache_with_edges)
    conn.execute("DROP TABLE ast_call_graph_state")
    conn.commit()
    assert call_graph_built(conn) is True, \
        "call_graph_built must fall back to edges-table count"

def test_call_graph_built_false_when_both_marker_and_edges_absent(tmp_empty_cache):
    conn = sqlite3.connect(tmp_empty_cache)
    assert call_graph_built(conn) is False

def test_consumers_agree_with_actual_edge_count(project_with_built_cache):
    """Every consumer of call_graph_built() must agree with the actual
    edges-table count. Pin this so the cluster (#981/#987/#990/#1001/#1004)
    cannot regress."""
    conn = sqlite3.connect(project_with_built_cache)
    edge_count = conn.execute("SELECT COUNT(*) FROM edges").fetchone()[0]
    assert edge_count > 0
    assert call_graph_built(conn) is True
    # And the consumers:
    assert run_callers("missing_symbol")["next_step"] is None  # not the "index empty" hint
    assert run_autoindex()["indexed"] is True
```

# Coordination

- Supersedes the consumer-side fix proposals in #981 (callers/callees), #1004 (autoindex). Those issues should still land their own test guards and message improvements, but the underlying fix lives here.
- Connects to #978 (open). #978 is layer A (don't enter the false-negative state); this is layer B (recover gracefully when we do).
- Recommend this issue be the **P0 tracking issue** for the 2026-06-16 dogfood cluster.

**Priority**

**P0** — single fix, closes or advances 5 P1 issues.

aimasteracc:
## Summary
- `call_graph_built()` now returns True when the marker is set (fast path) OR the `edges` table is populated (safety net). Recovers legacy/crashed caches that have real edges but no `ast_call_graph_state` marker — the root cause behind #981/#987/#990/#1001/#1004 (caches with 159K edges were treated as empty).
- `is_ast_index_stale()` does NOT share the marker-only assumption (it works off `ast_index` mtimes + source-tree completeness, orthogonal to the call-graph marker), so no alignment was needed.

## Implementation
Fast path unchanged: trust the marker iff explicitly set to truthy. Safety net added: when the marker is missing/unset, a single `SELECT COUNT(*) FROM edges` decides — populated to True, empty/absent to False. No source-tree walk (the function only has a `conn` and stays cheap, per #937/#1005).

**Self-heal:** SKIPPED. Stamping the marker from inside a read-path `call_graph_built(conn)` would INSERT+commit on a possibly-shared/read connection, risking write-lock contention with the indexer and concurrent readers. The fallback simply returns True; the marker heals naturally on the next `index_project` run.

## Tests (RED-first, exact `is`/`==`)
- `test_call_graph_built_recovers_from_missing_marker_table` — edges present, no marker table to True (RED on current code, confirmed).
- `test_call_graph_built_marker_set_takes_fast_path` — marker set to True.
- `test_call_graph_built_recovers_when_marker_zero_but_edges_exist` — built=0 + edges to True.
- `test_call_graph_built_false_when_no_marker_and_empty_edges` — empty edges + no marker to False.
- `test_missing_marker_table_with_edges_no_empty_index_hint` (callers+callees) — downstream: `--callers missing` on edges-without-marker cache yields next_step "not in the index", NOT "--full-index".

## Verification
Local test run BLOCKED by the single-instance lock (a parallel full-suite run holds it) — not killed/bypassed per locked rules; relying on CI. Validated all five scenarios + downstream tool path via standalone scripts; ruff + mypy clean (pre-commit passed).

Closes #1005. Should resolve/advance #987 #990 #1001 #1004 (verify each after merge).

🤖 Generated with Claude Code

### DIFF
--- a/tests/unit/mcp/tools/test_nav_not_found_contract.py
+++ b/tests/unit/mcp/tools/test_nav_not_found_contract.py
@@ -136,8 +136,10 @@ async def test_edges_without_built_marker_no_empty_index_hint(
         cache.index_project(workers=0)
         assert cache.has_call_edges() is True
         # Simulate the false-negative: clear the built marker while edges remain.
+        # #1005: the edges-table safety net recovers the cleared marker to True,
+        # so the populated index is never mislabelled "empty".
         callgraph_state.clear_call_graph_built(cache.get_conn())
-        assert cache.call_graph_built() is False
+        assert cache.call_graph_built() is True
     finally:
         cache.close()
 
--- a/tests/unit/test_call_graph_built_signal.py
+++ b/tests/unit/test_call_graph_built_signal.py
@@ -191,7 +191,15 @@ def _seed_partial_ast_cache_without_call_graph(root: Path) -> None:
 
 
 def _seed_call_edges_without_built_marker(root: Path) -> None:
-    """Create real CALLS edges while leaving the authoritative marker false."""
+    """Create real CALLS edges with the marker row cleared (built = 0).
+
+    #1005: clearing the marker row is a false-negative — the edges-table
+    safety net in ``call_graph_built()`` recovers the signal to True, exactly
+    as it does when the marker table is dropped entirely
+    (see ``_seed_call_edges_with_marker_table_dropped``). The tool must still
+    treat the index as populated so a missing symbol is "not in the index",
+    never "index empty".
+    """
     source_path = root / "calls.py"
     source_path.write_text(
         "def caller():\n    target()\n\ndef target():\n    return 1\n",
@@ -202,8 +210,10 @@ def _seed_call_edges_without_built_marker(root: Path) -> None:
         result = cache.index_file(str(source_path))
         assert result["status"] == "indexed"
         assert cache.has_call_edges() is True
+        # Cleared marker row, but edges remain → edges-table safety net (#1005)
+        # recovers the signal to True.
         callgraph_state.clear_call_graph_built(cache.get_conn())
-        assert cache.call_graph_built() is False
+        assert cache.call_graph_built() is True
     finally:
         cache.close()
 
--- a/tests/unit/test_call_graph_built_signal.py
+++ b/tests/unit/test_call_graph_built_signal.py
@@ -85,6 +85,69 @@ def test_call_graph_built_supports_tuple_rows() -> None:
         conn.close()
 
 
+def _make_edges_table(conn: sqlite3.Connection, *, with_row: bool) -> None:
+    """Create a minimal ``edges`` table; optionally seed one CALLS row."""
+    conn.execute(
+        "CREATE TABLE edges ("
+        "id INTEGER PRIMARY KEY AUTOINCREMENT, "
+        "source_node_id TEXT NOT NULL, "
+        "target_node_id TEXT NOT NULL, "
+        "kind TEXT NOT NULL)"
+    )
+    if with_row:
+        conn.execute(
+            "INSERT INTO edges (source_node_id, target_node_id, kind) "
+            "VALUES ('caller', 'callee', 'calls')"
+        )
+    conn.commit()
+
+
+def test_call_graph_built_recovers_from_missing_marker_table() -> None:
+    # #1005 root cause: a legacy/crashed cache can hold a fully populated edges
+    # table with NO ast_call_graph_state marker. call_graph_built() must treat a
+    # populated edges table as a safety net and return True.
+    conn = sqlite3.connect(":memory:")
+    try:
+        _make_edges_table(conn, with_row=True)
+        # No marker table exists at all.
+        assert callgraph_state.call_graph_built(conn) is True
+    finally:
+        conn.close()
+
+
+def test_call_graph_built_marker_set_takes_fast_path() -> None:
+    # Marker explicitly set → True via the fast path (edges table irrelevant).
+    conn = sqlite3.connect(":memory:")
+    try:
+        callgraph_state.mark_call_graph_built(conn)
+        assert callgraph_state.call_graph_built(conn) is True
+    finally:
+        conn.close()
+
+
+def test_call_graph_built_recovers_when_marker_zero_but_edges_exist() -> None:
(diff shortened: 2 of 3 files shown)


aimasteracc:
@codex review

PR #1052, opened 2026-06-18
fix(heatmap): extend complexity heatmap to all plugin languages

aimasteracc:
## Summary

- **Root cause**: `_extract_functions` in `complexity_heatmap.py` fell back to `[]` when `language` was absent from the hardcoded `_METHOD_NODES` map (only 8 languages). C#, Kotlin, Ruby, PHP, and Swift all hit this path, causing `analyze_project_heatmap` to report `total_files_analyzed: 0` with `success: true` — a silent false-success on real projects.
- **Fix**: Added a plugin-sync fallback path `_extract_functions_via_plugin` that calls `PluginManager().get_plugin(language).create_extractor().extract_functions(tree, source)` and maps `Function.complexity_score` → `FunctionComplexity`. The original AST-walk path is unchanged for the 8 hardcoded languages.
- **Honesty fix**: `analyze_project_heatmap` now appends a `"note"` field when `total_functions == 0` so callers receive an explicit signal instead of a clean-but-empty result.

## Test plan

- [ ] All 5 new tests in `TestPluginFallbackLanguages` are RED before the fix and GREEN after (verified locally)
- [ ] All 26 pre-existing tests in `test_complexity_heatmap.py` still pass (31 total, 0 failures)
- [ ] Exact integer assertions: C# fixture → `total_files_analyzed == 1`, `total_functions == 2`, `Add.complexity == 2`; Ruby fixture → `total_files_analyzed == 1`, `total_functions == 2`
- [ ] `test_empty_project_warns` verifies `"note"` key present when 0 functions found
- [ ] Pre-commit hooks passed: ruff, mypy, bandit, codemap-sync

### DIFF
--- a/tests/unit/test_complexity_heatmap.py
+++ b/tests/unit/test_complexity_heatmap.py
@@ -424,3 +424,106 @@ class TestComplexityCLI:
         assert data["mode"] == "function"
         assert data["name"] == "deeply_nested"
         assert data["complexity"] == 8
+
+
+# ---------------------------------------------------------------------------
+# Tests for languages NOT in the hardcoded _METHOD_NODES map (C#, Ruby, …).
+# These exercise the plugin-fallback path added to fix the silent-empty bug.
+# Exact integer counts are pinned per the "exact assertions only" rule.
+# ---------------------------------------------------------------------------
+
+_CSHARP_FIXTURE = """\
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #1055, opened 2026-06-18
fix(plugins): compute cyclomatic complexity for Ruby/Kotlin/PHP

aimasteracc:
## Summary

- **Bug**: Ruby, Kotlin, PHP, and Swift language plugins returned `complexity_score=1` for every function regardless of branching
- **Fix**: Implement decision-point counting (`1 + decisions`) via tree-sitter AST traversal in each plugin
- **Tests**: 9 new passing tests with exact pinned integer assertions; 3 skipped (Swift: tree-sitter-swift not installed)

## Decision points counted

| Language | Statement nodes | Leaf tokens |
|----------|----------------|-------------|
| Ruby | `if`, `elsif`, `unless`, `while`, `until`, `case`, `when`, `rescue`, `for`, `conditional`, `if_modifier`, `unless_modifier`, `while_modifier`, `until_modifier` | `&&`, `\|\|` |
| Kotlin | `if_expression`, `when_expression`, `for_statement`, `while_statement`, `do_while_statement`, `catch_block` | `&&`, `\|\|` |
| PHP | `if_statement`, `else_if_clause`, `switch_statement`, `for_statement`, `foreach_statement`, `while_statement`, `do_statement`, `catch_clause`, `conditional_expression` | `&&`, `\|\|`, `??` |
| Swift | `if_statement`, `guard_statement`, `switch_statement`, `for_statement`, `while_statement`, `repeat_while_statement`, `catch_block`, `ternary_expression`, `conjunction_expression`, `disjunction_expression` | (none — Swift uses dedicated node types) |

## Implementation notes

- Leaf vs non-leaf distinction: some languages (Ruby, Kotlin, PHP) reuse the same type name for both keyword tokens and statement nodes. `len(children) == 0` distinguishes them — keyword tokens are leaves, statement nodes are not.
- `_safe_children()` helper added to each module guards against `Mock()` objects in unit tests (returns `[]` on `TypeError`/`AttributeError`).
- Swift was already partially wired (`_swift_calculate_complexity` existed in `_swift_plugin_elements.py`) but the `complexity_score` field was not included in `_swift_function_fields`. Fixed.

## Test plan

- [ ] `uv run pytest tests/unit/languages/test_cyclomatic_complexity.py -v` → 9 passed, 3 skipped
- [ ] `uv run pytest tests/unit/languages/test_kotlin_target_75.py -v` → all pass (Mock safety regression check)
- [ ] CI green on all platforms

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1057, opened 2026-06-18
fix(plugins): compute cyclomatic complexity for Scala and Bash

aimasteracc:
## Summary

- Scala plugin was returning `complexity_score=1` for every function (no decision-point counting), same gap as Ruby/Kotlin/PHP fixed in #1055
- Bash plugin had the same flat `complexity_score=1` regression
- Adds `calculate_scala_complexity` and `calculate_bash_complexity` with `_safe_children` guards, mirroring the exact pattern from Kotlin/PHP helpers

## Key implementation notes

**Scala** — logical operators appear as `operator_identifier` *named* leaf nodes (not bare `&&`/`||` token types), so the check decodes `.text` rather than comparing `node.type` to `"&&"`. Decision types: `if_expression`, `match_expression`, `for_expression`, `while_expression`, `catch_clause`, `case_clause`.

**Bash** — `while_statement` covers *both* `while` and `until` loops in the grammar (same node type). `&&`/`||` are bare leaf token types directly (type literally equals `"&&"` / `"||"`). Decision types: `if_statement`, `elif_clause`, `while_statement`, `for_statement`, `c_style_for_statement`, `case_item`.

## Test plan

- RED-first: 4 tests failed before implementation (Scala×2, Bash×2)
- All 15 cyclomatic-complexity tests green after fix (Ruby×3 + Kotlin×3 + PHP×3 + Scala×3 + Bash×3)
- Exact integer pins per CLAUDE.md locked rule: `== 1` / `== 4` / `== 12` (no `>=` loose bounds)
- Node types verified by parsing real sample code with `tree_sitter_scala` / `tree_sitter_bash` before writing tests

### DIFF
--- a/tests/unit/languages/test_cyclomatic_complexity.py
+++ b/tests/unit/languages/test_cyclomatic_complexity.py
@@ -330,3 +330,219 @@ class TestPHPCyclomaticComplexity:
         assert len(funcs) == 1
         assert funcs[0].name == "process"
         assert funcs[0].complexity_score == 13
+
+
+# ---------------------------------------------------------------------------
+# Scala
+# ---------------------------------------------------------------------------
+
+SCALA_SIMPLE = """\
+def greet(name: String): String = {
+  s"Hello, $name!"
+}
+"""
+# No branches → complexity = 1
+
+SCALA_BRANCHY = """\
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

PR #1065, opened 2026-06-18
fix(formatters): render top-level functions in PHP --table full output

aimasteracc:
## Summary

- **Bug**: `--advanced --table full` on a PHP file with only top-level functions (no class) rendered only the file title — no function table.
- **Root cause**: `_append_module_level_functions()` in `_php_formatter_helpers.py` had an early `if not classes: return` guard that silently skipped the entire section when the file was purely procedural.
- **Fix**: Remove the 3-line early-return guard. `_module_level_methods()` already handles the empty-classes case correctly (returns all methods when no class ranges exist), so no other change was needed.

## Test plan

- [x] RED-first: both new tests were confirmed FAILING before the fix
- [x] GREEN after fix: `test_top_level_functions_rendered_without_classes` + `test_top_level_functions_exact_row_format` pass
- [x] `test_top_level_functions_exact_row_format` pins exact Cx values (`binary_search_iterative=4`, `greet=1`) per locked exact-assertion rule
- [x] Old `test_no_module_functions_without_classes` (asserting the bug) removed and replaced with the above
- [x] Full PHP/formatter/golden suite: `uv run pytest tests -n 0 -q -k "php or formatter or golden"` → 2717 passed, 0 PHP/formatter failures (2 pre-existing Swift failures unrelated to this change)
- [x] Golden master `php_sample_full.md` unchanged (class-based rendering unaffected)
- [x] End-to-end CLI: `uv run python -m tree_sitter_analyzer /tmp/php_fns.php --advanced --table full --project-root /tmp` now shows the Functions table with Cx column

### DIFF
--- a/tests/unit/formatters/test_php_formatter_helpers.py
+++ b/tests/unit/formatters/test_php_formatter_helpers.py
@@ -9,7 +9,6 @@ from tree_sitter_analyzer.formatters._php_formatter_helpers import (
 
 
 class TestGetVisibilitySymbol:
-
     def test_public(self):
         assert get_visibility_symbol("public") == "+"
 
@@ -30,7 +29,6 @@ class TestGetVisibilitySymbol:
 
 
 class TestFormatSignature:
-
     def test_no_params(self):
         method = {"parameters": [], "return_type": "void"}
         assert format_signature(method) == "():void"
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #1072, opened 2026-06-18
ci: activate Claude PR review bot on main

aimasteracc:
## What

Puts `.github/workflows/claude-review.yml` on the default branch (`main`) so the Claude reviewer added in #1071 actually runs.

## Why a hotfix → main

`anthropics/claude-code-action` has an anti-tampering guard: it only runs when the workflow file is **byte-identical to the version on the repo's default branch (main)**. The bot is inert until the file exists on main, and GitFlow forbids `ci/* → main` — so this uses the `hotfix/*` path (mechanically allowed by gitflow-guard). No release is in flight; develop is not frozen.

The file is verified byte-identical to the copy merged to develop in #1071.

## Effect

After merge, every PR targeting develop is reviewed by Claude (opus-4-8), plus on-demand via `@claude review`. Reuses the existing `ANTHROPIC_API_KEY` secret.

## Test plan
- [x] actionlint clean (pre-commit)
- [x] byte-identical to develop's #1071 version (validated with `diff`)
- [ ] first real review fires on the next develop-targeting PR after merge

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -1,0 +1,115 @@
+# Claude PR Review — a Claude-powered reviewer that guards every PR,
+# the same way the Codex connector does, but driven by this repo's own
+# workflow so it never depends on an external connector's rate window.
+#
+# Triggers:
+#   * automatically on every PR open / new pushes (synchronize) / reopen
+#   * on demand when someone comments "@claude review" on a PR
+#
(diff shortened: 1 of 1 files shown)


aimasteracc:
Closing: pushing a `hotfix/*` branch triggers `hotfix-automation.yml` which auto-publishes to PyPI — wrong for a CI-only change. The review bot will instead ride to main with the next normal release (develop→main), which is the only PyPI-safe path to main. Also switching auth to subscription OAuth token (see follow-up).

Issue #1058, opened 2026-06-18
TOON encoder is type-lossy + no decoder → not round-trippable (blocks raw-TOON wire)


aimasteracc:
## Summary
The TOON encoder is **type-lossy** and the repo has **no TOON decoder**, so our TOON output is not round-trippable — contradicting the "lossless to JSON" property the spec (and RFC-0012) assume. Found during the RFC-0018 dogfood/review on `develop` @ `304db4b4`.

## Evidence (measured)
`_toon_encoder_string_helpers.py:needs_quotes` only quotes on structural chars. A string that *looks* scalar is emitted bare, so a spec-compliant decoder reconstructs the wrong type:

| input (str) | encoded | decodes as | lossless? |
|---|---|---|---|
| `"100.0"` | `100.0` | float | ❌ |
| `"true"` / `"false"` | `true` | bool | ❌ |
| `"null"` | `null` | None | ❌ |
| `"42"` / `"1e5"` / `"-3"` | bare | number | ❌ |

`grep -rn "decode_toon\|from_toon\|ToonDecoder" tree_sitter_analyzer` → **0 hits**: no decoder exists, so the round-trip property is unimplemented and untested.

## Why it matters
This is latent today (we never decode our own TOON), but it **blocks** sending raw TOON on the wire (RFC-0018 Part 2): any string field that looks numeric/boolean (a version, a grade like `"100.0"`, an id) would silently corrupt.

## Fix (RFC-0018 Part 1 — prerequisite)
1. Extend `needs_quotes` to quote any string that would otherwise parse as a TOON number/`true`/`false`/`null` (spec v3.2 §scalars).
2. Add `formatters/toon_decoder.py` (`decode_toon`); can port `xaviviro/python-toon` rather than invent.
3. Round-trip oracle: `assert decode_toon(format_as_toon(x)) == x` over scalars, nested objects, uniform + **sparse/union** arrays (define missing-cell encoding explicitly — emit `null`, not an empty trailing cell).

Tracking: RFC-0018.

aimasteracc:
[...]

PR #1076, opened 2026-06-18
perf(mcp): omit empty agent_next_action guidance fields (RFC-0018 Part 3)

aimasteracc:
## What

Decision-tool `agent_next_action` blocks shipped **empty** placeholder command
fields on every no-op / healthy-file / terminal response:

```
"mcp_command": "", "cli_command": "", "post_edit_commands": []
```

These carry **zero agent signal** but cost tokens on **both surfaces** — the
fields are encoded into `toon_content` as well as the top-level dict — and the
healthy-file path is the *common* case, so the waste is paid on the majority of
`file_health` calls.

This PR omits the command fields **when empty**, at the response **build sites**
(so the TOON wire shrinks too, not just JSON), keeping populated refactor
actions fully intact.

## Why this is the RFC-0018 *Part 3 safe subset* (and not the headline wire flip)

During implementation I verified the headline "raw TOON as the wire default"
(RFC-0018 Part 2) is **not safely shippable yet**: `decode_toon` (#1058) is a
**scalar-only** decoder — it returns the input string unchanged for any TOON
*document* — and the encoder is non-round-trippable in three independent ways
(path normalization, flat-`list[str]`↔list-of-1-key-dict collision, sparse
array-table empty-vs-null ambiguity). Flipping the wire now would leave agents
unable to parse responses. That is a separate encoder/decoder program.

This PR is the **format-agnostic, decoder-independent** slice: pure payload
hygiene that reduces what every decision tool returns, on both surfaces, with
no contract risk.

## Changes
- `file_health_response.py::_build_noop_agent_action` and 3 terminal responses
  in `file_health_tool.py` (empty-file / non-code / syntax-error) omit the
  empty command keys.
- `_health_next_step` made `.get`-safe for the now-optional `cli_command`.
- New RED-first boundary test (`test_omit_empty_guidance_fields.py`) asserts
  absence in **both** the JSON envelope and the `toon_content` blob.
- Two prior tests that pinned the *empty* shape migrated to pin the *lean* shape.

## Scope guard
Only **empty** placeholders are dropped. `_build_refactor_agent_action`
(priority ≠ "none") keeps its real commands — covered by the unchanged
`test_file_health_result_includes_direct_agent_commands_for_smells`.

## Verification
- `tests/unit/mcp` full (5151 passed) + `test_agent_contracts` (78 passed) green
- `ruff` + `ruff format` + `mypy` clean on touched files
- Independent adversarial review: **APPROVE**, 0 P1/P2 (1 P3 test-quality nit fixed)

### DIFF
--- a/tests/unit/mcp/test_file_health_tool.py
+++ b/tests/unit/mcp/test_file_health_tool.py
@@ -96,12 +96,12 @@ def test_file_health_result_marks_healthy_files_as_no_action() -> None:
         "src/healthy.py", health, [], "/tmp/healthy.py", None
     )
 
+    # RFC-0018 Part 3: a no-op action omits the empty command fields
+    # (``mcp_command``/``cli_command``/``post_edit_commands``) — they carry
+    # zero agent signal and cost tokens on every healthy-file response.
     assert result["agent_next_action"] == {
         "priority": "none",
         "reason": "file is healthy enough; no immediate refactor needed",
-        "mcp_command": "",
-        "cli_command": "",
-        "post_edit_commands": [],
     }
     # M10 (round-26): file_health's top-level ``verdict`` mirrors into
     # ``agent_summary`` via ``mirror_summary_line``. Grade A → SAFE.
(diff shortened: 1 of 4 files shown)


aimasteracc:
@claude review

aimasteracc:
@codex

PR #1078, opened 2026-06-18
perf(mcp): drop duplicate 'results' array from trace_impact (RFC-0018 R10)

aimasteracc:
## What

`trace_impact` built **two keys pointing at the same list object**:

```python
"usages": usages,
"results": usages,   # ← exact duplicate (results IS usages)
```

`usages` is the largest field of a trace response, so this duplicate **doubled
the dominant cost of every trace response returned to an agent** for **zero
added signal**. This PR drops the `results` duplicate from both the populated
and not-found envelopes. `usages` is canonical (`usage_count` counts it).

> **Surface note (corrected after review):** `trace_impact` emits a **raw JSON
> dict** — it does not apply TOON encoding — so there was no `toon_content`
> duplication. The saving is on the JSON wire that trace already returns. Still
> a real "adjust the values returned to AI" reduction; just JSON, not TOON.

## Why it's safe (verified per §11, not trusted from the RFC)
- `results is usages` → exact same object; no information lost.
- No src consumer reads trace's `results` — `modification_guard_tool` reads
  `usages`; `search_envelope` (the only generic `results` reader) is **never**
  called by trace (confirmed: `displayed_count` is never set on trace).
- The cross-tool `results` search contract (`test_intent_aliases_integration`)
  targets `search_content` / `list_files`, **not** trace.
- `count` / `usage_count` / `source_call_count` preserve all count signal.

## Verification
- RED-first conformance test (populated + not-found): `results` absent, `usages` canonical.
- **R11 cost invariant**: the emitted JSON response is strictly smaller without
  the duplicate, and the saving is ≥ the serialized size of the `usages` array
  (relationship form, no hand-waved ceiling — per CLAUDE.md §11/§0).
- Full `tests/unit/mcp` + `tests/integration/mcp` + `test_agent_contracts` +
  golden masters: **green** (5481 passed).
- `ruff` / `mypy` clean.
- Independent adversarial review: **APPROVE**, 0 P1/P2 (2 P3 docs/coverage notes addressed).

## RFC context
RFC-0018 R10 ("no duplicate top-level arrays"), the ship-first / zero
wire-encoder-dependency slice. Independent of the blocked raw-TOON-wire flip.

### DIFF
--- a/tests/unit/mcp/test_trace_no_duplicate_results_array.py
+++ b/tests/unit/mcp/test_trace_no_duplicate_results_array.py
@@ -1,0 +1,92 @@
+"""RFC-0018 R10 — trace_impact must not ship a duplicate ``results`` array.
+
+``trace_impact`` built ``"usages": usages`` and ``"results": usages`` — the
+SAME list object under two keys. The array is the largest part of the
+response and is re-encoded into ``toon_content`` as well, so the duplicate
+doubled the dominant cost on both surfaces for zero added signal.
+
+``usages`` is the canonical key (``usage_count`` counts it). ``results``
+(the cross-tool *search* key) is dropped from trace; trace does not route
+through ``search_envelope``, so nothing downstream depends on it.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import os
+import tempfile
+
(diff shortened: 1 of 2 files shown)


aimasteracc:
@claude review

PR #1064, opened 2026-06-18
fix(formatters): add Cx column to Go and Bash full table output

aimasteracc:
## Summary

- **Go**: `_format_func_row` and `_format_method_row` now include `complexity_score`; `_go_formatter_helpers` headers updated from 5-column to 6-column layout (`| Func | Signature | Vis | Lines | Cx | Doc |` for functions, `| Receiver | Func | Signature | Vis | Lines | Cx | Doc |` for methods)
- **Bash**: new `BashTableFormatter` (registered for `"bash"` / `"sh"`) renders the standard 6-column method table (`| Name | Signature | Vis | Lines | Cx | Doc |` full, `| Name | Sig | V | L | Cx | Doc |` compact), replacing the legacy fallback that had no Cx column
- **Golden master** `tests/golden_masters/full/go_sample_full.md` regenerated with Cx column (all 20 functions have complexity_score = 1)
- **Codemap** `docs/CODEMAPS/formatters.md` updated to document `GoTableFormatter` and `BashTableFormatter`

**Audit result**: all other languages (Java, Python, JS/TS, PHP, Ruby, C#, C/C++, Kotlin, Rust, Scala, Swift) already rendered the Cx column. Only Go and Bash were missing it.

## Test plan

- [ ] `tests/unit/formatters/test_cx_column_coverage.py` — 15 exact-pin assertions
  - Go function header: `| Func | Signature | Vis | Lines | Cx | Doc |`
  - Go function row Cx=1: `| Compute | (n int) int | exported | 5-12 | 1 | - |`
  - Go function row Cx=5: `| Compute | (n int) int | exported | 5-12 | 5 | - |`
  - Go method header: `| Receiver | Func | Signature | Vis | Lines | Cx | Doc |`
  - Go method row Cx=3: `| Service | Run | () error | exported | 10-20 | 3 | - |`
  - Negative: old 5-column header `| Func | Signature | Vis | Lines | Doc |` absent
  - Bash function header: `| Name | Signature | Vis | Lines | Cx | Doc |`
  - Bash function row Cx=1: `| deploy | () | public | 3-15 | 1 | - |`
  - Bash function row Cx=4: `| deploy | () | public | 3-15 | 4 | - |`
  - Bash compact header: `| Name | Sig | V | L | Cx | Doc |`
  - Bash compact row Cx=2: `| deploy | (0) | + | 3-15 | 2 | - |`
  - Registry resolves `"bash"` → `BashTableFormatter`
[...]

PR #1082, opened 2026-06-18
feat(rust): show Cx (complexity) column in full table

aimasteracc:
## Why

The Go full table exposes a **Cx** column (#1064), but the Rust full table did not — so the Rust cyclomatic complexity computed by #1081 was **invisible in the markdown table** (only reachable via JSON/MCP/csv). This closes the parity gap so the complexity fix is actually visible to humans reading the Rust table.

## Change

Add a `Cx` column to the Rust full Functions table, matching Go's `Lines | Cx | Doc` ordering:
```
| Function | Signature | Vis | Async | Lines | Cx | Doc |
```

Compact table left unchanged — Go's compact table has no Cx either (parity).

## Tests (RED-first)
- `test_full_table_has_complexity_column` (header carries "Cx"; row shows the value)
- Re-pinned `rust_sample_full.md` golden (new column only)

## Verification
formatters + integration golden → 2330 passed

### DIFF
--- a/tests/golden_masters/full/rust_sample_full.md
+++ b/tests/golden_masters/full/rust_sample_full.md
@@ -10,14 +10,14 @@
 | BorrowedItem | struct | pub | 77-79 | 1 | - |
 
 ## Functions
-| Function | Signature | Vis | Async | Lines | Doc |
-|----------|-----------|-----|-------|-------|-----|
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex

PR #1085, opened 2026-06-18
fix(complexity): count && / || in Java/C/C++/C# cyclomatic complexity

aimasteracc:
The C-family plugins (Java/C/C++/C#) counted control-flow statements but **not** the `&&` / `||` short-circuit operators, while every other plugin (Python/JS/TS/Go/Rust/Ruby/Kotlin/PHP/Scala/Swift/Bash) does. This under-counted cyclomatic complexity for those four languages, corrupting the complexity heatmap, `project_health` grading, and hotspot ranking — the same under-counting class fixed for Go (#1080) and Rust (#1081), one level deeper.

## What changed
Count each logical `&&` / `||` as one decision point, but **only when it is an operand of a `binary_expression`**. That guard excludes the C++ rvalue-reference `&&` (parent `reference_declarator`, e.g. move constructors and `T&&` parameters), which is a type token, not a branch, and must not inflate complexity.

Empirical before/after (probe with one `&&` + one `||`):

| Lang | before | after |
|---|---|---|
| Java | 6 | 8 |
| C | 8 | 10 |
| C++ | 8 | 10 |
| C# | 9 | 11 |

C++ move ctor `R(R&& o)` / `T&&` params correctly stay **1** (rvalue-ref guard). Java/JS/TS/Go/Rust/Ruby/Kotlin/PHP/Scala/Swift/Bash already counted booleans and are unchanged.

## Tests
RED-first `TestJava/C/Cpp/CSharpCyclomaticComplexity` in `test_cyclomatic_complexity.py`, mirroring the existing Go/Rust classes: `simple=1`, `binary_search=4` (boolean-free control, unchanged by the fix), `rich` incl. one `&&` + one `||` with **exact** pins. Re-pinned the `cpp_sample` / `csharp_sample` / `java_bigservice` / `java_userservice` goldens (full+compact+csv, complexity-column only) and the formatter-integration `validateUser` assertions (`3` → `5`, deterministic re-pin). Full suite green.

## Follow-up (separate PR, not in scope here)
Found during this sweep but **not** fixed here to keep the PR focused: Java never counts `switch` or the ternary because its decision set uses the stale node-type names `switch_statement` / `conditional_expression`, but tree-sitter-java emits `switch_expression` / `ternary_expression`. That is a different bug class (wrong node-type names, not missing boolean ops) and will get its own PR.

Found by cross-language complexity dogfood.

### DIFF
--- a/docs/CODEMAPS/languages.md
+++ b/docs/CODEMAPS/languages.md
@@ -39,6 +39,12 @@ Not every registered plugin is wired into the indexer to the same depth:
 | JSON | `languages/json_plugin.py` | inline | basic structure |
 | Bash | `languages/bash_plugin.py` | inline | functions, commands |
 
+## Shared helpers
+
+Cross-cutting logic shared by several plugins (not a language plugin itself):
+
+- `languages/_complexity_logical.py` — `is_executable_logical_operator()`: counts a `&&`/`||` token toward cyclomatic complexity only when it drives executable control flow. Used by the C/C++/C#/Java walkers to exclude booleans in non-executable contexts (`noexcept`/`requires` specifiers, `#if A && B` preprocessor conditions, default arguments, attributes/annotations, `static_assert`). The cross-language convention is "1 + decision points; each `&&`/`||` is one decision; switch/match counts once" (matching Go/Rust/Swift).
+
 ## Plugin Contract
 
 Every language plugin implements:
--- a/tests/golden_masters/compact/cpp_sample_compact.md
+++ b/tests/golden_masters/compact/cpp_sample_compact.md
@@ -30,7 +30,7 @@
 | name | ():str | + | 106-106 | 1 | - |
 | width | ():d | + | 109-109 | 1 | - |
 | height | ():d | + | 110-110 | 1 | - |
-| operator== | (const Rectangle&):b | + | 113-115 | 1 | - |
+| operator== | (const Rectangle&):b | + | 113-115 | 2 | - |
 | square | (d):Rectangle | + | 118-120 | 1 | - |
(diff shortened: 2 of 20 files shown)


aimasteracc:
@codex review

aimasteracc:
## Codex review triage (commit 67b06c96)

All 4 findings verified **real** and **fixed**. Each was a `&&`/`||` in a non-executable context inflating an otherwise branch-free function from Cx 1 → 2.

|---|---|---|---|
| 1 | P2 C++ `noexcept(a && b)` / trailing `requires (A && B)` | ✅ Real (empirically Cx 2) | anchor guard excludes `noexcept` / `requires_clause` |
| 2 | P2 C preprocessor `#if A && B` | ✅ Real | anchor guard excludes `preproc_if` / `preproc_elif` |
| 3 | P2 C++ preprocessor `#if A && B` | ✅ Real | same |
| 4 | P3 C# non-body decl (`[Attr(a&&b)]`, default arg) | ✅ Real | anchor guard excludes `attribute_argument` / `parameter` |

Also caught and fixed the same class for **C++ default arguments** (`bool f = a && b`), **C++ `static_assert`**, and **C# `#if A && B`** that the report didn't enumerate.

Mechanism: new shared `is_executable_logical_operator()` walks from the `&&`/`||` past chained `binary_expression`/parenthesis nodes to the expression *anchor*, and counts the operator only when that anchor is not a per-language non-executable node. Distinguishes a preproc *condition* boolean (skipped) from a real `if (a && b)` *inside* a `#if` guard (counted). 9 new edge tests pin Cx 1 for each context; a body-boolean test guards against over-correction.

@codex review

PR #1087, opened 2026-06-19
fix(complexity): Java was silently ignoring switch / ternary / do-while

aimasteracc:
## Problem

A cross-language cyclomatic-complexity parity audit (same logical function run through every language) found Java under-counting by 3 vs the C# reference:

| Lang | Cx | |
|---|---|---|
| C# | 10 | ✓ reference |
| **Java** | **7** | missing switch + ternary + do-while |

`_JAVA_DECISION_NODES` listed two node types **tree-sitter-java never emits**:
- `switch_statement` — the grammar produces **`switch_expression`**
- `conditional_expression` — the grammar produces **`ternary_expression`**

…and omitted **`do_statement`** entirely. The dead names silently never matched, so Java cyclomatic complexity dropped **every switch, every ternary, and every do-while loop** — under-counting the flagship language and corrupting the complexity heatmap / `project_health` grade / hotspot ranking.

## Fix

Use the real node names (`switch_expression`, `ternary_expression`, `do_statement`), verified against the grammar. Switch counts **once** per the cross-language construct-once convention (matching Go/Rust/Swift/C#).

## Tests (RED-first)

Each construct pinned **in isolation** (switch-only / ternary-only / do-while-only fixtures with exact Cx) so a future miss can't hide behind another construct's count — plus a combined fixture. Re-pinned `java_bigservice` full/compact/csv goldens: three switch-bearing methods rose by 1 (performBackup 1→2, generateReport 2→3, manageInventory 5→6). TOON golden unaffected (complexity normalized out). Local suite green (4239 passed).

@codex review

### DIFF
--- a/tests/golden_masters/compact/java_bigservice_compact.md
+++ b/tests/golden_masters/compact/java_bigservice_compact.md
@@ -29,11 +29,11 @@
 | cleanupCache | ():void | - | 313-326 | 3 | - |
 | updateCacheStatistics | ():void | - | 331-335 | 1 | - |
 | calculateCacheHitRatio | ():d | - | 340-343 | 1 | - |
-| performBackup | (S):void | + | 348-367 | 1 | - |
+| performBackup | (S):void | + | 348-367 | 2 | - |
 | performFullBackup | ():void | - | 372-395 | 6 | - |
 | performIncrementalBackup | ():void | - | 400-414 | 4 | - |
 | performDifferentialBackup | ():void | - | 419-433 | 4 | - |
-| generateReport | (S,M<S, O>):void | + | 438-471 | 2 | - |
+| generateReport | (S,M<S, O>):void | + | 438-471 | 3 | - |
 | prepareReportData | (S,M<S, O>):void | - | 476-490 | 4 | - |
 | generateSalesReport | (M<S, O>):void | - | 495-513 | 6 | - |
 | generatePerformanceReport | (M<S, O>):void | - | 518-534 | 5 | - |
@@ -44,7 +44,7 @@
 | getCustomerInfo | (S):M<S, O> | + | 622-652 | 6 | - |
 | deleteCustomer | (S):b | + | 657-691 | 10 | - |
 | processOrder | (S,L<S>,d):S | + | 696-736 | 12 | - |
-| manageInventory | (S,S,i):void | + | 741-773 | 5 | - |
+| manageInventory | (S,S,i):void | + | 741-773 | 6 | - |
 | addInventory | (S,i):void | - | 778-792 | 4 | - |
 | removeInventory | (S,i):void | - | 797-811 | 4 | - |
 | updateInventory | (S,i):void | - | 816-830 | 4 | - |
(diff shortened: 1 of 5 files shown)


aimasteracc:
**Independent triage of the Codex P1 — verdict: REAL, and broader than stated. This PR is correct but it is not the whole fix.**

Confirmed the multi-path inconsistency empirically. Complexity is counted in **at least four independent places**, and the extractor fix (this PR + #1085) only touches one:

1. **extractor** — `_java_ast_helpers.py` (this PR) ✅ now construct-once with correct names.
2. **`health_scorer.py`** — has its OWN `DECISION_NODE_TYPES: dict[str, set[str]]`; the `"java"` set still lists the dead `switch_statement` / `conditional_expression` names, so `project_health` grading undercounts every Java switch/ternary exactly like the extractor did pre-fix. ❌
3. **`complexity_heatmap.py` / `_ast_extraction.py`** — count `switch_block_statement_group` **per case**, so `analyze_file_complexity(..., 'java')` returns Cx=5 for a 4-case switch where the extractor now returns 2. ❌ (different bug: per-case vs construct-once.)
4. **`complexity_heatmap_tool.py`** — consumes #3.

So after this PR, `BigService.performBackup` is Cx=2 in the golden/formatter path but Cx≠2 in the heatmap/health path — the user-facing heatmap / hotspot / health grade (the entire motivation for #1080–#1087) stay inconsistent.

**This PR is still correct and should land** — it fixes the extractor path and worsens nothing. But the P1 is a genuine, separate, systemic gap: **complexity has no single source of truth.** Recommend a focused follow-up that either (a) routes `health_scorer` / `complexity_heatmap` / `_ast_extraction` through one shared decision-node table, or at minimum re-pins their Java sets to `switch_expression` / `ternary_expression` / `do_statement` with the construct-once convention. Given it spans the public heatmap/health tools and changes grades, it likely warrants an RFC (per CLAUDE.md). Filing this as a tracked finding so it isn't lost.

aimasteracc:
## Codex P1 triage — REAL, and bigger than Java

Verified. `analyze_file_complexity(BigService.java, 'java')` reports `performBackup` Cx=5 (heatmap path) while the extractor/formatter now reports Cx=2. **Real divergence.**

But the root cause is architectural and **not Java-specific**: there are **three independent complexity engines** that disagree for *every* language:

1. **Plugin extractors** (`_java_ast_helpers` etc.) — the canonical `complexity_score`, which this PR fixes.
2. **`complexity_heatmap._COMPLEXITY_NODES`** — self-walks with its own per-language node sets: Java counts `switch_block_statement_group` per case + `else_clause`; **c/go/rust count *every* `binary_expression`** (so `a+b`, `a<b` all inflate CC); js/ts count `switch_case`+`switch_default` per case.
3. **`health_scorer.DECISION_NODE_TYPES`** — a third copy, still listing the dead `switch_statement` name.

This PR correctly fixes engine #1 (the `--table`/structure/MCP surface agents read most). Fixing only Java's row in #2/#3 would leave c/go/rust/js/ts heatmaps just as wrong.

**Verdict:** P1 is real but is its own focused change — **unify #2 and #3 onto the extractor** (`complexity_heatmap._extract_functions` already has `_extract_functions_via_plugin`; route all languages through it and retire the duplicate tables). I'm opening that as the immediate follow-up PR. This PR does not regress the heatmap (it was already divergent) and is a correct, standalone improvement to the canonical engine. Merging now; unification PR follows right after per the P1-before-or-right-after-merge rule.

aimasteracc:
Tracked the systemic P1 follow-up as #1094 (the multi-path complexity inconsistency — health_scorer / heatmap / ast_extraction still disagree with the extractor). This PR itself is correct and shouldn't be blocked on it; #1094 captures the plane-level fix.

PR #1086, opened 2026-06-19
refactor(formatters): remove dead IStructureFormatter abstraction (zero implementers)

aimasteracc:
## What
Removes the dead `IStructureFormatter` ABC — **zero implementers repo-wide**.

## Why
Ponytail audit #1083 finding #1 (dead abstraction). `IStructureFormatter` had one abstract method `format_structure` and **no subclasses** (verified: `grep -rn '(IStructureFormatter)' --include='*.py' .` → 0). It was:
- imported `# noqa: F401` (explicitly known-unused) in `formatter_registry.py`
- re-exported in `formatters/__init__.py` (import + `__all__`)
- documented as 'legacy adapters' in `docs/CODEMAPS/formatters.md`

The live `format_structure` lives on `BaseFormatter`; nothing routes through this interface.

## Changes (pure dead-weight removal, no behavior change)
- delete the `IStructureFormatter` class + its now-unused `from typing import Any` import
- drop the registry re-export (and its now-unneeded `noqa`; `IFormatter` stays — it's used)
- drop the `__init__` import + `__all__` entry
- drop the codemap row; fix 're-exports both' → 'IFormatter'

## Test plan
- [x] `grep IStructureFormatter tree_sitter_analyzer/ docs/` → zero remaining
- [x] `ruff check` on touched files → clean
- [x] import smoke: `from tree_sitter_analyzer.formatters import IFormatter, FormatterRegistry, JsonFormatter` OK
- [x] `pytest tests/unit/formatters/` → 2252 passed, 7 skipped
- [x] no tests reference `IStructureFormatter` (only an auto-regenerated hypothesis constants cache)

Net: ~-10 LOC. First of the #1083 junk-removal cuts.

Issue #1083, opened 2026-06-18
Ponytail audit: 1 dead abstraction (IStructureFormatter) + 2 inert API params + import bloat (~-100 LOC)


aimasteracc:
**Lens:** [Ponytail](https://github.com/DietrichGebert/ponytail) over-engineering review — hunts dead code, reinvented stdlib, YAGNI, verbose logic. Not a correctness/security/perf pass. Findings below are verified against `develop` @ `0ff2f4b5`. Format: `file:Lline: tag what. replacement.`

### Findings

**1. `tree_sitter_analyzer/formatters/_formatter_interface.py:L24` — `delete:` dead abstraction**
`IStructureFormatter` (ABC, one abstract method `format_structure`) has **zero implementers** repo-wide. It's imported `# noqa: F401` (i.e. explicitly known-unused) in `formatter_registry.py:L15`, re-exported in `formatters/__init__.py:L35,L44`, and documented as "legacy adapters" in `docs/CODEMAPS/formatters.md:L40`. The live `format_structure` lives on `BaseFormatter` (`base_formatter.py`), which nothing routes through this interface.
Verify: `grep -rn "(IStructureFormatter)" . --include='*.py'` → 0 subclasses.
Replacement: delete the class + its re-export + the noqa import + the codemap row.

**2. `tree_sitter_analyzer/api.py:L90,L92` — `yagni:` params accepted then dropped**
`analyze_file()` declares `include_details=False  # backward compatibility` and `include_complexity=False  # backward compatibility`, but the body calls `_analyze_file_sync(file_path, language, queries, include_elements, include_queries)` — neither param is forwarded, and `_analyze_file_sync` (L60) doesn't accept them. So a caller like `profile_analysis.py:L25` (`analyze_file(..., include_complexity=True, include_details=True)`) gets neither. Dead flexibility: the knobs do nothing.
Replacement: either delete both params (they're inert), or wire them through `_analyze_file_sync` → `AnalysisRequest` if the high-level API is meant to honor them. As written they're pure surface area. (Note: the *wire-it* option crosses into correctness; the *delete* option is the pure over-engineering cut. Maintainer's call which.)

**3. `tree_sitter_analyzer/default_table_formatter.py:L11-77` & `legacy_table_formatter.py` — `shrink:` import bloat**
[...]

PR #1088, opened 2026-06-19
refactor(api): remove inert include_details/include_complexity params from analyze_file

aimasteracc:
## What
Removes two dead params from the public `api.analyze_file()` wrapper.

## Why
Ponytail audit #1083 finding #2. `analyze_file()` declared `include_details` and `include_complexity` (both commented `# Add for backward compatibility`), but the body calls `_analyze_file_sync()` **without forwarding them**, and `_analyze_file_sync` doesn't accept them — the two knobs did **literally nothing**.

## Verified no caller exercised them
The only two call sites that mention these names build an `AnalysisRequest` directly (the real, honored path) — **not** `api.analyze_file`:
- `grammar_coverage/validator.py:309` → `AnalysisRequest(include_complexity=False, include_details=True)` then `plugin.analyze_file(path, request)`
- `profile_analysis.py:25` → `AnalysisRequest(include_complexity=True, include_details=True)`

No test passes these to `api.analyze_file`. `analyze_file` is not a top-level package export.

> Note: #1083's wording attributed `profile_analysis.py` to the `analyze_file` wrapper — that was a misread; it uses `AnalysisRequest`. Net effect is the same: the wrapper params have **zero** real callers.

## Changes
- delete the 2 inert params + the stale `include_complexity` docstring line. Behavior-preserving (they were no-ops).

## Test plan
- [x] `ruff check api.py` clean
- [x] signature smoke: params now `[file_path, language, queries, include_elements, include_queries]`; `analyze_file('…')` still works
- [x] api/engine unit suite green (137 passed) in main checkout pre-isolation
- [x] no test passes the removed kwargs to `analyze_file`

Net: -3 LOC. Second of the #1083 junk cuts (after #1086). #1083 finding #3 (import "bloat") is **rejected** separately — it is ruff-enforced (`combine-as-imports=false` splits as-imports one-per-line), not accidental bloat.

### DIFF
--- a/tests/integration/core/test_api.py
+++ b/tests/integration/core/test_api.py
@@ -118,11 +118,9 @@ if __name__ == "__main__":
             Path(sample_java_file).unlink()
 
     def test_analyze_file_with_options(self, sample_java_file: str) -> None:
-        """Test file analysis with additional options"""
+        """Test file analysis returns a valid result dict for a string path"""
         try:
-            result = api.analyze_file(
-                sample_java_file, include_complexity=True, include_details=True
-            )
+            result = api.analyze_file(sample_java_file)
 
             assert isinstance(result, dict)
             assert result["success"] is True
--- a/tree_sitter_analyzer/api.py
+++ b/tree_sitter_analyzer/api.py
@@ -87,9 +87,7 @@ def analyze_file(
     language: str | None = None,
     queries: list[str] | None = None,
     include_elements: bool = True,
-    include_details: bool = False,  # Add for backward compatibility
     include_queries: bool = True,
-    include_complexity: bool = False,  # Add for backward compatibility
 ) -> dict[str, Any]:
     """
     Analyze a source code file.
@@ -103,7 +101,6 @@ def analyze_file(
         queries: List of query names to execute (all available if not specified)
         include_elements: Whether to extract code elements
         include_queries: Whether to execute queries
-        include_complexity: Whether to include complexity metrics (backward compatibility)
 
     Returns:
         Analysis results dictionary


aimasteracc:
**Independent verification (review arm) — verdict: SAFE to merge.**

Traced the "inert" claim against the whole tree:
- `analyze_file`'s body calls `_analyze_file_sync(file_path, language, queries, include_elements, include_queries)` — it never forwards `include_details` / `include_complexity`, so they are genuinely no-ops. ✅
- The scary-looking `grammar_coverage/validator.py:309-310` (`include_complexity=False, include_details=True`) targets **`AnalysisRequest(...)`**, a different object that legitimately owns those fields (`core/request.py`) — NOT `api.analyze_file`. Unaffected. ✅
- The two internal `api.analyze_file` callers (query-by-name, element-extraction) pass neither param; no test passes them either. ✅

One small non-blocking note: `api.py` is a public surface and `analyze(**kwargs)` forwards arbitrary kwargs, so an external caller currently passing `include_complexity=True` switches from a silent no-op to `TypeError`. Since the param did nothing, failing loudly is arguably an improvement — worth a one-line CHANGELOG mention if the project tracks API back-compat.

aimasteracc:
**Correction to my earlier verdict — it is NOT safe as-is; CI is red.**

I was wrong that "no test passes them." My grep was line-based and missed a **two-line call**. CI (PR-fast, all axes + Quality Gate) fails with:

```
E   TypeError: analyze_file() got an unexpected keyword argument 'include_complexity'
FAILED tests/integration/core/test_api.py::TestAPIFacade::test_analyze_file_with_options
```

`tests/integration/core/test_api.py:120` does:
```python
result = api.analyze_file(
    sample_java_file, include_complexity=True, include_details=True
)
```

So removing the params is correct (they were inert) but the PR must **also update this test** — drop the two now-removed kwargs from that call (they did nothing, so the assertions are unaffected). With that one-line test edit the diff stays focused and goes green.

Net: the refactor is still the right call; it's just incomplete — one caller (a test) still passes the inert kwargs. (Good reminder that an executed test beats a static grep — this is exactly the kind of multi-line call a line grep misses.)

aimasteracc:
Accepting Codex's P1 — valid. `api.analyze_file` is a deliberately-stable public facade, and `include_complexity`/`include_details` are a test-encoded compat contract (`test_analyze_file_with_options`). Even though the params are currently inert (not forwarded), deleting them turns a silent no-op into a TypeError for any external caller — a user-visible break, not behavior-preserving dead-code removal. Per the project's semver / Format-Stability / RFC-for-public-API rules, this should not land as a casual cleanup. Closing. The inert-ness is better addressed by *wiring* the params (a future non-breaking correctness change), not by deletion. (Owner decision: garbage-removal campaign scoped to internal-only dead code.)

PR #1089, opened 2026-06-19
fix(complexity): count Java switch / ternary / do-while (stale node names)

aimasteracc:
Follow-up to #1085, found by the **cross-language complexity decision-node audit**.

## Bug
`tree-sitter-java` emits `switch_expression` (for both switch statements *and* switch expressions) and `ternary_expression` for `?:`, but `_JAVA_DECISION_NODES` listed `switch_statement` / `conditional_expression` — names the grammar **never produces**. So every `switch` and ternary was silently uncounted in cyclomatic complexity. `do_statement` (do-while) was also entirely absent from the set. Java therefore under-counted complexity, corrupting the heatmap / `project_health` grading / hotspot ranking for switch/ternary/do-while-heavy code.

The audit confirmed **Java was the only plugin with stale decision-node names** — C/C++/C#/Go/Kotlin/TS all match their grammars.

## Fix
`switch_statement` → `switch_expression`, `conditional_expression` → `ternary_expression`, and add `do_statement`. Each construct counts **once** (matching the Go/Rust/Swift convention — a 3-case switch is +1, not +3; verified by test).

| construct | pre-fix Cx | post-fix Cx |
|---|---|---|
| one `switch` | 1 | 2 |
| one `?:` ternary | 1 | 2 |
| one `do`-while | 1 | 2 |

`if` / `while` / `for` / `enhanced-for` / `catch` / `&&` / `||` (the #1085 work) are unchanged; `binary_search` control stays 4.

## Tests
RED-first `test_switch_counted_once` / `test_ternary_counted` / `test_do_while_counted` (each measured 1 pre-fix, exact-pinned to 2). Re-pinned the `java_bigservice` goldens (full+compact+csv): `performBackup` 1→2, `generateReport` 2→3, `manageInventory` 5→6 — each +1 from its switch. `Sample.java` / `java_userservice` have no switch/ternary/do-while and are unchanged. Full suite green (the only locally-flaky tests were 3 e2e latency/stderr budgets under CPU contention; they pass clean in isolation and on dedicated CI runners).

### DIFF
--- a/tests/golden_masters/compact/java_bigservice_compact.md
+++ b/tests/golden_masters/compact/java_bigservice_compact.md
@@ -29,11 +29,11 @@
 | cleanupCache | ():void | - | 313-326 | 3 | - |
 | updateCacheStatistics | ():void | - | 331-335 | 1 | - |
 | calculateCacheHitRatio | ():d | - | 340-343 | 1 | - |
-| performBackup | (S):void | + | 348-367 | 1 | - |
+| performBackup | (S):void | + | 348-367 | 2 | - |
 | performFullBackup | ():void | - | 372-395 | 6 | - |
 | performIncrementalBackup | ():void | - | 400-414 | 4 | - |
 | performDifferentialBackup | ():void | - | 419-433 | 4 | - |
-| generateReport | (S,M<S, O>):void | + | 438-471 | 2 | - |
+| generateReport | (S,M<S, O>):void | + | 438-471 | 3 | - |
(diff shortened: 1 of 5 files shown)


aimasteracc:
Closing as a duplicate of #1087 (created ~7 min earlier by a parallel session) — same fix (switch_expression / ternary_expression / do_statement), same goldens, same RED-first per-construct tests. Deferring to #1087. Verified #1087's approach is correct (cross-language audit confirmed Java was the only plugin with stale decision-node names).

PR #1092, opened 2026-06-19
fix(formatters): render module-level functions in TS/JS full table

aimasteracc:
## Problem

The TypeScript and JavaScript **full-table** formatters rendered only classes and their methods. Every **top-level / module-level function** (one not inside any class) was silently dropped:

- A flat module of functions → the full table showed nothing but the `# filename` header.
- A mixed file → the classes rendered, the module functions vanished.

For a code-intelligence tool whose job is to show structure to agents, this hides a core part of every TS/JS module. Found by dogfooding `--table full` on a module-level TS function (extracted, `method_count=1`, but absent from the table).

## Fix

Add a `## Global Functions` section (same `Cx` column as class methods) listing functions outside every class range — mirroring the C/Go full table and the existing TS **signatures**-table behaviour (which already had `_module_level_functions`).

## Tests

- TS `test_format_full_table` now asserts the `## Global Functions` section and a previously-dropped function (`mapArray`); exact length re-pinned 879 → 1159.
- New JS tests cover the flat-module and class+top-level cases.
- **No golden masters change** — the regression fixtures (`test_enum.ts` / `test_class.js`) have no top-level functions, so this was invisible to the golden suite (exactly why it went unnoticed).

@codex review

### DIFF
--- a/docs/CODEMAPS/formatters.md
+++ b/docs/CODEMAPS/formatters.md
@@ -92,6 +92,13 @@ TypeScript formatter signatures module:
   containers; overloads each appear as separate lines; used by
   `TypeScriptTableFormatter._format_signatures_table` via `structure action=signatures`)
 
+TS/JS full-table module-level functions:
+- `formatters/_typescript_formatter_full.py` and
+  `formatters/_javascript_formatter_full_mixin.py` render top-level (non-class)
+  functions in a `## Global Functions` section (same `Cx` column as class
+  methods). JS reads both `methods` (class methods) and `functions` (top-level)
+  since the JS plugin stores them in disjoint lists.
+
 ## CSV Control-Char Safety
 
 `formatters/_csv_safety.py` (`csv_safe_row` / `csv_safe_cell`) strips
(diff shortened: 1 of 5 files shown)


aimasteracc:
## Codex P2 triage — both real, both fixed (35ee2ce9)

|---|---|---|---|
| 1 | P2 JS drops `functions`-key globals | ✅ Real — `JavaScriptPlugin` stores top-level functions under `functions`, class methods under `methods` (disjoint); my section only read `methods`, so real plugin payloads dropped them (tests passed only because they used `methods`) | Collect `methods + functions` before filtering globals (JS lists are disjoint, so concatenate — not the TS `or`). New `test_plugin_shaped_functions_key_render` covers the `functions`-key shape. |
| 2 | P2 codemap not updated | ✅ Real — the codemap-sync hook only fires on *new* formatter files, not behavior changes | Added a `TS/JS full-table module-level functions` entry to `docs/CODEMAPS/formatters.md`. |

2374 formatter/golden tests green. Thanks — #1 was a genuine real-data gap my tests masked.

aimasteracc:
**Independent review (review arm) — verdict: both Codex P2s are already fixed on the current HEAD; good to merge.**

- *"Include JS `functions` when rendering globals"* — already addressed. Current HEAD collects both lists: `callables = methods + _list_or_empty(data.get("functions", []))`, and I confirmed the JS plugin returns top-level functions under `functions` (`extract_functions` → `Function`, `is_method=False`). So a flat `{classes:[], functions:[{name:'alpha'}]}` payload now renders `alpha`. Codex reviewed an earlier commit.
- *"Update the formatter codemap"* — already addressed; `docs/CODEMAPS/formatters.md` is in this PR's diff.

Both are stale-commit findings. No action needed.

PR #1093, opened 2026-06-19
feat(kotlin): show Cx (complexity) column in full table

aimasteracc:
## Problem

Kotlin already computes `complexity_score` (`calculate_kotlin_complexity`), but the full-table formatter never displayed it — the function table showed a `Suspend` column and no `Cx`. Kotlin was the **only** language whose full table omits the complexity metric (Go/Java/JS/TS/Rust/C#/C++/Python all show `Cx`).

Found by a cross-language complexity-display audit (the value was computed, e.g. `cx=10`, but invisible in `--table full`).

## Fix

Add a `Cx` column after `Lines` (keeping the Kotlin-specific `Suspend` column), mirroring the Rust full-table fix (#1082). Full table only — compact unchanged, matching #1082's scope.

## Tests

Re-pinned `kotlin_sample` full golden (Cx column added to existing rows; compact/csv/toon unchanged). New unit test asserts the header and a populated Cx value.

@codex review

### DIFF
--- a/tests/golden_masters/full/kotlin_sample_full.md
+++ b/tests/golden_masters/full/kotlin_sample_full.md
@@ -18,20 +18,20 @@ import java.util.Collections
 | LegacyUserManager | class | public | 57-60 | 0 | 1 |
 
 ## Functions
-| Function | Signature | Vis | Lines | Suspend | Doc |
-|----------|-----------|-----|-------|---------|-----|
-| User | fun(id: Long, username: String, email: String, active: Boolean) | pub | 8-13 | - | - |
-| display | fun(): String | pub | 16-16 | - | - |
-| summary | fun(): String | pub | 18-20 | - | - |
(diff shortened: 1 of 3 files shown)


aimasteracc:
**Independent review (review arm) — verdict: Codex P1 (codemap) is non-blocking here; consistent with precedent.**

The Cx-column addition makes Kotlin conform to the **already-documented** full-table shape — `docs/CODEMAPS/formatters.md` already specifies `| … | Cx | Doc |` generically (lines ~71–75), so this doesn't change the documented formatter surface. The direct precedent **#1082 (Rust Cx column, identical change class)** merged with the same three-file shape (formatter + test + golden) and **no codemap edit**, and the mechanical `tsa-codemap-sync` pre-commit hook passed on this commit. A one-line "Kotlin full table now shows Cx" note would be nice-to-have but isn't required for parity/correctness. Not a blocker.

Issue #1096, opened 2026-06-19
switch/match/case counted per-case in 5 languages (JS/TS/Ruby/Scala/Bash) — inconsistent with construct-once convention (#1090)


aimasteracc:
## Summary

The JavaScript and TypeScript extractors count `switch` **per case** instead of **once**, so they disagree with the construct-once convention every other language now uses (Go/Rust/Swift/Java/C/C++/C#). This is the same bug #1090 fixed for C/C++ — JS/TS were not touched.

## Evidence (on develop @ `22988616`, after #1085/#1087/#1090)

Identical control-flow structure (`if(&&)` + `else-if(||)` + `for` + `while` + `switch{3 cases}`):

| lang | Cx | expected (construct-once) |
|---|---|---|
| Java / C / C++ / C# | **8** | 8 ✅ |
| JS / TS | **11** | 8 ❌ (+3 = the three `case`s) |

Isolating just a switch (no other branches), construct-once should give Cx=2 regardless of case count:

| switch body | JS Cx | should be |
|---|---|---|
| 1 case  | 3 | 2 |
| 3 cases | 5 | 2 |
| 5 cases | 7 | 2 |

So JS/TS count `switch(1) + each case(n)`; Java returns 2 for all of the above.

## Impact

JS/TS heatmap / hotspot ranking / `project_health` grade systematically over-weight switch-heavy functions, and cross-language comparisons are apples-to-oranges (a 5-case switch reads as 5× the branches it is under the project's own convention).

## Fix

Mirror #1090: in the JS/TS extractor decision-node handling, count the switch construct once (drop `switch_case` / per-case nodes from the decision set, keep `switch_statement`). RED-first: pin `switch{N cases}` → Cx 2 for N∈{1,3,5}; re-pin any JS/TS goldens whose switch methods drop.

_Found by the independent review arm via a holistic cross-language consistency probe after the #1080–#1090 complexity sprint. Complements #1094 (multi-path inconsistency); this one is a cross-language gap within the extractor path itself._

aimasteracc:
**Update — the per-case `switch` over-count is broader than JS/TS: it affects 5 languages, and 2 of them have tests pinning the wrong behavior.**

Ran the same isolated probe (a `switch`/`match`/`when`/`case` with **3 arms, no other branches**; construct-once ⇒ Cx **2** for every language) across all remaining plugins on develop @ `22988616`:

| language | Cx (3 arms) | convention |
|---|---|---|
[...]

Issue #1094, opened 2026-06-19
Cyclomatic complexity has 4 inconsistent implementations (extractor vs health_scorer vs heatmap) — switch/ternary counts disagree


aimasteracc:
## Summary

Cyclomatic complexity is computed by **at least four independent implementations** that disagree, so the same Java method gets different Cx in the formatter/golden path vs. the heatmap/health path. Surfaced by Codex as a P1 on #1087 (the Java `switch`/`ternary`/`do-while` fix) and independently confirmed empirically.

## The duplicated paths

1. **Extractor** — `languages/_java_ast_helpers.py` (+ `c_helpers.py` / `_cpp_complexity_helpers.py` / `csharp_helpers.py`). Fixed by #1085 (`&&`/`||`) and #1087 (`switch_expression`/`ternary_expression`/`do_statement`, construct-once).
2. **`health_scorer.py`** — its own `DECISION_NODE_TYPES: dict[str, set[str]]`; the `"java"` set still lists the dead `switch_statement` / `conditional_expression` names → `project_health` grading silently undercounts every Java switch/ternary, exactly like the extractor did pre-#1087.
3. **`complexity_heatmap.py` / `_ast_extraction.py`** — count `switch_block_statement_group` **per case** (not construct-once), so `analyze_file_complexity(file, 'java')` returns Cx≈5 for a 4-case switch where the extractor now returns 2.
4. **`complexity_heatmap_tool.py`** — consumes #3.

## Concrete symptom

After #1085+#1087, `examples/BigService.java::performBackup` is **Cx=2** in the `--table full` / golden output (extractor) but the heatmap/health path reports a different value, so the user-facing **heatmap / hotspot ranking / health grade** — the entire motivation for #1080–#1087 — remain inconsistent with the construct-once convention the extractor now uses.

## Proposed direction (needs an owner; likely RFC-worthy since it moves health grades)

- Establish a **single source of truth** for decision-node counting that every path (extractor, health_scorer, heatmap, ast_extraction) routes through, OR
- At minimum, re-pin the `health_scorer` / `heatmap` / `ast_extraction` Java sets to `switch_expression` / `ternary_expression` / `do_statement` with the **construct-once** convention, and add a cross-path invariant test: `extractor_cx(method) == heatmap_cx(method) == health_cx(method)` for a fixture with one switch + one ternary + one do-while.

[...]

PR #1098, opened 2026-06-19
fix(complexity): Ruby/Scala/Bash count switch/match/case once, not per arm

aimasteracc:
Completes the construct-once cyclomatic-complexity convention across **all** languages — closes the per-arm gap tracked in #1096.

## The gap
On develop a 3-arm switch/match/case reported, in the three languages the audit flagged:

| language | construct | reported Cx | correct |
|---|---|---|---|
| Ruby | `case` / `when` | **4** (each `when`) | 2 |
| Scala | `match` | **5** (each `case_clause`) | 2 |
| Scala | `catch` | **3** (the `case_clause` inside it) | 2 |
| Bash | `case` | **4** (each `case_item`) | 2 |

…while **Java / C / C++ / C# / Go / Rust / Swift / Kotlin / PHP / JS / TS** all count the construct once (#1080/#1085/#1087/#1090/#1097). So heatmap / hotspot ranking / `project_health` over-weighted switch-heavy Ruby/Scala/Bash code.

## Fix
- **Ruby** — drop `when` from `_RUBY_DECISION_STATEMENT_TYPES` (keep `case`).
- **Scala** — drop `case_clause` from `_SCALA_DECISION_TYPES` (keep `match_expression` + `catch_clause`); this also fixes the `catch` over-count.
- **Bash** — drop `case_item`, add `case_statement` to `_BASH_DECISION_TYPES`.

After: a 3-arm switch/match/case is **Cx 2** in all three.

## Tests
RED-first `test_case_counts_once_not_per_when` (Ruby), `test_match_counts_once_not_per_case` + `test_catch_counts_once_not_per_case` (Scala), `test_case_counts_once_not_per_item` (Bash). Re-pinned the Scala (12→8) and Bash (12→10) RICH fixtures — they had **pinned the per-arm counts**, exactly the kind of test that enshrines a bug. No goldens change (the ruby/scala/bash sample files have no multi-arm case constructs). Found by the cross-language complexity audit.

With this, **all 13 languages share the construct-once convention** for the extractor path. The remaining complexity work is the cross-*path* inconsistency (extractor vs health_scorer vs heatmap) tracked in #1094.

### DIFF
--- a/tests/unit/languages/test_cyclomatic_complexity.py
+++ b/tests/unit/languages/test_cyclomatic_complexity.py
@@ -108,6 +108,23 @@ class TestRubyCyclomaticComplexity:
         assert funcs[0].name == "process"
         assert funcs[0].complexity_score == 8
 
+    def test_case_counts_once_not_per_when(self):
+        """A valued `case x` with N `when` arms is one decision → complexity 2."""
+        funcs = _ruby_functions(
+            "def f(x)\n case x\n when 1 then 0\n when 2 then 0\n"
+            " when 3 then 0\n else 0\n end\nend"
+        )
+        assert funcs[0].name == "f"
+        assert funcs[0].complexity_score == 2
+
+    def test_conditionless_case_counts_each_when(self):
+        """A subjectless `case` is an if/elsif chain — each `when` is a decision."""
+        funcs = _ruby_functions(
+            "def f(x)\n case\n when x > 0 then 1\n when x < 0 then -1\n end\nend"
+        )
+        assert funcs[0].name == "f"
+        assert funcs[0].complexity_score == 3
+
 
 # ---------------------------------------------------------------------------
 # Kotlin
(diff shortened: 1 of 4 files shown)


aimasteracc:
All three Codex P2s were real (the first cut dropped the per-arm node globally, which over-corrected) — fixed in the latest push:

1. **Standalone partial-function `case_block`** — now counted once when its parent is not `match_expression`/`catch_clause` (e.g. `xs.map { case n => ... }` → Cx 2 instead of 1).
2. **`case` guards** — `guard` is now in the Scala decision set, so `case n if n > 0` adds a branch (`x match { case n if n>0 ...; case _ ... }` → Cx 3).
3. **Conditionless Ruby `case`** — a subjectless `case` is detected via `child_by_field_name("value")`; its `when` arms are counted individually (if/elsif semantics → Cx 3), while a valued `case x` stays construct-once (Cx 2).

Added RED-first tests for all three. Thanks — these were genuine regressions my blunt first version introduced.

PR #1099, opened 2026-06-19
RFC-0019: cyclomatic complexity single source of truth (#1094)

aimasteracc:
Draft RFC for #1094 — the last piece of the cross-language complexity moat.

## Why
#1080–#1098 fixed cyclomatic complexity **eight times** (once per language) but only in the *extractor* path. Two other implementations were never touched and still disagree on the same function:
- `health_scorer.DECISION_NODE_TYPES` — stale Java/JS/TS node names (`switch_statement`/`conditional_expression` instead of `switch_expression`/`ternary_expression`, no `do_statement`) → `project_health` undercounts switch/ternary/do-while.
- `_ast_extraction._count_complexity_in_node` — counts switch arms per-case → `analyze_file_complexity` reports ~5 for a 4-case switch where the golden `--table` reports 2.

So heatmap / hotspot ranking / health grade disagree with `--table` and with each other, and the same bug class has to be fixed three times forever.

## What the RFC proposes
A single `cyclomatic_complexity(node, language)` dispatcher over the **extractor** per-language walkers (the most-tested, already-correct source of truth post #1080–#1098); delete the two duplicate decision tables and point `health_scorer` / `_ast_extraction` at the dispatcher. No algorithm change — pure de-duplication. Keystone test: a cross-path invariant `extractor_cx == health_cx == heatmap_cx` for the same fixture.

Drawback acknowledged: health grades / heatmap numbers move (down) for affected code and need conscious exact re-pins — but they become *correct* and *consistent*.

Reviewers: please sanity-check the dispatcher-vs-`element.complexity_score` choice (Alternatives A) and the grade-movement drawback before I implement.

### DIFF
--- a/rfcs/0019-complexity-single-source-of-truth.md
+++ b/rfcs/0019-complexity-single-source-of-truth.md
@@ -38,6 +38,16 @@ never touched and still disagree:
    arm**, so `analyze_file_complexity(...)` reports Cx ≈ 5 for a 4-case switch
    where the extractor (and the golden `--table full`) now report 2.
 
+The heatmap's `_COMPLEXITY_NODES` table (the third definition) has drifted the
+furthest from the extractor convention — prototyping found it: (a) counts switch
+**per arm** (`switch_case`/`switch_block_statement_group`/`match_arm`/
+`expression_switch_case`/`type_switch_case`/`select_case`); (b) counts
+`else_clause` as a decision (it is **not** one — no other path counts `else`);
+(c) uses stale/divergent operator nodes (`logical_expression` /
+`conditional_expression` for JS/TS where the grammar emits `binary_expression` /
+`ternary_expression`). So the heatmap is wrong on multiple axes, not just
+per-case — aligning it is the bulk of the re-pinning.
+
 Concrete: `examples/BigService.java::performBackup` is `Cx 2` in the golden
 formatter output but the heatmap/health path scores it differently, so the
 user-facing heatmap / hotspot ranking / health grade are inconsistent with the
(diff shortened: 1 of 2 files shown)


aimasteracc:
Excellent review — all four are real; this is exactly the dead-end the RFC process is for. Verdicts:

1. **Preserve heatmap `decision_points` / cache** — ✅ agreed and already the crux of the "decision_points must survive" section: the source of truth is the **decision-node TYPE definitions** (count *and* breakdown derive from one place), not a scalar dispatcher. Will make the acceptance criteria explicit about `decision_points` + the v1 cache rows staying intact.

2. **Fix Python before making the extractor canonical** — ✅ **the most important finding.** Confirmed empirically: `PythonElementExtractor._calculate_complexity_optimized` is a `re.findall(r"\bkeyword\b", text)` counter — so a function whose *comment/docstring/string* contains `if`/`for`/`and`/… is over-counted (a keyword-only comment → Cx 6), and `case` is counted per-arm. Same class as the JS/TS bug #1097 just fixed. **Python must get the AST rewrite first** — I'll do it as a standalone PR (it's a real bug regardless of this RFC), then Python is eligible as a source of truth.

3. **Bash/Scala in the invariant sweep** — ✅ correct, they produce `complexity_score`; the sweep is 15 (13 fully-wired + Bash/Scala), not 13. Will fix the acceptance criterion.

4. **`FUNCTION_NODE_TYPES` must cover every health language** — ✅ real migration hazard; switching `score_complexity` to per-function summing drops any language missing from `FUNCTION_NODE_TYPES` (PHP/Kotlin/…). Will add "FUNCTION_NODE_TYPES covers every language with a health table" to acceptance criteria, with a guard test.

Updating the RFC with all four; starting on #2 (Python AST complexity) now since it stands alone.

aimasteracc:
Prerequisite ✅ done: **#1100** (merged) makes the Python extractor an AST walk instead of a keyword-regex counter — confirmed it was counting keywords in comments/docstrings/strings, `match` per-arm, and `with`. So Python is now eligible as a canonical complexity source. Acceptance criteria updated to fold in all four review findings.

PR #1101, opened 2026-06-19
fix(complexity): heatmap (live + cache) derives CC from the extractor (#1094)

aimasteracc:
First implementation slice of RFC-0019 — closes the **heatmap** half of #1094 (the health half is a parallel PR).

## The bug
The `viz` / complexity-heatmap / hotspot tools disagreed with the extractor (and `--table` / golden masters) on the same function:

| | Java/JS 3-case switch |
|---|---|
| extractor (`--table`, golden) | **2** |
| heatmap (before) | **5** (per-arm + stale operator nodes + counts `else`) |

And the heatmap had **two paths that would drift apart**: the live walk and the cache-backed path (which re-derived the count from the per-arm `decision_points` sum).

## The fix — route both paths through the extractor `complexity_score` (the single source of truth)
- **Live** (`_extract_functions_from_ast`): look each function up from the plugin extractor by start line.
- **Index time** (`_ast_extraction._extract_symbols`): stamp each function symbol with the extractor’s canonical `complexity`.
- **Cache read** (`analyze_file_complexity_from_cache`): use the stored `complexity`; older rows without it re-parse via the live path (never the per-arm sum).
- Bump `_AST_CACHE_EXTRACTOR_VERSION` **13 → 14** so existing rows re-index.

`decision_points` (the per-type breakdown the viz tool shows) is untouched.

## Verified
Empirically, extractor == live == cache = **2** for the Java switch (was 5). The enforced invariant `test_complexity_cross_path_invariant.py` is **un-xfailed** and parametrized over **Java / JS / TS / Go / Rust** (extractor == live == cache for a multi-arm switch/match). Full suite green (21370 passed; only the usual 3 e2e latency/stderr budgets flaked under local CPU contention — they pass in isolation). **No `test_complexity_heatmap` / golden re-pins** — those fixtures have no multi-arm switches, so only switch-heavy code moves (correctly).

Remaining for #1094: `health_scorer` (parallel PR), then the RFC’s full table de-duplication is moot since both consumers now use the extractor value.

### DIFF
--- a/tests/unit/test_complexity_cross_path_invariant.py
+++ b/tests/unit/test_complexity_cross_path_invariant.py
@@ -1,43 +1,52 @@
 """Cross-path cyclomatic-complexity invariant (RFC-0019 / #1094).
 
-The extractor (``element.complexity_score``, used by ``--table`` and the golden
-masters) and the heatmap path (``complexity_heatmap.analyze_file_complexity``,
-used by the ``viz``/heatmap MCP tools and ``project_health``) must agree on the
-cyclomatic complexity of the same function.
+Every consumer must agree on a function's cyclomatic complexity:
 
-Today they do NOT: the heatmap path counts each ``switch`` arm separately while
-the extractor counts the construct once. These tests pin that invariant as a
-*strict xfail* — they document the known inconsistency as an executable,
-falsifiable contract. When RFC-0019 routes both paths through one source of
-truth, ``strict=True`` makes them fail-as-xpass, forcing the xfail marker to be
-removed in the same change.
+- the **extractor** (``element.complexity_score``, used by ``--table`` and the
+  golden masters — the single source of truth),
+- the **live heatmap** (``complexity_heatmap.analyze_file_complexity``, used by
+  the ``viz``/heatmap MCP tools and ``project_health``),
+- the **cache-backed heatmap** (``analyze_file_complexity_from_cache``).
+
+This was the #1094 inconsistency: the heatmap counted each ``switch`` arm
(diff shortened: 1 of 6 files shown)


aimasteracc:
## Adversarial review triage (2 independent reviewers)

Two adversarial review agents (parity/cache + regression) ran against this PR. Both converged on **one** real bug; everything else verified clean.

|---|---|---|
| `_AST_CACHE_EXTRACTOR_VERSION` bumped in the **dead** copy (`_ast_cache_indexer.py`) while the **live** one (`ast_cache.py`) stayed at 13 | **REAL** — also caught by `test_extractor_version_*` (the 5 red checks) | Fixed: both sites now `= 14`, test re-pinned to 14 and renamed `_matches_in_both_sites` (commit 9284403) |
| `decision_points` breakdown preserved on both live + cache paths | OK | none — `test_decision_points_populated` + MCP surface verified |
| Parity beyond switch (ternary / &&-|| / match / try-except across py/js/ts/go/rust/java) | OK | none — extractor == live == cache for every reported fn |
| v13-row fallback to live parse | OK (correct) | resolved by the version bump above (v13 rows now re-index) |
| Index-time cost: `_annotate_canonical_complexity` adds ~2.85× to **extraction** (8.9→25.2 ms on BigService.java, ~+8 ms/file) | Real cost, not a correctness bug | **Accepted** — index-time is the right (amortized) place vs per-query re-parse; tracked as a follow-up optimization (fuse the canonical-CC pass into the existing symbol walk) |

Net: the version-bump bug is fixed; the remaining finding is a documented, accepted perf trade-off.

PR #1104, opened 2026-06-19
feat(dead-code): add path scoping to health action=dead (#1084)

### LINKED ISSUE #1084: health.dead: path-scoping params (file_path/directory/query) silently ignored — guidance says 'filter by path' but none exists; 1341 'dead' is corpus/benchmark noise
**Lens:** perpetual dogfood/QA arm — issue-only, no code changes. Sibling to #1083 (which audited *product code* over-engineering); this one audits the **MCP health facade's own affordances** through the same [Ponytail](https://github.com/DietrichGebert/ponytail) "dead flexibility / broken affordance" lens. Verified against `develop` @ `0ff2f4b5`.

### What I did (the real use case)
Ponytail-style sweep: *"show me the deletable dead functions in the product package `tree_sitter_analyzer/`."* The tool's `next_step` explicitly says: `Raise max_dead / ... or **filter by path** to see more.` So I tried to scope it.

### What happens — three different param names, byte-identical output
| Call | Expected | Actual |
| --- | --- | --- |
| `health action=dead file_path=tree_sitter_analyzer/mcp` | dead code under `mcp/` | identical 1341, identical corpus/benchmark list |
| `health action=dead directory=tree_sitter_analyzer/mcp` | dead code under `mcp/` | identical 1341 — **returns `corpus/…` and `benchmarks/…` hits that are NOT under `tree_sitter_analyzer/mcp`** |
| `health action=dead query=tree_sitter_analyzer/mcp` | dead code matching path | identical 1341 |

The `directory=tree_sitter_analyzer/mcp` run returning `corpus/python/...` and `benchmarks/...` rows (both outside that directory) is hard proof the param is inert. The `action=dead` schema confirms it: documented params are only `mode, include_test_files, max_dead, max_imports, max_variables` — **there is no path/directory/file_path filter at all.** So the `next_step` advertises a knob that does not exist.

### Why it matters (the number is noise)
The headline `total_dead_functions_transitive: 1341` is dominated by code that is **dead by design, not deletable**:
- `corpus/python/**` — standalone parser-fixture samples that exist *precisely* to be parsed and never called (`foo`, `async_foo`, comprehension vars…). Flagging them as "dead" is a false-positive category.
- `benchmarks/**` — `cmd_run`, `cmd_status`, `_run_subprocess`… reachable only through argparse/CLI entrypoints the static call-graph can't see (`unreachable_from_entry`).
- root one-off scripts — `analyze_coverage_json.py`, `analyze_specific_sql.py`, `fix_indentation.py`.

All 50 rows in the default window come from these three buckets; **zero** are from the actual product package. An agent doing the intended "find deletable dead product code" task cannot get there — no scoping works, and the real signal is buried under fixtures.

### Inconsistency within the same facade
`health action=heatmap` **does** honor `directory` (via the default `mode=project`): `heatmap directory=tree_sitter_analyzer/core max_files=10` correctly returns a directory-scoped ranking. So path-scoping exists on one action of the facade and is silently dropped on another. Pick one contract.

### Suggested resolution (maintainer's call — issue-only)
1. Either wire a real `path`/`directory` filter into `action=dead` (parity with `heatmap`), **or** delete the "filter by path" phrase from `next_step` so guidance stops pointing at a nonexistent knob (Ponytail: don't advertise dead flexibility).
2. Default-exclude `corpus/`, `benchmarks/`, and repo-root scripts from dead-code candidates (they're not product code) so the headline count means something.

---
### Minor related note (doc-only, same facade)
`health action=heatmap`'s facade description lists `Params: mode, file_path, function_name, language, directory, max_files` but never enumerates the **mode values** (`file` / `function` / `project`). It cost 3 failed calls (`mode=file` and `mode=function` both hard-require `file_path`) before discovering that `project` is the project/directory-wide mode and is the default when `mode` is omitted. Listing the three mode values in the description would remove that.

— filed by the perpetual dogfood/QA arm (issue-only, no code changes).

aimasteracc:
Closes #1084.

## Problem
`health action=dead` (CLI `--dead-code`) advertised `filter by path` in its `next_step`, but **no path parameter existed** — `file_path`/`directory`/`query` were all silently ignored. Meanwhile `health action=heatmap` on the *same facade* honors `directory`, so path-scoping existed on one action and was silently dropped on another. The headline `total_dead_functions_transitive: 1341` was dominated by `corpus/`, `benchmarks/`, and root one-off scripts (dead **by design**, not deletable); zero of the default 50 rows came from the product package, and no scoping could get an agent there.

## Fix
Add a `path` parameter (MCP) + `--dead-code-path` flag (CLI parity) that scopes the reported dead functions / unused imports / unreferenced variables to definitions under a **project-relative, segment-matched** path prefix.

- **Transitive reachability is still computed across the whole project** — only the *reported* items are scoped. Filtering candidates *before* analysis would break cross-file dead-chain detection (a function is dead only because its caller elsewhere is dead), so the filter runs on results, before any count, so every total reflects the scoped set.
- **Segment-aware**: `tree_sitter_analyzer/m` does **not** match `tree_sitter_analyzer/mcp/x.py`; trailing slash normalized; backslashes normalized.
- `next_step`'s `filter by path` guidance is now truthful (chose to wire the filter rather than delete the phrase — gives agents the affordance heatmap already has).

Verified end-to-end: `--dead-code-path tree_sitter_analyzer/mcp` → 238 scoped product candidates vs 1284 unscoped.

## Tests
7 path-scoping cases (exact assertions per the LOCKED exact-assertion rule): scope-to-package, exclude corpus/benchmarks, single-subtree, trailing-slash normalization, **segment-boundary** (partial segment must not match), no-match→`INFO`. Plus the existing 19 counts-consistency tests unchanged.

Regenerated `docs/api/facade-actions.md`; bumped README en/ja/zh CLI-flag count 295→296.

### Out of scope (follow-up)
The issue also suggested default-excluding `corpus/`/`benchmarks/`/root-scripts from candidates. That's a separate concern (changes the default candidate set, somewhat opinionated) and is left for a focused follow-up — the `path` filter already lets agents scope to product code.


aimasteracc:
@codex review

aimasteracc:
## Codex review triage

|---|---|---|
| **P2** — Normalize root-relative path aliases (`.` / `./pkg` matched nothing because result paths are project-relative) | **REAL — valid edge case** | Fixed in e3412f9f: `_under_path` strips a leading `./` and treats `.`/empty as the whole-project (root) scope. Added 2 tests (`path='.'`→all 5; `path='./tree_sitter_analyzer/mcp'`→2). Segment-boundary strictness preserved (`.../m` still doesn't match `.../mcp/`). |

Good catch — `--dead-code-path .` is the natural way to say "whole project" and `./pkg` is a common copy-paste, both of which my first cut silently rejected. dev≠review caught it before merge.

The earlier E2E black-box red was the known `test_check_project_health_under_10s` latency flake (MCP server >55s under runner contention; `test_tools_list` passed in the same job — unrelated to this diff); the new commit re-runs CI fresh. codecov/patch is advisory.

PR #1105, opened 2026-06-19
test(e2e): warm index before measuring project_health latency (kill order-dependent flake)

aimasteracc:
## Problem
`TestToolLatencyBudgets::test_check_project_health_under_10s` is an **order-dependent flake**. The on-disk AST cache persists across tests within a job, so whether `check_project_health` ran warm depended on test ordering (pytest-randomly). When the test ran first/cold it had to analyze the whole ~1.8k-file repo (~50–90s on a loaded CI runner), blowing the `10s × _CI_FACTOR` (=50s) budget and hitting the 55s client timeout — `TimeoutError: MCP server did not send a complete JSON line within 55s`.

This is **independent of any diff**: it failed #1104's CI 3 times in a row while the *identical* code passed locally with a warm cache (8.6s), and disabling #1101's complexity annotation left the cold cost unchanged (~86s vs ~88s) — i.e. the cold cost is the inherent index/analysis of the repo, not a regression. It's been on the known-flake list for a while; this removes the root cause.

## Fix
`project_health` must analyze every file to score health, so unlike `codegraph_metrics` it cannot return a cheap "run index first" hint — cold-cache responsiveness is already covered by `test_codegraph_metrics_cold_cache_under_5s` (fresh tmpdir). This test's intent is **steady-state** latency on an indexed repo, so warm the AST cache once (unmeasured, 300s timeout) and measure the realistic indexed path. Now order-independent.

## Verification
- From a **cleared cache** (the exact flake scenario): was a hard 55s timeout → now **passes** (88s unmeasured warm-up + fast measured call).
- Full `TestToolLatencyBudgets` class: 3 passed.

This unblocks the merge pipeline (every PR risked this flake on an unlucky/cold runner — it's what's currently blocking #1104, whose diff is verified innocent).

### DIFF
--- a/tests/e2e/test_mcp_smoke.py
+++ b/tests/e2e/test_mcp_smoke.py
@@ -255,6 +255,11 @@ class TestToolLatencyBudgets:
             f"codegraph_metrics returned a JSON-RPC error: {response.get('error')}"
         )
 
+    # The unmeasured cold warm-up below can take well over a minute on a loaded
+    # runner; the CI e2e job runs pytest with a global ``--timeout=60``, so this
+    # per-test override keeps pytest-timeout from killing the warm-up before the
+    # measured (warm) call runs. The 50s budget assertion still guards latency.
+    @pytest.mark.timeout(400)
     def test_check_project_health_under_10s(self, mcp_server: MCPClient) -> None:
         """check_project_health on the TSA repo itself must complete in 10s (×3 in CI).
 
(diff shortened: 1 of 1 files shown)


aimasteracc:
@codex review

aimasteracc:
## Codex review triage

|---|---|---|
| **P1** — `--timeout=60` (ci.yml:140) kills the cold warm-up before the measured call, leaving the flake unfixed on slow runners | **REAL — and important** | Fixed in 47468604: `@pytest.mark.timeout(400)` overrides the global per-test timeout for this one test so the warm-up completes. The 50s budget assertion on the *measured* call is unchanged. |

Sharp catch — without this, my first commit only passed because that runner warmed the cache in <60s; a genuinely cold/slow runner (the exact scenario this PR targets) would still have been killed at 60s. **Verified**: with `--timeout=60` + a cleared cache (the simulated CI failure condition), the test now passes — 85s unmeasured warm-up, no kill. dev≠review caught the incompleteness before merge.

PR #1157, opened 2026-07-21
fix(call-graph): stop hardcoding a 7-language file-scan allowlist

aimasteracc:
## Summary
- `CallGraph.build()` hardcoded a `supported_exts` allowlist to 11 extensions covering only python/js/ts/java/go/c/cpp, completely independent of `function_extraction.py`'s per-language dispatch tables (`_FUNC_DEF_TYPES`/`_CALL_NODE_TYPES`), which already implement rust/csharp/kotlin/ruby/php/swift/lua.
- Because `CallGraph` is the shared backbone for `call_graph_tool`, `dead_code_analyzer`, and every `codegraph_*` MCP tool (context/explore/impact/navigate/overview/pr_review/relation/visualize/metrics), these 7 languages silently returned empty results across the entire CodeGraph-parity surface even though their extraction logic already worked.
- `SUPPORTED_EXTS` is now derived from `EXT_TO_LANG` (the project's documented single source of truth for ext→language, see `languages/lang_extension_map.py`) filtered by a new `function_extraction.SUPPORTED_LANGUAGES` constant, so the two lists can't drift apart again.
- Added a contract test (`test_call_graph_supported_exts_covers_function_extraction_languages`) plus per-language integration fixtures for the 7 previously-excluded languages.

## Verification
- New parity tests (rust/csharp/kotlin/ruby/php/swift/lua) pass; full `tests/unit -k "call_graph or dead_code or codegraph"` suite: 1202 passed, 16 pre-existing skips, 0 new failures.
- Empirically verified against `tests/golden/`: `--call-graph all_functions` went from 277 → 506 detected functions, now including rust/kotlin/ruby/php/swift.
- `mypy`/`ruff` clean on touched files.
- One pre-existing, unrelated failure confirmed on base `develop` too: `test_dead_code_analyzer.py::TestDeadCodeErrorHandling::test_symlink_handled` (Windows symlink privilege, environment-only).

## Out of scope (intentional)
- **Scala**: `Parser.parse_file` currently returns no tree for `.scala` in this environment — a separate root-cause investigation, not this allowlist bug.
- **Bash**: has no `_FUNC_DEF_TYPES`/`_CALL_NODE_TYPES` entries at all — a feature gap, not a regression of this bug.
- Lua's cross-file import resolution (`import_extractors`) still has no lua branch — same-file call resolution now works (verified `callers_of`/`callees_of`), cross-file import-based resolution is a follow-up.

## Test plan
- [x] `pytest tests/unit/test_call_graph_integration.py -v`
- [x] `pytest tests/contracts/ -q`
- [x] Full call-graph/codegraph/dead-code regression sweep (1202 passed)
- [x] `mypy` / `ruff check` clean
- [x] Empirical CLI check against golden corpus

### DIFF
--- a/docs/CODEMAPS/languages.md
+++ b/docs/CODEMAPS/languages.md
@@ -10,7 +10,7 @@ Not every registered plugin is wired into the indexer to the same depth:
 
 - **13 fully wired** (full symbol + call graph): Python, Java, JavaScript, TypeScript, Go, Rust, C, C++, C#, Swift, Kotlin, Ruby, PHP
 - **2 symbol-indexed** (call-graph wiring pending): Bash, Scala — both graduated in v1.22.0
-- **1 partial symbol-indexed**: Lua function symbols; resolved calls and indexed imports pending
+- **1 partial symbol-indexed**: Lua function symbols and same-file resolved calls; cross-file import-based resolution pending
 - **5 data/markup** (reachable via the single-file CLI path): HTML, CSS, Markdown, SQL, YAML
 - **1 scaffold** (plugin exists, indexer wiring pending): JSON
 
--- a/tests/contracts/test_internal_architecture_contract.py
+++ b/tests/contracts/test_internal_architecture_contract.py
@@ -59,6 +59,40 @@ def test_ast_cache_call_edge_extraction_does_not_depend_on_call_graph() -> None:
     assert "tree_sitter_analyzer.call_graph" not in imports
 
 
+def test_call_graph_supported_exts_covers_function_extraction_languages() -> None:
+    """CallGraph's file-scan allowlist must not silently narrow past what
+    function_extraction.py already knows how to walk.
+
+    Regression: CallGraph.build() historically hardcoded its own 7-language
+    supported_exts set, independent of function_extraction's per-language
+    dispatch tables. Rust/C#/Kotlin/Ruby/PHP/Swift/Lua all had working
+    _FUNC_DEF_TYPES/_CALL_NODE_TYPES entries but were silently excluded from
+    every CodeGraph-parity tool (--call-graph, --callers, --callees,
+    dead-code detection, codegraph_* MCP tools) because the file scanner
+    dropped their files before parsing. See call_graph.py's SUPPORTED_EXTS
+    docstring.
+    """
+    from tree_sitter_analyzer.call_graph import SUPPORTED_EXTS
+    from tree_sitter_analyzer.function_extraction import SUPPORTED_LANGUAGES
+    from tree_sitter_analyzer.languages.lang_extension_map import EXT_TO_LANG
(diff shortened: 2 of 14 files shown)


aimasteracc:
Triaged Codex's 7 findings (all real, none false-positive). Fixed 4 in 6c13cc91, documented 3 as follow-ups.

| # | Finding | Priority | Verdict | Action |
|---|---|---|---|---|
| 1 | `_KNOWN_ENTRY_PATTERNS` has no entries for the 7 new languages → dead-code scan flags everything (incl. `main`) as dead | P1 | **Real, fixed** | Added `rust`/`csharp`/`kotlin`/`swift` patterns (unambiguous single-entry conventions). `ruby`/`php`/`lua` deliberately left out — no single canonical entry-function convention to guess at; tracked as follow-up. |
| 2 | Cross-file import scope not preserved (`_collect_import_map`/`_resolve_import_path` only handle python/js) | P2 | **Real, deferred** | Genuine gap, but fixing it means extending every new extractor to populate `names` plus widening `_resolve_import_path`'s suffix probing — a separately-testable change in its own right. Filing as follow-up rather than bundling into this fix. |
| 3 | `_EXCLUDE_DIRS` omits `target`/`obj`/`vendor`/`Pods`/`DerivedData` → scans vendored/build trees | P1 | **Real, fixed** | Replaced the 4th independently-maintained copy with `constants.EXCLUDE_DIRS` (canonical set) — same fix philosophy as the PR itself. |
| 4 | `_extract_recursive()` only populates `class`/receiver for python/java/go | P1 | **Real, deferred** | Confirmed: even the *original* 7 languages (js/ts/c/cpp) already lacked this — it's a pre-existing gap this PR widens the reach of, not a new regression. Worth fixing but scoped bigger than this PR; tracked as follow-up. |
| 5 | New regression tests lack an issue/incident reference (T-5) | P1 | **Real, fixed** | Added "PR #1157, 2026-07-21 dogfood pass" reference to both. |
| 6 | New Swift/Lua tests run unconditionally; neither grammar is in default deps | P1 | **Real, fixed** | Added the same `skipif` guard pattern `test_swift_plugin.py`/`test_lua_plugin.py` already use. |
| 7 | `SUPPORTED_EXTS` (derived from `EXT_TO_LANG`) misses `.kts`/`.pyw`/`.pyi`/`.mjs`/`.cjs`/`.mts`/`.cts` | P2 | **Real, deferred** | Confirmed these extensions are absent from `EXT_TO_LANG` itself (not just filtered out here) — a pre-existing gap in the canonical map, out of scope for a "make CallGraph match function_extraction" fix. Follow-up. |

Re-verified after fixes: 1206 passed / 16 pre-existing skips across call_graph/dead_code/codegraph/contracts test suites, mypy/ruff clean.

PR #1158, opened 2026-07-21
fix(plugins): fold drifted consistency script into pytest contract; fix scala delegation

aimasteracc:
## Summary
- `scripts/check_plugin_consistency.py` was an orphaned script (confirmed via grep: not referenced by any CI workflow, pre-commit hook, or doc) that duplicated `tests/contracts/test_plugin_architecture_contract.py` logic and had drifted out of sync with it — its `GRANDFATHERED_SINGLE_FILE` set was missing `bash_plugin.py`/`json_plugin.py`/`scala_plugin.py`, so running it standalone reported false failures.
- It also carried one check the pytest contract never enforced: "analyze_file must not call self.extract_elements()". That gap let `ScalaPlugin.analyze_file` call `self.extract_elements()` directly (bypassing the `create_extractor()` delegation pattern every other plugin follows) without CI ever catching it.
- Folded the missing check into the contract test as `test_analyze_file_does_not_call_self_extract_elements`, so there's one source of truth, and deleted the now-fully-redundant standalone script.
- Fixed the one real violation it caught: extracted the shared element-dict-building logic into a module-level `_build_scala_elements_dict()` helper used by both `ScalaPlugin.analyze_file` and the public `extract_elements()` method, so neither calls the other.

## Verification
- `tests/contracts/test_plugin_architecture_contract.py`: 7 passed (including the new check).
- Full scala-keyword test sweep (`pytest tests/ -k scala`): 260 passed, 1 pre-existing xfail, 0 behavior change.
- `tests/contracts/` full suite: 39 passed.
- `mypy`/`ruff` clean on touched files.

## Test plan
- [x] `pytest tests/contracts/test_plugin_architecture_contract.py -v`
- [x] `pytest tests/ -k scala -q`
- [x] `pytest tests/contracts/ -q`
- [x] `mypy` / `ruff check` clean


aimasteracc:
Triaged Codex's P2 finding — confirmed real regression, fixed in 7818a488.

**Finding**: refactoring `analyze_file` to call `_build_scala_elements_dict()` directly (bypassing `self.extract_elements()`) dropped the inner try/except that caught extraction failures gracefully. An exception inside any single `extract_xxx()` call now propagated to the outer except, returning `success=False` with an empty source instead of the previous `success=True` with empty elements but preserved source_code/line_count.

**Fix**: wrapped the extraction call in `analyze_file` with its own try/except that falls back to an empty elements dict, matching `extract_elements()`'s existing contract. Added `test_analyze_file_extraction_exception_stays_successful` to lock this in.

PR #1161, opened 2026-07-22
perf(constraints): prefilter caller edges in SQLite

aimasteracc:
## Value

Moves caller-prefix rejection into SQLite before the Python constraint hot loop. This reduces rows materialized and regex-checked on every `change_impact` / `safe_to_edit` call while preserving constraint semantics.

## Safety bounds

- Literal caller prefixes use parameterized `instr(file_path, ?) = 1` predicates.
- Empty/wildcard prefixes retain the full-scan fallback.
- More than 256 distinct prefixes also fall back to a full scan, preventing an unbounded SQL statement or SQLite expression-depth failure.
- Regression tests cover wildcard fallback, irrelevant-edge rejection, and the large-prefix fallback.

## Evidence on latest `develop`

- Windows synthetic workload: 50,000 call edges × 5 rules, 8,000 violations, 7-run median 192.7 ms.
- Focused constraint suite with branch coverage: 19 passed, 1 quarantined performance test skipped.
- Patch coverage: no added executable misses.
- Full suite: 1275 passed, 1 skipped.
- `uv build`: source distribution and wheel built successfully.
- Ruff check/format, focused MyPy, and `git diff --check`: pass.


aimasteracc:
@codex review

aimasteracc:
@codex review

PR #1170, opened 2026-07-26
fix(index): define positive max_files contract

### LINKED ISSUE #1169: Define consistent `max_files=0` semantics across indexing surfaces
## Problem

`max_files=0` currently has incompatible meanings across public entry points:

- `cache/indexer.py::walk_and_partition` and incremental sync stop before processing the first file, so zero means “process none” (and may report truncation).
- Some MCP/CLI paths normalize with `value or default`, so zero silently means “use the default limit”.
- Full index passes the integer through directly, so its behavior differs from those callers.

The same user input can therefore disable indexing, select a default, or produce misleading truncation depending on the surface.

## Proposed contract

Choose and document one cross-surface meaning:

1. reject zero as invalid and require a positive integer, or
2. reserve zero for “unlimited” and represent “process none” separately.

The first option is likely safer because it prevents accidental unbounded work and removes truthiness-based coercion.

## Acceptance criteria

- Central validation/normalization is shared by CLI, MCP, full index, incremental sync, and low-level index walking.
- Boolean values are rejected even though `bool` is an `int` subclass.
- Negative, zero, omitted, and positive values have explicit documented behavior.
- No caller uses `value or default` for this option.
- Exact regression tests cover every public surface and truncation reporting.
- User-facing help/schema text states the chosen contract.

aimasteracc:
## Value

Closes #1169 by giving every project-indexing surface one explicit `max_files` contract: omitted selects the documented bounded default; positive integral values are normalized; zero, negative values, and booleans are rejected before any index mutation.

## What changed

- Added one shared normalizer for CLI, MCP, AST-cache walking, incremental sync, knowledge indexing, and cold-cache auto-warming.
- Added `minimum: 1` plus zero-invalid guidance to the relevant MCP schemas and CLI flags.
- Removed per-surface coercion differences, including cold incremental warm-up ignoring the requested cap.
- Added `truncated_by_max_files` to incremental results and both full-index phase reports.
- Preserved the established MCP compatibility for integral numeric strings/floats while rejecting booleans.
- Documented the contract in the CLI reference, MCP specification, and TSA indexing skill.

## Safety and regression coverage

- Invalid forced rebuilds are proven to fail before deleting an existing cache.
- Truncated scans are proven not to invalidate unseen indexed files.
- CLI payloads/parser flags, five MCP index tools, full-index phases, incremental sync, and the low-level AST path have exact tests.

## Verification

- Focused coverage: 228 passed.
- Patch coverage: no added executable misses.
- Full suite: 1273 passed, 7 skipped.
- `uv build`: sdist and wheel succeeded.
- Ruff check: passed.
- Changed-file Ruff format: passed.
- Focused MyPy: passed.
- `git diff --check`: passed.

Full-repository `ruff format --check .` still reports 24 pre-existing files from `develop`; this PR does not modify or reformat them.

### DIFF
--- a/docs/api/mcp_tools_specification.md
+++ b/docs/api/mcp_tools_specification.md
@@ -1517,6 +1517,12 @@ uv run tree-sitter-analyzer --parser-readiness --parser-readiness-include-suppor
 
 The v1.13.0 release adds 40 specialised tools for autonomous-agent workflows: AST cache + FTS5 indexing, function-level CodeGraph parity (callers/callees/blast-radius/visualise), pre-edit safety verdicts, anti-pattern / dead-code / route detection, architectural-constraint enforcement, and a persistent decision journal. All tools default to TOON output for 50-70% token savings versus JSON. Tools are listed alphabetically.
 
+For project-indexing tools, `max_files` is always a positive integer. Omitting
+it selects the tool's documented default. Zero, negative values, booleans, and
+non-integral values are rejected; zero is never an alias for “unlimited” or “no work.”
+Capped incremental results expose `truncated_by_max_files` so callers can
+distinguish a complete scan from a bounded prefix.
+
 ### 16. ast_cache
 
 **Purpose**: Pre-indexed AST cache with FTS5 search and incremental sync (CodeGraph parity). The only tool that persists AST data across sessions — other codegraph tools build on top of this cache.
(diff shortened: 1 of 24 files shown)


aimasteracc:
@codex review

aimasteracc:
Superseded by #1171 because this repository's GitFlow guard requires a `fix/*` head branch. The replacement points to the exact same reviewed tree/commit content.

PR #1234, opened 2026-08-08
feat(benchmarks): add zero-cost trust rehearsal

aimasteracc:
## Summary
- add an offline denial probe and a bound v2 fixture probe through attestation, judge, spec, and evidence-digest bindings
- seal synthetic evidence while invoking no model adapter, provider callback, or network request
- keep the receipt E0, unpublishable, and production dispatch disabled

## Boundary
The bound fixture assumes the independent-judge bit only to traverse the remaining verifier path. It is not independent evidence and does not authorize a real canary.

## Verification
- `uv run pytest tests/unit/test_production_rehearsal.py -q` → 6 passed
- integrated quick gate → 1388 passed, 1 skipped


aimasteracc:
@codex review

PR #1244, opened 2026-08-08
fix(ci): verify package aggregate identity shape

aimasteracc:
Fix protected run 31283976023: package aggregate uses source_commit + workflow, not axis source. Preserve exact push/run/attempt/ref identity.


aimasteracc:
@codex review

PR #1293, opened 2026-08-16
feat(task): P0.4 zero-write source-generation oracle (first slice)

aimasteracc:
## Summary

RFC-0022 **P0.4 read-existing groundwork — first slice**: a zero-write source-generation oracle. The task layer currently fails closed on every platform because `edit.impact(access_mode="read_existing")` has no zero-write backend; the strace authority (#1259) exists to certify one. This is the generation half of that backend.

**`tree_sitter_analyzer/diff_snapshot_readonly.py`** — `oracle_generation_readonly` mirrors the frozen oracle (`source_oracle_git.oracle_generation`) frame for frame (root identity, object format, HEAD, index bytes, dirty/untracked inventories, per-path content digests, settings, patch digest), but:

- every git invocation runs against the **live index** with `GIT_OPTIONAL_LOCKS=0` via its own bounded **zero-write runner** — no temporary index, no temporary `diff.orderFile` (the P0.4 invocation set needs no order file; a repository-level `diff.orderFile` config fails the route closed);
- the frozen oracle's conservative framing (its stat-invalidation makes git report every tracked path dirty) is replicated exactly, so the generation tokens are **byte-identical** to the frozen oracle's on identical source state.

**Differential contract tests** (`test_diff_snapshot_readonly.py`): token equality vs the frozen oracle across clean/dirty/untracked/staged states, determinism, source-change sensitivity, no temp-index materialization, read-only env invariants.

## Why this matters (agent-love mission P1)

The P0.4 gate is the blocker between "the task layer exists" and "the task layer runs end-to-end on Linux CI". This slice proves the generation half works read-only; the remaining work is in-memory blob/patch materialization and wiring `edit.impact` to the backend, then the Linux strace axis certifies zero writes.

## Gates

- Quick gate: 1992 passed
- ruff + mypy clean
- Patch coverage gate: no added executable misses
- Differential: `oracle_generation` vs `oracle_generation_readonly` tokens identical on all fixture states

## Follow-ups

- In-memory blob/patch materialization (old/new bytes, records) for the P0.2 golden corpus
- Wire `edit.impact(access_mode="read_existing")` to the backend
- Linux strace axis certifies the route; macOS authority remains a separate RFC-tracked item

### DIFF
--- a/tests/unit/test_diff_snapshot_readonly.py
+++ b/tests/unit/test_diff_snapshot_readonly.py
@@ -50,6 +50,7 @@ def _epochs(root: str, mode: str) -> tuple[GitEpoch, GitEpoch]:
 
 @POSIX_SNAPSHOT_TEST
 @POSIX_SNAPSHOT_TEST
+@POSIX_SNAPSHOT_TEST
 def test_clean_repo_generations_match(git_repo: str) -> None:
     for mode in ("diff", "staged"):
         frozen_epoch, readonly_epoch = _epochs(git_repo, mode)
@@ -58,6 +59,7 @@ def test_clean_repo_generations_match(git_repo: str) -> None:
         assert frozen_epoch.untracked_paths == readonly_epoch.untracked_paths
 
 
+@POSIX_SNAPSHOT_TEST
 @POSIX_SNAPSHOT_TEST
 @POSIX_SNAPSHOT_TEST
 def test_dirty_and_untracked_generations_match(git_repo: str) -> None:
@@ -86,6 +88,7 @@ def test_dirty_and_untracked_generations_match(git_repo: str) -> None:
 
 @POSIX_SNAPSHOT_TEST
 @POSIX_SNAPSHOT_TEST
+@POSIX_SNAPSHOT_TEST
 def test_generation_changes_when_source_changes(git_repo: str) -> None:
     import pathlib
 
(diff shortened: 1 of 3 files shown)


aimasteracc:
**Codex review triage — 6/7 fixed in `10beb1a1`** (regression tests for each):

| # | Comment | Verdict | Action |
|---|---|---|---|
| 1 | Temp-writing Git runner (P1) | Real | Every oracle git invocation now routes through `git_readonly.run_git_readonly` (zero writes, no order file) — including rev-parse/ls-tree/config, which previously used the temp-writing `git_output` |
| 2 | Index hint semantics (P1) | Real | assume-unchanged/skip-worktree parsed from captured index bytes; hinted paths excluded from content framing + dirty inventory (differential test with `update-index --assume-unchanged`) |
| 3 | Configured diff ordering (P1) | Real | Active `diff.orderFile` now fails closed (`DIFF_SNAPSHOT_UNSUPPORTED_ORDERFILE`) before the staged patch |
| 4 | Index reads bound to captured bytes (P1) | Real | Pure-Python index binary parser (v2/v3) replaces the live `ls-files` re-open — inventory derives from the exact framed bytes |
| 5 | Live object-directory dependency (P1) | **Deferred** | `git ls-tree` still reads the object DB (read-only). Loose/pack object reading in Python is the next slice; certification cannot be claimed until it lands — documented in the module + PR |
| 6 | Oversized module (P2) | Real | Zero-write runner split into `tree_sitter_analyzer/git_readonly.py`; oracle module back under 500 lines |
| 7 | check-attr output cap (P2) | Real | Ceiling raised to the bounded inventory bound (16 MiB) |

Verified: quick gate 1992 passed; ruff + mypy clean; local patch-coverage gate clean.

PR #621, opened 2026-06-12
feat(index): docstring/return_type/params serialized into symbols_json

aimasteracc:
Closes #614 (RFC-0016 prerequisite + BM25-enrichment arm — the arm was measured and REJECTED).

## Summary
`symbols_json` now carries `docstring` (PEP 257 first-statement string, capped 500 chars, Python-only this PR), `return_type` (generic field lookup — python/rust/ts), and `params` (already present, pinned). Absent values = absent keys, never empty-string noise. `_AST_CACHE_EXTRACTOR_VERSION` 5→6. Measured coverage on this repo: 20,918/34,409 Python symbols carry a docstring (60.8%); size +1.54% (+3.5 MB on 230 MB).

## The enrichment arm: measured, then dropped
The FTS docstring column was fully implemented and pilot-measured at TWO weights (10 queries, .recon/rfc0016-pilot-queries.md):
| | develop baseline | weight 1.0 (pod) | weight 0.3 (lead, query-side, fresh index) |
|---|---|---|---|
| target-in-top-5 | 5/10 | 4/10 (M1 control regressed) | 4/10 (M1 recovered #1, Q1 lost) |
| conceptual gaps closed (M2/Q2/Q4/Q5/Q6) | — | **0** | **0** |

Net ≤ baseline either way → the FTS half was surgically removed before this PR (commit history shows implement→measure→drop); only serialization ships. **This finalizes the data case for RFC-0016 step 2 (semantic arm)** — both cheap arms (stemming #606, BM25-enrichment) leave the same 5 conceptual queries unreachable.

## Test plan
- [x] RED-first serialization tests: 14 (cap==500 exact, absent-key semantics, multiline strip, rust return_type, non-python no docstring)
- [x] Regression 228 passed incl. agent contracts + porter stemming + cascade (-n 0); golden -k python zero drift
- [x] Ratchet exit=0 (merged develop post-batch-20); ruff/mypy clean
- [x] Both version-bump sites; size + coverage numbers measured (commands in pod report)

PR #609, opened 2026-06-12
fix(search): cascade re-sort dropped test-file demotion for conceptual queries

aimasteracc:
Closes #607 (found by the RFC-0016 pilot baseline).

## Root cause (4 hypotheses, each live-verdicted)
`fts_search_ranked` demotes correctly (its output IS production-first), but `search_symbols_cascade` re-sorts purely by `relevance_score` (`_ast_cache_search.py:78-85`) — the first production row dropped to position 21 and `_fts_to_results` truncated to limit before the tool-level demotion saw anything but tests. `query_wants_tests` misfire: refuted. No-production-match: refuted (1,014 prod rows in the raw window). Window saturation: refuted as cause.

## Fix (11 lines)
`rank_tier(file, wants_tests=query_wants_tests(query))` becomes the PRIMARY key of the cascade's final sort; secondary/tertiary unchanged. Both directions pinned: test-intent queries still surface tests first.

## Pilot delta (same fresh 46,935-symbol index)
| Query | before | after |
|---|---|---|
| M1 route-to-facade | NO (top-5 = 5 tests) | **YES** (`legacy_to_facade`) |
| Q2 UML file scope | NO (5 tests) | **YES** (`class_diagram` r2) |
| Q3/Q4/Q7 | NO | still NO — but top-5 test rows 5/5 → **0/5**; residual = genuine vocabulary gaps (Q3's labeled target `_STOP_WORDS` isn't even indexed — module-level variable extraction gap, flagged separately) |
| Q1/Q8 controls | YES | YES unchanged |

Symbol-search target-in-top-5: **2/10 → 4/10**; all-test pollution eliminated on all 5 polluted queries. Side effect: the same cascade feeds nav context and Hyphae #name selectors — their suites green.

## Test plan
- [x] RED-first: cascade-level + end-to-end tests reproduced the all-test top-5, GREEN after; test-intent direction pinned both ways (10/10)
- [x] Regression: 14 FTS/search consumer files 267 passed + fts_fast_path 27 (-n 0)
- [x] ruff/mypy clean; ratchet exit=0; patch coverage 98.86% (1 pre-existing miss)
- [x] Lead independently re-ran 37 tests green + push diff=0

PR #606, opened 2026-06-12
feat(search): FTS5 porter stemming — dispatching matches dispatch

aimasteracc:
Closes #604 (phase 0 of RFC-0016 — SOLO schema PR per house rules).

## Summary
`ast_symbols_fts` (the only FTS5 table) now uses `tokenize='porter unicode61'`. Porter applies symmetrically to index + query, so exact-identifier queries return identical production sets (pinned differentially old-DDL vs new-DDL); exact-match tiers are immune by construction (SQL `name = ?` / substring post-filter).

**Migration v12**: DROP + recreate + repopulate from `ast_symbol_rows` with rowid join preserved — existing caches upgrade in place, **no full reindex**; FTS5-less builds degrade silently (v3/v9 precedent); intentionally outside EXPECTED_SCHEMA_VERSIONS (environment-dependent table).

**Measured recall on this repo (46,929 FTS symbols)**:
| query | before | after |
|---|---|---|
| `dispatching` | **0 hits** | **39 (16 production)** incl. `dispatch_legacy`, `_dispatch`, `_dispatch_tool` |
| `dispatch` (exact) | 30 (16 prod) | 39 — production set identical 16/16, same order |

RFC-0016 re-baseline: stemming gap **closed**; residual = purely conceptual bridging + INFO-verdict honesty (recorded on #517 after merge).

## Test plan
- [x] RED-first: 5 stemming-dependent tests failed pre-fix, 4 unchanged-behavior pins green → 10/10
- [x] Differential pins: exact-token query sets identical under old vs new DDL
- [x] Regression: 11 FTS-consuming test files 283 passed + context tool 64 (-n 0)
- [x] ruff/mypy clean; ratchet exit=0
- [x] Lead independently re-ran suite + in-memory porter probe (dispatching→dispatch_legacy)

PR #1310, opened 2026-08-19
docs(rfc): RFC-0027 secretary layer — instant answers, comprehension, calibrated learning

aimasteracc:
## Summary

RFC-0025 specifies the **sensing** layers (L1–L5). Sensing alone does not make a secretary. RFC-0027 specifies the layers above it: **L6–L10**. It does not restate or modify RFC-0025's design.

| layer | what it adds |
|---|---|
| **L6** instant answer path | generation-keyed memoization, declared query cost, stable cross-turn symbol ids |
| **L7** comprehension | project card, deterministic reading tour, `start_here` for newcomers/contributors |
| **L8** judgment | minimal edit set, refactor queue in code, wrong-signal verdict fixes, enforceable cycles |
| **L9** calibration ledger | record predictions, ingest real outcomes, publish precision/recall |
| **L10** semantic index | embedding-backed similarity, structurally subordinate to resolved evidence |

## Why now — three measured problems

**1. The hottest path has no memoization.** Dogfood measurement (2026-08-19, this repo, in-process, 3 repeats):

| path | cold | warm |
|---|---|---|
| `edit action=safe` | 2.40 s | 2.41 s / 2.59 s — **no warm benefit at all** |
| `callers` | 24.8 s | 16–17 ms |
| `--callers` (CLI subprocess) | 28–35 s | 28–35 s every invocation |

`edit action=safe` is the one call an agent makes before **every** edit. A 2.4 s reflex is not proprioception. The `callers` cliff is RFC-0025 L1's problem; the *no repeat benefit* is L6's.

**2. Three high-value capabilities are built, tested, and reachable from nothing.**

| capability | implementation | reachable? |
|---|---|---|
| minimal edit set (rename) | `codegraph_refactor_tool.py:29` + `rename_symbol.py:291`, 15 passing tests | no facade, no CLI flag |
| project card (purpose + module descriptions) | `get_project_summary_tool.py:46` | no facade, no CLI flag |
| refactor priority formula | markdown skill only | 0 code hits, 0 tests |

`build_project_index_tool.py:244` routes agents to `next_step="get_project_summary"` — a name that resolves nowhere — and `tests/unit/cli/test_install_skills.py:571` **pins that orphan state as expected**. A test is currently protecting a wiring bug: the same shape as the 2026-06-08 TOON incident.

**3. "Gets smarter the more you use it" is currently false.** Exactly one read-back path exists, fed by a *manually written* journal, and `.ast-cache/decision_journal.db` does not exist on this repository — it has never been exercised. Nothing records whether an answer was right. Under CLAUDE.md §11 that makes it a belief, and beliefs rot silently.

## The core contribution: L9, the calibration ledger

Record every prediction → ingest the real outcome of the verification command TSA itself emitted → publish precision and recall per prediction kind, per project, pinned in CI.

**Recall is the honesty metric.** A missed dependent produces a wrong edit; a spurious one costs only tokens. So recall regressions are a release blocker and precision regressions are not.

This is the part of the design with no equivalent in current code-intelligence tooling. A competitor can clone a cache or an embedding index; cloning this means publishing their own error rate.

## The locked soundness fence (§L9.4, §L10)

Learned and similarity signal **may** re-rank results that resolved evidence already admits, **may** widen a candidate set with entries labelled `evidence: "observed_cochange"` / `"similarity"` plus a confidence, and **may** populate the observed-latency estimate.

They **may not** narrow or override a fail-closed verdict, promote an unresolved edge to resolved, or appear inside the certified `causal_envelope`.

> The conservative-resolution moat is the product. A learning loop that can talk the analyzer out of `unknown` destroys the exact property that makes it trustworthy. Learning makes the *ordering* smarter; it never makes the *claims* looser.

## Scope discipline

- **No new MCP facade.** All eight actions land on existing facades, so the RFC-0022 menu-experiment gate is untouched.
- **No LLM client** added to the package. L7 output is derived and byte-deterministic at a fixed generation.
- **No change** to the locked TOON-on-MCP / JSON-on-CLI default.
- **No public performance claim.** Every number is an E0 dogfood measurement; public wording stays bound by the RFC-0021 evidence ladder.
- L6 depends on RFC-0025 L1 for the generation stamp but does not specify it.

## Also in this PR

RFC-0025, RFC-0026 and RFC-0027 added to `rfcs/README.md` — the index had not been updated since RFC-0024, so the two active proprioception RFCs were unlisted.

## Review focus

1. Is the generation stamp genuinely sufficient as a sound cache key (§L6.1), or is there an input that can change a certified answer without bumping it?
2. Is the L9.4 fence tight enough to state as **locked**?
3. Open question 1: process-local vs persisted answer cache. Persisting is what kills the 28–35 s per-CLI-invocation cost, but widens the blast radius of a wrong generation stamp.

### DIFF
--- a/rfcs/0027-agent-secretary-and-calibration.md
+++ b/rfcs/0027-agent-secretary-and-calibration.md
@@ -5,10 +5,15 @@
 - **Created**: 2026-08-19
 - **Last updated**: 2026-08-19
 - **Tracking issue**: TBD
-- **Relationship to other RFCs**: complements [RFC-0025](0025-instant-causal-proprioception.md)
-  (the *sensing* layers L1–L5). This RFC specifies the layers *above* sensing:
-  L6–L10. It does not restate or modify RFC-0025's design; where it depends on a
-  RFC-0025 layer, the dependency is named explicitly.
+- **Relationship to other RFCs**:
+  - complements [RFC-0025](0025-instant-causal-proprioception.md) (the *sensing*
+    layers L1–L5). This RFC specifies the layers *above* sensing: L6–L10. It does
+    not restate or modify RFC-0025's design; where it depends on a RFC-0025
+    layer, the dependency is named explicitly.
+  - §L10 **supersedes [RFC-0016](0016-semantic-symbol-search.md) conditionally**.
+    RFC-0016 was rejected on measured evidence (2026-06-13 embedding pilot,
+    2/5 on the conceptual-gap gate). L10 may only proceed after a re-pilot
+    passes; see §L10 for the blocking precondition.
 - **Affected source paths** (pin them — reviewers watch for drift here):
   - `tree_sitter_analyzer/cache/answer_cache.py` (new — L6)
   - `tree_sitter_analyzer/mcp/tools/get_project_summary_tool.py` (L7 — currently unreachable)
@@ -120,8 +125,10 @@ class AnswerKey:
 
     tool: str
     action: str
-    normalized_args: str   # canonical JSON, keys sorted, project-relative paths
-    generation: str        # e.g. "idxsrc-v3:..."
+    normalized_args: str    # canonical JSON, keys sorted, project-relative paths
+    generation: str         # source-tree state, e.g. "idxsrc-v3:..."
+    producer_version: str   # action_version + schema version + resolver-rule digest
+    extra_inputs: str       # digest of declared non-source inputs (config, constraints)
 
 
 def lookup(key: AnswerKey) -> CachedAnswer | None:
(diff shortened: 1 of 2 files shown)


aimasteracc:
## Codex review triage — all 8 findings, all valid, all fixed in `e1e8bc3b`

Per the locked rule, every comment gets an explicit verdict. **Zero were dismissed.**

| # | finding | verdict | fix |
|---|---|---|---|
| C1 | producer versions missing from the cache key | **VALID** | Key now carries `producer_version` + `extra_inputs`. The source generation alone does *not* determine a certified answer — a TSA upgrade, `action_version` bump, resolver-rule change, or a `architectural-constraints.yml` edit all change it at an unchanged tree. As written the cache would have replayed the previous release's verdict under an old schema. |
| C5 | cache hits on side-effecting routes | **VALID** | Added an explicit `CACHEABLE_ACTIONS` allowlist plus a contract test asserting it is a **subset** of the audited side-effect-free set, so it cannot be widened by hand alone. Index-lifecycle / doc-sync / snapshot-acquire / `record_outcome` are structurally excluded. |
| C7 | 24-bit short ids collide | **VALID — severe** | Confirmed the math: birthday bound puts 50% collision at ~**4,800 symbols**; at the RFC-0025 target of 100k, collisions are effectively certain. Canonical full RFC-0023 identity is now always present; the short form is a git-style abbreviation lengthened until unique **within the project index**, with `SYMBOL_ID_AMBIGUOUS` as a fail-closed resolution error that lists candidates rather than picking one. |
| C6 | `plan_rename` could write | **VALID** | `mode="preview"` pinned internally; apply-like arguments **rejected**, not forwarded — the mode is not a caller parameter at this surface at all. Contract test proves zero filesystem writes on adversarial input. |
| C2 | test recall scored against failures | **VALID** | This was the sharpest catch. A green verification has an empty failure set, so a *perfectly correct* prediction would have scored as all-false-positive and recall would be undefined on every passing run. Outcome now records `tests_executed` as the comparison set; `tests_failed` is retained as the strongest positive confirmation. |
| C3 | `risk_verdict` had nothing to calibrate against | **VALID** | Added `verification_passed` + `escaped_allowed_paths` and a pinned derivation table (`benign` / `harmful` / `indeterminate`). `indeterminate` is excluded from the denominator and counted separately, so an unrun verification cannot be laundered into either a good or a bad score. |
| C8 | zero-denominator undefined | **VALID** | Pinned an exact empty-set table. `None` = undefined-for-this-observation: excluded from that metric's aggregate but **still counted in the observation total**, so the aggregate cannot be inflated by dropping rows. `trivial` observations get their own count so an inactive project cannot show a perfect score. |
| **C4** | **L10 reintroduces rejected RFC-0016** | **VALID — authoring failure on my part** | see below |

### C4 in full

Verified: `rfcs/0016-semantic-symbol-search.md` is `Status: rejected (data-driven, 2026-06-13 — embedding pilot at deployment scale scored 2/5 on the conceptual-gap gate)`, after stemming (#606), demotion (#609) and BM25-docstring (#621) were each measured first.

The original RFC-0027 proposed the same design and **did not mention this**. That is precisely the CLAUDE.md §11 failure mode — designing from belief without checking the existing measurement — committed in the RFC that cites §11 as its own motivation.

Nothing in RFC-0027 makes embeddings retrieve better. The only genuine difference is positioning: RFC-0016 put semantic search forward as a **retrieval surface** competing with BM25 on top-5 hit rate; L10 puts it forward as a **subordinate discovery aid** that can never be an answer by itself. That is a hypothesis, not a result.

So L10 now **conditionally supersedes** RFC-0016 and carries a **blocking acceptance precondition**: re-run RFC-0016's pilot harness unchanged, plus one added measurement for the new positioning. If it fails, **L10 is struck, RFC-0016 stays `rejected`, and the rest of RFC-0027 ships without it** — the same disposition RFC-0016 itself took.

### Test plan additions

Six new cases: upgrade-replay cache miss, allowlist-subset property, `plan_rename` write-freedom under adversarial input, ambiguous-id resolution, all-green recall (the C2 case), and every row of both new tables with exact values.

aimasteracc:
@codex review

PR #1311, opened 2026-08-19
fix(cache): fail closed on unmodelled dynamic loader positions

aimasteracc:
## Why

`import_projection_complete` defaulted to `True` and was flipped to `False` at
~20 enumerated syntax sites. That is an **allowlist of dangerous shapes**: any
position the walker had not enumerated certified the projection while the
dependency edge was silently dropped — fail-**open**.

The set of positions a value can flow through in JS/TS/Python is effectively
unbounded (tagged templates, decorators, default params, destructuring
defaults, generators, optional chaining, getters, Proxy, …), so enumeration can
never terminate. Five review rounds each added more sites and each left more
open.

Five holes confirmed empirically on the base branch (all returned
`complete=True` with the dependency absent):

| case | language | code |
|---|---|---|
| walrus eval alias | python | `(run := eval)("x")` |
| runpy | python | `import runpy` / `runpy.run_module("pkg.util")` |
| `__builtins__` loader | python | `load = __builtins__.__import__` / `load("pkg.util")` |
| `process.getBuiltinModule` | javascript | `process.getBuiltinModule("node:module").createRequire` |
| throw/catch transfer | javascript | `try { throw require; } catch (load) { load("./util.js"); }` |

## What

**Invert the default to a denylist.** A file's import projection is complete
only if *every* reference to a loader-visible name occupies a position the
walker also projects:

- the callee of a call on a **static module literal** (`require("./util.js")`,
  `import("./util.js")`, `importlib.import_module("pkg.util")`, `__import__("pkg")`)
- a **file-scoped / module-scope alias binding** the alias tracker follows
  (`const load = require;` / `load = importlib.import_module`)
- the operand of a **`typeof` query** (see precision section below)

Everything else — an argument, a return value, a thrown value, a field, a
collection element, a walrus target, an unrecognised owner, or **any node kind
not explicitly whitelisted** — sets `complete = False`. Sound by construction:
unknown syntax degrades to an honest `unknown` instead of a false certification.

Also projects `runpy.run_module` / `runpy.run_path`: they execute a module, so
they carry a real dependency edge that was previously invisible.

Extractor cache version 37 → 38.

## Precision: `typeof` and `import.meta` (review round 7)

A first cut of this inversion failed closed on two positions that are
*provably* safe, which withheld the causal envelope from a large fraction of
real-world JavaScript. Both are fixed in `147bc8f6`.

| construct | before | after |
|---|---|---|
| `typeof module !== 'undefined'` | False | **True** |
| `typeof require === 'function'` | False | **True** |
| `typeof window !== 'undefined'` | False | **True** |
| `import.meta.url` | False | **True** |
| `module.exports = f` | True | True |
| `globalThis.x = 1` | True | True |

**`typeof` is a proof, not a heuristic.** It evaluates its operand and yields
that operand's *type tag* as a string; it cannot retain, alias, call, or
propagate the value. There is no program in which `typeof require` results in
the loader escaping — so the operand belongs in the safe set on the same
footing as a direct `require("literal")` callee. The cost of omitting it was
not hypothetical: `typeof module !== 'undefined'` is the canonical UMD guard
and `typeof window !== 'undefined'` is in essentially every isomorphic file.

Only **reference links** (parentheses and member access) are walked to reach
the operand, so a call nested inside it is still classified on its own merits.
`delete` shares the `unary_expression` node but *rebinds* its operand, so the
operator is matched exactly and `delete require;` stays False.

**`import.meta` was a real diagnosis, not a special case.** `import.meta`
parses as a `meta_property` node, **not** as a member of the dynamic-import
operator, so the named `import` token inside it was classified as a loader
reference and the member-path whitelist was never reached. The `"meta"` entry
in `_JSTS_SAFE_LOADER_MEMBERS` was therefore unreachable dead config; it is
removed, and the `import` token of a `meta_property` is now treated as a
static keyword exactly as for `import_statement`.

The safe set was widened by **`typeof` only**. `void` was considered — it is
the same proof class — and deliberately **excluded**: no measured benefit and
no evidence of use. Verified in both the javascript and typescript grammars:

```
--- MUST STAY False: the five confirmed holes + export default ---
  False | (run := eval)("x")
  False | load = __builtins__.__import__ / load("pkg.util")
  False | process.getBuiltinModule("node:module").createRequire ...
  False | try { throw require; } catch (load) { load("./util.js"); }
  False | export default require;
--- runpy row: dependency projected ---
  True  ['import runpy', 'runpy.run_module("pkg.util")']
--- adversarial (typeof must not whitelist the identifier everywhere) ---
  False | const t = typeof require; registry.push(require);
  False | const t = typeof eval("require(0)");
  False | delete require;
  False | delete module.require;
  False | typeof require.main; const m = require.main;
```

## Net deletion

`_symbol_walker.py`: **1880 → 1197 lines (−683, −36%)**. 25 enumeration
helpers became unreachable and are deleted (verified zero remaining references
repo-wide):

`_python_value_stores_loader`, `_python_loader_mapping_owner`,
`_python_loader_call_result`, `_python_mapping_has_dynamic_loader_owner`,
`_python_function_annotations`, `_python_control_flow_header_values`,
`_python_module_control_bindings`, `_python_assignment_target_leaves`,
`_jsts_value_stores_module_loader`, `_jsts_pattern_binds_module_loader`,
`_jsts_module_owner_reference`, `_jsts_global_owner_reference`,
`_jsts_indirect_module_loader_call`, `_jsts_require_utility_member`,
`_jsts_dynamic_module_member`, `_jsts_global_eval_reference`,
`_jsts_static_first_argument`, `_jsts_node_module_api_call`,
`_jsts_import_exposes_create_require`, `_jsts_node_create_require_reference`,
`_jsts_module_loader_factory_call`, `_jsts_pattern_selects_require_property`,
`_jsts_pattern_selects_create_require_property`, plus the
`_mark_jsts_loader_binding` position enumeration and the nested-scope block of
`_python_dynamic_loader_analysis`.

The `createRequire` rule alone subsumes five of those JS helpers: **any**
member access or import specifier named `createRequire` is unsafe unless a
file-scoped declaration shadows it — which is why the previously-unreachable
`process.getBuiltinModule(...).createRequire` chain now closes without adding
a new enumerated site.

## RED-first

All 7 new assertions failed on the base branch before the fix; the 4 baseline
pins passed throughout.

## Changed existing assertions

| test | old | new | verdict |
|---|---|---|---|
| `const { other: value } = module` | True | **False** | Old assertion was a precision carve-out requiring object-pattern key modelling. Retaining the `module` object is not provably safe; failing closed is sound. |
| `getattr(importlib, "other")` | True | **False** | Handing out the module object that carries the loader is not provably safe. |
| `importlib.__dict__.get("other")` | True | **False** | Same — dunder access on a loader-carrying module. |
| `vars(importlib).get("other")` | True | **False** | Same. |
| `_python_module_control_bindings` unit test | — | deleted | Helper is unreachable after the inversion. |
| `test_malformed_indirect_commonjs_loader_member_is_not_matched` | — | replaced | White-box test of 7 deleted helpers; replaced by 3 focused tests on the surviving `_jsts_reference_path`. |
| extractor version fixtures (37 → 38) | — | mechanical | Cache invalidation for the semantic change. |

No assertion was weakened to make the inversion pass.

## Base branch note

The code this fixes only exists on `feature/proprioception-causal-envelope`
(PR #1308) — `import_projection_complete` is not on `develop`. This PR is
therefore **stacked on #1308** rather than targeting `develop` directly, so the
diff stays focused on this one change instead of dragging in #1308's 49
commits. Retarget to `develop` after #1308 merges.

## Verification

- `tests/unit/test_symbols_json_enrichment.py`: 314 passed
- `tests/unit/mcp/test_safe_to_edit_tool.py`: 324 passed, 8 skipped
- `uv run ruff check .`: All checks passed
- `uv run mypy tree_sitter_analyzer/`: 23 errors — identical to the base-branch baseline
- `pre-commit` (all hooks, changed files): all Passed
- `tests/unit` full: 43 failures before and after; the differing entries are the
  known flaky Windows symlink/coverage rotation and reproduce identically on the
  base branch
- `--change-impact --change-impact-mode branch` focused command: 952 passed,
  78 skipped, 2 failed — both pre-existing baseline failures

### Pre-existing baseline — this PR adds none of it

So a reviewer does not attribute these to this change, both numbers were
measured on the **base branch** (`feature/proprioception-causal-envelope`) by
stashing this work:

| metric | base branch | this PR |
|---|---|---|
| `uv run mypy tree_sitter_analyzer/` | **23 errors** | **23 errors** (unchanged) |
| `uv run pytest tests/unit` | **43 failures** | **40 failures** |

The mypy count is identical — an intermediate draft introduced 2
`no-any-return` errors and both were fixed before commit. The `tests/unit`
failures are a flaky Windows rotation (symlink permissions, coverage-report
reads, TOON boundary) whose membership varies run to run; every differing
entry was re-run in isolation on the base branch and reproduced there
identically. **No failure in either column is caused by this change.**

### DIFF
--- a/tests/unit/mcp/test_safe_to_edit_tool.py
+++ b/tests/unit/mcp/test_safe_to_edit_tool.py
@@ -3969,12 +3969,12 @@ def test_snapshot_syntax_envelope_keeps_complete_exercising_tests() -> None:
         "INSERT INTO ast_index VALUES "
         "('app.py', '[]', "
         '\'{"truncated_depth": false, "import_projection_complete": true, '
-        '"syntax_error": false}\', 37)'
+        '"syntax_error": false}\', 38)'
     )
     conn.executemany(
         "INSERT INTO ast_index VALUES (?, '[]', "
         '\'{"truncated_depth": false, "import_projection_complete": true, '
-        '"syntax_error": false}\', 37)',
+        '"syntax_error": false}\', 38)',
         [(f"tests/test_app_{index}.py",) for index in range(12)],
     )
     conn.executemany(
@@ -4011,7 +4011,7 @@ def test_snapshot_syntax_envelope_excludes_unrelated_nearby_test() -> None:
     conn.executemany(
         "INSERT INTO ast_index VALUES (?, '[]', "
         '\'{"truncated_depth": false, "import_projection_complete": true, '
-        '"syntax_error": false}\', 37)',
+        '"syntax_error": false}\', 38)',
         [("app.py",), ("tests/test_app.py",)],
     )
 
@@ -4529,7 +4529,7 @@ def test_snapshot_syntax_envelope_certifies_single_java_file() -> None:
     conn.execute(
         "INSERT INTO ast_index VALUES (?, ?, "
         '\'{"truncated_depth": false, "import_projection_complete": true, '
-        '"syntax_error": false}\', 37)',
+        '"syntax_error": false}\', 38)',
         (
             "src/main/java/com/acme/Util.java",
             json.dumps([{"text": "package com.acme;"}]),
@@ -4638,7 +4638,7 @@ def test_snapshot_syntax_envelope_rejects_conflicting_java_packages() -> None:
     conn.executemany(
         "INSERT INTO ast_index VALUES (?, ?, "
         '\'{"truncated_depth": false, "import_projection_complete": true, '
-        '"syntax_error": false}\', 37)',
+        '"syntax_error": false}\', 38)',
         [
             (
                 target,
@@ -4808,7 +4808,7 @@ def test_snapshot_syntax_envelope_rejects_uncaptured_include_root() -> None:
     conn.executemany(
         "INSERT INTO ast_index VALUES (?, ?, "
         '\'{"truncated_depth": false, "import_projection_complete": true, '
-        '"syntax_error": false}\', 37)',
+        '"syntax_error": false}\', 38)',
         [
             (
                 importer,
@@ -5061,7 +5061,7 @@ def _symbol_conn(raw_symbols: object) -> sqlite3.Connection:
         "CREATE TABLE ast_index ("
         "file_path TEXT, symbols_json TEXT, extractor_version INTEGER)"
     )
-    conn.execute("INSERT INTO ast_index VALUES ('app.py', ?, 37)", (raw_symbols,))
+    conn.execute("INSERT INTO ast_index VALUES ('app.py', ?, 38)", (raw_symbols,))
     return conn
 
 
--- a/tests/unit/test_symbols_json_enrichment.py
+++ b/tests/unit/test_symbols_json_enrichment.py
@@ -23,7 +23,6 @@ from tree_sitter_analyzer.cache import _symbol_walker as walker_module
 from tree_sitter_analyzer.cache._symbol_walker import (
     _cpp_code_mask,
     _python_dynamic_loader_analysis,
-    _python_module_control_bindings,
     _python_module_scope_statements,
     _SymbolWalker,
     _walk_for_symbols,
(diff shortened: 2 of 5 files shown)


aimasteracc:
@codex review

PR #1312, opened 2026-08-19
fix: resolve import edges across directories on Windows, and admit Node module suffixes

aimasteracc:
Stacked on #1311 (`feature/loader-projection-fail-closed`), itself stacked on #1308. Target will be retargeted to `develop` with the rest of the stack.

Two independent changes, in order of importance. The first was found while building an honest end-to-end test for the second, and is much larger than the work that uncovered it.

---

# 1. Windows: import edges never resolved across a subdirectory (P2, pre-existing)

**This was not introduced by this PR.** It has been shipping. It is fixed here because the public-API test for part 2 could not pass without it.

**Root cause, one sentence:** the AST cache keys `file_path` POSIX-style (forward slashes), while both resolvers in `import_graph.py` built candidate paths with `os.path.join`/`os.path.normpath` — which emits **backslashes on Windows** — so the `candidate in project_files` membership test could never match once a path contained a separator.

**Blast radius:**

- **Language-independent.** Python is affected identically to JS/TS — this is not a Node-suffix issue.
- **Any project with a subdirectory.** Flat, root-level imports resolved fine (no separator to mismatch), which is exactly why it hid: the existing fixtures are flat (`a.py` → `b.py`).
- Since essentially every real project has subdirectories, **the import graph has been effectively empty for Windows users**, and `health action=imports`, cycle detection, and cross-file resolution all inherited it.
- **Invisible on Linux CI** — `normpath` already produces POSIX separators there, so every axis stayed green.

**Measured on a real `ASTCache` index, before → after:**

| fixture | before | after |
|---|---|---|
| nested `.mts` ESM (`src/app.mts` → `src/util.mts`) | **0 edges** | 1 |
| nested `.cjs` require (`src/legacy.cjs` → `src/helper.cjs`) | **0** | 1 |
| nested `.js` ESM (`src/app.js` → `src/util.js`) | **0** | 1 |
| nested Python (`pkg/a.py` → `pkg/b.py`) | **0** | 1 |
| flat Python (`a.py` → `b.py`) | 1 | 1 (unchanged) |

Independently reproduced by the reviewer on clean `develop` with a separate fixture: `edge_count=1` on develop (flat resolves, nested does not) vs `edge_count=2` with this PR.

**Fix:** `_file_lookup()` indexes `project_files` by a separator-agnostic key, and `_match_file()` probes through it. Values are the **original** cache keys, so returned edges still reference real paths. Both `_resolve_python_import` and `_resolve_js_import` now go through it.

## How this was missed

The existing tests called the private resolver helpers (`_resolve_js_import`, `_resolve_python_import`) directly, passing path sets the test itself had constructed. They proved **the helpers worked**; nothing proved the **public path** did, or that the helpers were ever reached with production-shaped data. My own first-round tests repeated the mistake in a sharper form — I keyed the fake-cache fixture with `_np()`, normalising the fixture *to the implementation* instead of to production, which masked the separator mismatch completely.

What I added: **public-API tests that drive `ImportGraph.build()` and `CrossFileResolver.build()` against a real `ASTCache` index**, with POSIX keys as the real cache emits them — no fake cache, no path massaging. `test_build_over_real_indexed_mts_project_produces_edge` and `test_build_over_real_indexed_nested_python_produces_edge` exist specifically so this cannot silently return to zero. Same failure shape as CLAUDE.md §11: a conformance test on a private helper cannot notice that the helper is never called.

---

# 2. Node module suffixes (`.mjs`/`.cjs`/`.mts`/`.cts`) missing from the graph fingerprint and four resolvers

`#1308` taught `project_graph.py` to resolve the four Node module suffixes. Five other hand-maintained suffix tables did not get the memo.

### 2a — P1: stale graph cache

`cache/fingerprint.py` `_SOURCE_EXTS` listed 16 extensions and **none** of the four. `compute_graph_fingerprint` returns `(file_count, max_mtime_ns)` scanned over that list only, so adding, removing or editing a `.mjs` file changed **neither** component. The fingerprint compared equal, `DependencyAnalysisTool._get_graph()` and `DependencyGraph._cache_key_for()` saw no change, and a **stale graph was reused** — dependency and blast-radius answers stayed wrong until an unrelated recognized file changed or the process restarted.

### 2b — P2: dropped import edges in four resolvers

| Location | Was | Problem |
|---|---|---|
| `import_graph.py:265` | `("", ".js", ".ts", ".jsx", ".tsx", "/index.js", "/index.ts")` | `./util` never probed `util.mts` |
| `call_graph.py:277`, `:286` | `("", ".py", ".js", ".ts", ".jsx", ".tsx")` | same, both branches |
| `cross_file_resolver.py:276` | `(".js", ".jsx", ".ts", ".tsx")` | **worse than a miss** — it *strips* the extension, and `"util.mts".endswith(".ts")` is `True`, so shortest-first registered `util.mts` under the mangled name `util.m`. Now longest-first. |
| `cross_file_resolver.py:440` | `("", ".py", ".js", ".ts", "/__init__.py")` | the **lookup** side. Registration stored `src/util.mts` correctly and this never found it. Now selected by the importing file's language, kept as **two disjoint lists** — a Python file must not probe `.mts`, a `.mts` file must not probe `__init__.py`. |

### 2c — P2: the dispatch made the JS probe unreachable for ESM

`import_graph.py` `_resolve_import` routed on a **text prefix**. `import { run } from './util'` starts with `import `, so every ESM statement went to the **Python** resolver and only `require(...)` ever reached the JS branch. The suffix table fixed in 2b was therefore **dead code for exactly the `.mjs`/`.mts` files this PR exists to serve.**

Dispatch is now on the **source file's language** via `EXT_TO_LANG`. `import ` is ambiguous between Python and ESM and that ambiguity was the root cause; the old text sniff survives only as a fallback for extensions absent from `EXT_TO_LANG`, so unmapped languages don't regress.

### 2d — P2: dynamic-import regex accepted a literal prefix of a computed specifier

`safe_to_edit_helpers.py` matched `import("./dev.js" + suffix)` and recorded `./dev.js` as if it were the complete runtime specifier — **inventing an edge to a module that is never loaded while omitting the one that actually is.** The template-literal branch had the same hole (`` [^`$]+ `` blocks `${...}` interpolation but not `` `./dev.js` + s ``). Both now require the closing delimiter to be followed by `,` (legal import options: `import(spec, { with: { type: "json" } })`) or the call's `)`.

**Decision:** a computed specifier yields **no spec at all**, not a different one. The runtime target is unknowable statically; recording nothing is a known-unknown, whereas recording the prefix is an actively wrong edge that *also* hides the real one.

---

# 3. Structural: five copies of the extension list collapsed into one derived constant

The root cause of every part-2 bug was the same — the extension surface was duplicated across five files and the copies drifted. `tree_sitter_analyzer/constants.py` now owns it:

| Constant | Purpose |
|---|---|
| `JS_TS_MODULE_EXTS` | the 8 JS/TS module extensions in **probe** order (plain first, so `./util` still prefers `util.js` over `util.mjs` — existing precedence unchanged) |
| `JS_TS_INDEX_SUFFIXES` | **derived** `/index.<ext>` forms |
| `JS_TS_MODULE_EXTS_LONGEST_FIRST` | **derived**, longest-first, for the call sites that *strip* rather than append |
| `JS_TS_LANGUAGES` | `EXT_TO_LANG` language names sharing the Node module system, for language-based dispatch |
| `GRAPH_SOURCE_EXTS` | full graph source surface, now shared by `fingerprint.py` and `project_graph.py` so the fingerprint can never again ignore an extension the graph indexes |

`LONGEST_FIRST` is deliberately kept separate rather than collapsed: append-probe and strip semantics need opposite orders, and merging them is what produced the `util.m` bug. A comment at each site says why.

---

## Audit method — corrected mid-review

My first report claimed "no additional incomplete suffix tables" after grepping for `.mjs/.cjs/.mts/.cts`. **That method cannot find a table that lacks those suffixes** — it only finds the ones already correct. It missed `cross_file_resolver.py:440` entirely.

Re-audited by **shape**: every tuple/list/set literal containing ≥2 dotted extensions across the whole package — 26 JS/TS-family hits, each given an explicit verdict.

- **4 needed the suffixes → fixed here**, including `nav_facade.py:160` which listed `.mjs`/`.cjs` but not `.mts`/`.cts`.
- **7 correctly do NOT need them** (by-design partitions: the JS plugin owns `.mjs`/`.cjs` while the TS plugin owns `.mts`/`.cts`; JSX detection, since `.mts` cannot contain JSX; the deliberate allowJs set). Two of these were **false positives of my own detector**, which reports at line granularity and split dicts that are complete as a whole — every hit needed a human verdict.
- **8 genuinely incomplete but a different cluster → deliberately NOT fixed here.** `middleware_detector`, `test_gap_analyzer`, `query_symbol_search`, `symbol_lineage_tool`, `trace_impact_tool`, `_codegraph_explore_helpers`, `gitignore_detector`, `_typescript_formatter_signatures_table`. These are "which files count as source?" inventory sets, not import resolvers. This is a **real shipping gap, not theoretical**; it is filed as a separate focused follow-up rather than turning a resolver PR into a 14-file cross-cutting diff.

---

## Files changed (13 across two commits)

| File | Why |
|---|---|
| `import_graph.py` | separator fix (§1), language dispatch (§2c), suffix probe (§2b) |
| `cross_file_resolver.py` | registration strip order + lookup probe (§2b) |
| `cache/fingerprint.py` | stale-cache fix (§2a) |
| `project_graph.py` | both tables now use the shared constants |
| `call_graph.py` | suffix probe, both branches (§2b) |
| `mcp/tools/utils/safe_to_edit_helpers.py` | dynamic-import regex (§2d) |
| `mcp/tools/nav_facade.py` | missing `.mts`/`.cts` |
| `constants.py` | the shared constants (§3) |
| 5 existing test files | 30 new tests — **no new test files** (T-1) |

### Behavior expansions worth flagging

- `project_graph` gains `/index.jsx` and `/index.tsx` probes (its index list previously skipped those two while listing the other six). Additive only.
- `import_graph` gains `/index.*` forms for all 8 extensions (previously only `.js`/`.ts`). Additive only.

## Answers to the review questions

**`typescript_plugin/plugin.py:41` — verified, NO change needed.** It already returns `['.ts','.tsx','.d.ts','.mts','.cts']`, and `JavaScriptPlugin` already returns `.mjs`/`.cjs`. `EXT_TO_LANG` and `detect_language_from_file` map all four. The plugin surface was never the gap; the graph/resolver tables were.

**Template-literal variant — same hole, fixed.** See §2d.

## Verification

- **RED first:** 30 new tests, **26 fail** on the relevant parent commit (20 of the 23 from commit 1, 6 of the 7 from commit 2). The 4 that pass on both are deliberate guards — they pin that the tightened dynamic-import regex still accepts legitimate specifiers, and that language-based dispatch did not break Python. The load-bearing ones go through the public API against a real index.
- Emitted `--change-impact` focused command → **184 passed, 3 skipped**
- `uv run ruff check .` → **All checks passed!**
- `uv run mypy tree_sitter_analyzer/` → **23 errors** = the baseline re-measured by stashing (not 24)
- `uv run pre-commit run --files <changed>` → **every hook Passed** (mypy, bandit, Weak Assertion Ratchet, T-1 banned names)
- `uv run pytest -q` → 14 failures, the **same 14 files** as the base commit (`test_collect_no1_006b_baseline`, `test_fresh_install_qualification`, `test_outdated_uv_qualification`, `test_ast_cache`, `test_codegraph_context_tool`) — pre-existing, unrelated
- **E2E dogfood:** `src/app.mts → src/util.mts` and `src/legacy.cjs → src/helper.cjs` both present; `CrossFileResolver` binds `run → src/util.mts`

Repo rules honored: no new test files (T-1), exact assertions only, one behavior per test, `fix:` prefix.

### DIFF
--- a/tests/unit/mcp/test_graph_cache_invalidation.py
+++ b/tests/unit/mcp/test_graph_cache_invalidation.py
@@ -75,6 +75,48 @@ class TestGraphFingerprint:
         fp3 = compute_graph_fingerprint(str(project_root))
         assert fp3.file_count == fp1.file_count
 
+    def test_changes_on_added_mjs_file(self, project_root: Path) -> None:
+        """A new ``.mjs`` file must bump file_count.
+
+        ``project_graph`` resolves ``.mjs``/``.cjs``/``.mts``/``.cts``, so the
+        fingerprint must admit them too — otherwise the graph cache never
+        invalidates on Node module edits and stale answers are served.
+        """
+        fp1 = compute_graph_fingerprint(str(project_root))
+        (project_root / "pkg" / "esm.mjs").write_text("export const a = 1;\n")
+        fp2 = compute_graph_fingerprint(str(project_root))
+        assert fp2.file_count == fp1.file_count + 1
+
+    def test_changes_on_added_mts_file(self, project_root: Path) -> None:
+        """A new ``.mts`` file must bump file_count."""
+        fp1 = compute_graph_fingerprint(str(project_root))
+        (project_root / "pkg" / "esm.mts").write_text("export const a: number = 1;\n")
+        fp2 = compute_graph_fingerprint(str(project_root))
+        assert fp2.file_count == fp1.file_count + 1
+
+    def test_changes_on_added_cjs_file(self, project_root: Path) -> None:
+        """A new ``.cjs`` file must bump file_count."""
+        fp1 = compute_graph_fingerprint(str(project_root))
+        (project_root / "pkg" / "legacy.cjs").write_text("module.exports = {};\n")
+        fp2 = compute_graph_fingerprint(str(project_root))
+        assert fp2.file_count == fp1.file_count + 1
+
+    def test_changes_on_added_cts_file(self, project_root: Path) -> None:
+        """A new ``.cts`` file must bump file_count."""
+        fp1 = compute_graph_fingerprint(str(project_root))
+        (project_root / "pkg" / "legacy.cts").write_text("export = {};\n")
+        fp2 = compute_graph_fingerprint(str(project_root))
+        assert fp2.file_count == fp1.file_count + 1
+
+    def test_changes_on_touched_mjs_file(self, project_root: Path) -> None:
+        """Editing an existing ``.mjs`` file must move max_mtime_ns."""
+        (project_root / "pkg" / "esm.mjs").write_text("export const a = 1;\n")
+        fp1 = compute_graph_fingerprint(str(project_root))
+        time.sleep(0.05)  # ensure mtime granularity
+        os.utime(project_root / "pkg" / "esm.mjs")
+        fp2 = compute_graph_fingerprint(str(project_root))
+        assert fp2.max_mtime_ns > fp1.max_mtime_ns
+
     def test_idempotent_when_unchanged(self, project_root: Path) -> None:
         """Calling twice without edits returns the same fingerprint."""
         fp1 = compute_graph_fingerprint(str(project_root))
--- a/tests/unit/mcp/tools/test_safe_to_edit_dependency_view.py
+++ b/tests/unit/mcp/tools/test_safe_to_edit_dependency_view.py
@@ -312,6 +312,34 @@ def test_iter_dependency_source_files_skips_hidden_files(tmp_path: Path) -> None
     assert [path.name for path in files] == ["main.py"]
 
 
+def test_dynamic_import_plain_string_yields_spec() -> None:
+    assert _extract_import_specs('import("./dev.js");\n', ".js") == {"./dev.js"}
+
+
+def test_dynamic_import_with_options_object_yields_spec() -> None:
+    source = 'import("./dev.js", { with: { type: "json" } });\n'
+    assert _extract_import_specs(source, ".js") == {"./dev.js"}
+
+
+def test_dynamic_import_with_concatenated_suffix_yields_no_spec() -> None:
+    """A computed specifier must not be recorded by its literal prefix.
+
+    ``import("./dev.js" + suffix)`` selects a module that is unknowable
+    statically. Recording ``./dev.js`` invents an edge that is wrong AND
+    hides the real one, so the prefix must not be admitted at all.
+    """
+    assert _extract_import_specs('import("./dev.js" + suffix);\n', ".js") == set()
+
+
+def test_dynamic_import_template_literal_yields_spec() -> None:
+    assert _extract_import_specs("import(`./dev.js`);\n", ".js") == {"./dev.js"}
+
+
+def test_dynamic_import_template_literal_with_concat_yields_no_spec() -> None:
+    """The template-literal branch has the same computed-prefix hole."""
+    assert _extract_import_specs("import(`./dev.js` + suffix);\n", ".js") == set()
+
+
 def test_import_spec_helpers_cover_unsupported_and_unresolved_cases(
     tmp_path: Path,
 ) -> None:
--- a/tests/unit/test_call_graph_integration.py
+++ b/tests/unit/test_call_graph_integration.py
@@ -178,6 +178,39 @@ class TestCallGraphLanguageParity:
         assert names == {"loadData", "processData", "main"}
 
 
+class TestResolveImportPath:
+    """``CallGraph._resolve_import_path`` extension probing.
+
+    The probe table must cover the Node module suffixes the project graph
+    already admits (``.mjs``/``.cjs``/``.mts``/``.cts``), otherwise an
+    admitted ``.mts`` file importing an extensionless sibling silently
+    loses its cross-file call edges.
+    """
+
+    @staticmethod
+    def _probe(source_rel: str, spec: str, target_rel: str) -> str:
+        cg = CallGraph(str(PY_PROJECT))
+        return cg._resolve_import_path(
+            source_rel, spec, True, {target_rel: f"/abs/{target_rel}"}
+        )
+
+    def test_resolves_mts_sibling(self):
+        target = str(Path("src") / "util.mts")
+        assert self._probe(str(Path("src") / "app.mts"), "./util", target) == target
+
+    def test_resolves_mjs_sibling(self):
+        target = str(Path("src") / "util.mjs")
+        assert self._probe(str(Path("src") / "app.mjs"), "./util", target) == target
+
+    def test_resolves_cts_sibling(self):
+        target = str(Path("src") / "util.cts")
+        assert self._probe(str(Path("src") / "app.cts"), "./util", target) == target
+
+    def test_resolves_cjs_sibling(self):
+        target = str(Path("src") / "util.cjs")
+        assert self._probe(str(Path("src") / "app.cjs"), "./util", target) == target
+
+
 class TestCallGraphCallersOf:
     @_WINDOWS_SKIP_PY_FIXTURE
     def test_callers_of_called_function(self):
--- a/tests/unit/test_cross_file_resolver.py
+++ b/tests/unit/test_cross_file_resolver.py
@@ -99,6 +99,56 @@ class TestCrossFileResolverBuild:
         assert "utils" in mod_paths
 
 
+@pytest.fixture
+def mts_project(tmp_path):
+    """A real indexed project whose .mts entry imports an extensionless sibling."""
+    project = tmp_path / "mtsproj"
+    project.mkdir()
+    (project / "src").mkdir()
+
+    (project / "src" / "util.mts").write_text(
+        "export function run(value: number): number {\n    return value + 1;\n}\n"
+    )
+    (project / "src" / "app.mts").write_text(
+        "import { run } from './util';\n"
+        "\n"
+        "export function main(): number {\n"
+        "    return run(1);\n"
+        "}\n"
+    )
+
+    cache = ASTCache(str(project))
+    cache.index_project(max_files=100)
+    yield project, cache
+    cache.close()
+
+
+class TestNodeModuleCrossFileResolution:
+    """PUBLIC API: an ESM import in a .mts file must bind across files.
+
+    Regression for the Codex finding on #1312: ``_register_module_path``
+    registered ``src/util.mts`` correctly, but the LOOKUP in
+    ``_resolve_module_to_file`` probed only ``("", ".py", ".js", ".ts",
+    "/__init__.py")`` — so the relative ``./util`` never matched and the
+    binding was dropped. Registration worked; resolution did not.
+    """
+
+    def test_esm_relative_import_binds_to_mts_sibling(self, mts_project):
+        _project, cache = mts_project
+        resolver = CrossFileResolver(cache)
+        resolver.build()
+        assert resolver._name_to_source["src/app.mts"]["run"] == "src/util.mts"
+
+    def test_resolve_callee_finds_cross_file_mts_definition(self, mts_project):
+        _project, cache = mts_project
+        resolver = CrossFileResolver(cache)
+        resolver.build()
+        files = [
+            path for path, _confidence in resolver.resolve_callee("run", "src/app.mts")
+        ]
+        assert "src/util.mts" in files
+
+
 class TestRegisterNodeModulePaths:
     """``_register_module_path`` must strip the full Node module suffix.
 
--- a/tests/unit/test_cross_file_resolver.py
+++ b/tests/unit/test_cross_file_resolver.py
@@ -99,6 +99,44 @@ class TestCrossFileResolverBuild:
         assert "utils" in mod_paths
 
 
+class TestRegisterNodeModulePaths:
+    """``_register_module_path`` must strip the full Node module suffix.
+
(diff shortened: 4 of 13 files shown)


aimasteracc:
@codex review

aimasteracc:
## Codex review triage — both findings CONFIRMED, both fixed in bd168bc2

| # | Codex comment | Verdict | Action |
|---|---|---|---|
| 1 | `cross_file_resolver.py:440` — `for ext in ("", ".py", ".js", ".ts", "/__init__.py")` means `./util` from `src/app.mts` never resolves | **CONFIRMED — real problem** | Fixed. Probe list now selected by the importing file's language (`_relative_probe_exts`). Kept as **two disjoint lists** as requested: a Python file must not probe `.mts`, a `.mts` file must not probe `__init__.py`. |
| 2 | `import_graph.py:334` — prefix dispatch sends ESM `import ... from` to the **Python** resolver, so the JS suffix table is unreachable | **CONFIRMED — real problem, and the more serious one** | Fixed. Dispatch is now on the **source file's language** via `EXT_TO_LANG`, not on statement text. `import ` is ambiguous between Python and ESM and that ambiguity was the root cause. The old text sniff survives only as a fallback for extensions absent from `EXT_TO_LANG`. |

You were right that the fix did not work end-to-end, and right about why.

## Finding 3 — discovered *because* the E2E test was written honestly

Writing the public-API test surfaced a third, **pre-existing and larger** bug: the AST cache keys `file_path` POSIX-style, but both resolvers built candidates with `os.path.join`/`normpath` — backslashes on Windows. The membership test never matched, so **`ImportGraph` produced ZERO edges for any project with a subdirectory — in Python as well as JS/TS.** Linux CI cannot see it because `normpath` is already POSIX there.

Measured on a real index, before → after:

| fixture | before | after |
|---|---|---|
| nested `.mts` ESM | 0 edges | 1 |
| nested `.cjs` require | 0 | 1 |
| nested `.js` ESM | 0 | 1 |
| nested Python | 0 | 1 |

Fixed via `_file_lookup`/`_match_file`, which compare separator-agnostically and return the **original** cache key so edges still reference real paths.

## On the audit method — the correction I owed

My previous report said *"no additional incomplete suffix tables; audited every `.mjs/.cjs/.mts/.cts` site."* That method was structurally incapable of finding finding 1: **grepping for the suffixes you are adding can only find the tables that already have them.** The tables that need fixing are exactly the ones that do not contain the search term. Noted and internalised.

Re-audited by **shape** instead: every tuple/list/set literal containing ≥2 dotted extensions, across the whole package — 26 JS/TS-family hits. Full table with a verdict each:

**Needed the suffixes → fixed in this PR (4)**

| Site | Note |
|---|---|
| `cross_file_resolver.py:440` | finding 1 |
| `import_graph.py:265` + dispatch at `:334` | finding 2 |
| `nav_facade.py:160` | listed `.mjs`/`.cjs` but **not** `.mts`/`.cts` — genuinely incomplete, now uses the shared constant |
| `constants.py:212` | the new shared source of truth |

**Correctly does NOT need them (by design) (7)**

| Site | Why |
|---|---|
| `javascript_plugin/plugin.py:29` | JS plugin correctly owns `.mjs`/`.cjs`; `.mts`/`.cts` belong to the TS plugin |
| `typescript_plugin/plugin.py:41` | mirror of the above — correct partition, not a gap |
| `typescript_plugin/extractor.py:217` | JSX detection; `.mts` cannot contain JSX |
| `safe_to_edit_helpers.py:842` | `{".ts",".tsx",".mts",".cts"}` — deliberately TS-only |
| `safe_to_edit_helpers.py:1965` | `{".js",".jsx",".mjs",".cjs"}` under `language=="typescript"` — deliberately the allowJs set |
| `safe_to_edit_helpers.py:704-707` | `_TYPESCRIPT_SOURCE_SUBSTITUTIONS` — keyed *by* suffix; `.mjs`/`.cjs` have their own rows. Complete as a dict; my line-granularity detector split it. |
| `safe_to_edit_helpers.py:917/920` | those lines are the *values* of the `.d.ts`/`.tsx` rows in a table that already carries `.d.mts`/`.d.cts`/`.mts`/`.cts`. Complete as a whole. |

**Genuinely incomplete, but a DIFFERENT cluster → deliberately NOT fixed here (8)**

`middleware_detector.py:55`, `test_gap_analyzer.py:82`, `query_symbol_search.py:26`, `symbol_lineage_tool.py:1014`, `trace_impact_tool.py:75`, `_codegraph_explore_helpers.py:24`, `gitignore_detector.py:226`, `_typescript_formatter_signatures_table.py:93` (this last one mangles `foo.mts` → `foo.m` for display only).

These are all "which files count as source?" inventory sets, not import resolvers. Fixing them here would turn a focused resolver PR into a 14-file cross-cutting diff, which the repo's PR-hygiene rule closes rather than merges. **Recommend one focused follow-up PR for the inventory cluster.** Flagging rather than silently skipping.

Two caveats worth stating plainly: my detector over-reports at line granularity (see the two `safe_to_edit_helpers` false positives above), so each hit needed a human verdict; and the 8 deferred sites are a real, currently-shipping gap, not a theoretical one.

## Tests — public API, real index

8 new tests, 6 RED on the parent commit. The load-bearing ones drive `ImportGraph.build()` and `CrossFileResolver.build()` against a **real `ASTCache` index**, no fake cache and no path massaging.

The earlier fake-cache tests passed while the feature was broken because I had normalised the fixture keys with `_np()` — matching the fixture to the implementation instead of to production. That is what hid the separator bug; the keys are now POSIX, as the real cache emits them.

## Verification

- `uv run pytest <6 files emitted by --change-impact> -q` → **184 passed, 3 skipped**
- `uv run ruff check .` → **All checks passed!**
- `uv run mypy tree_sitter_analyzer/` → **23 errors** = the baseline measured by stashing (not 24)
- `uv run pre-commit run --files <the 6 changed>` → **every hook Passed**
- `uv run pytest -q` → 14 failures, the **same 14 files** as the base commit (`test_collect_no1_006b_baseline`, `test_fresh_install_qualification`, `test_outdated_uv_qualification`, `test_ast_cache`, `test_codegraph_context_tool`) — pre-existing, unrelated
- E2E dogfood re-run: `src/app.mts → src/util.mts` and `src/legacy.cjs → src/helper.cjs` both present; `CrossFileResolver` binds `run → src/util.mts`

aimasteracc:
## Triage addendum — independent corroboration of Finding 3, and PR retitled

### Finding 3 independently reproduced by the reviewer

Confirmed on clean `develop` with a fixture built separately from mine:

```
fixture: src/app.py imports src/util.py (nested) + flat_a.py imports flat_b.py (root)

develop:     edge_count=1   FLAT resolves, NESTED DOES NOT
with #1312:  edge_count=2   both resolve
```

Two independent fixtures, two independent reproductions, same conclusion. This corroborates that the bug is **language-independent** — the reviewer's fixture is pure Python, mine was `.mts`/`.cjs`/`.js`/Python — and that it is a **pre-existing** defect on `develop`, not something this PR introduced.

Restating the blast radius plainly, since it is larger than my original wording conveyed: on Windows, `ImportGraph` drops **every** import edge that crosses a subdirectory. Because essentially every real project has subdirectories, **the import graph has been effectively empty for Windows users**, and `health action=imports`, cycle detection, and cross-file resolution all inherited it. Linux CI cannot observe this because POSIX paths have no separator to mismatch.

### PR retitled and body restructured

The old title — `fix: admit Node module suffixes across graph fingerprint and resolvers` — no longer described the most important change in the diff, which is the CLAUDE.md "misleading title" problem. A reviewer skimming it would not know a Windows-wide correctness bug was inside.

New title: **`fix: resolve import edges across directories on Windows, and admit Node module suffixes`**

Body restructured so the separator bug leads as §1 with its own before/after table, one-sentence root cause, and an explicit blast-radius statement including that it was pre-existing. The Node-suffix work is §2, and the shared-constant refactor is §3 (five hand-maintained copies collapsed into one derived constant — `JS_TS_INDEX_SUFFIXES` and `JS_TS_MODULE_EXTS_LONGEST_FIRST` are both *derived* from `JS_TS_MODULE_EXTS`, so they cannot drift). Added a **"How this was missed"** section explaining that the pre-existing tests exercised the private resolver helpers directly — proving the helpers worked while nothing proved the public path did — and naming the public-API tests added so the next person understands why they exist.

### Corrections to my own earlier numbers

Measured rather than asserted: **30** new tests across the two commits, not 31. **26** fail on the relevant parent commit (20 of the 23 from commit 1, 6 of the 7 from commit 2). The 4 that pass on both are deliberate guards — they pin that the tightened dynamic-import regex still accepts legitimate specifiers, and that language-based dispatch did not break Python resolution. 13 files changed.

### Confirmed scope boundary

The 8 remaining incomplete tables (`middleware_detector`, `test_gap_analyzer`, `query_symbol_search`, `symbol_lineage_tool`, `trace_impact_tool`, `_codegraph_explore_helpers`, `gitignore_detector`, `_typescript_formatter_signatures_table`) are **not** pulled into this PR, per the reviewer's instruction — they are tracked as a separate focused follow-up.

No code changes in this update; documentation only. HEAD remains `bd168bc2`.

PR #1313, opened 2026-08-19
docs: stop TESTING.md from teaching what its own rules forbid

aimasteracc:
## Summary

`docs/TESTING.md` is the canonical testing reference — bound by three contract tests and the document every agent and contributor is pointed at. It **defines** the weak-assertion rule at lines 105–126 and then **violates that same rule** three times further down, and it recommends a filename that a pre-commit hook rejects outright.

Seven fixes.

### Banned test-file names — rule T-1 is a BLOCKER

| line | was | now |
|---|---|---|
| 163 | recommended `test_<module>_comprehensive.py` as a **primary option** | the one-file-per-module rule, plus an explicit list of all five banned patterns |
| 388 / 398 / 401 | example commands invoking `tests/unit/test_exceptions_comprehensive.py` | `tests/unit/test_exceptions_core.py` with a real class and a real method |

The example commands were doubly bad: they modelled the banned name **and** named a file that does not exist. The deepest of the three now runs and passes exactly as written:

```
$ uv run pytest "tests/unit/test_exceptions_core.py::TestTreeSitterAnalyzerError::test_message_and_error_code"
1 passed in 6.74s
```

### Weak assertions — LOCKED exact-assertion rule (user-locked 2026-06-10)

| line | was | now |
|---|---|---|
| 642 | `assert len(result["elements"]["functions"]) >= 1` | `== 2` |
| 492 | bare `assert result is not None` (fixture example) | `assert result.language == "python"` |
| 555 | bare `assert result is not None` (async example) | `assert result.status == "ok"` |

The last two are the sharpest case: **line 109 of this same document** says a placeholder existence check is rejected "unless the same test also asserts concrete behavior" — and then lines 492 and 555 present exactly that unpaired form as good practice.

### Deliberately NOT changed

The `>= 1` / `is not None` occurrences at lines 108–111 are **negative examples** inside the "the ratchet rejects" list, and the one at line 125 is deliberately paired with a behavioral assertion to illustrate the *allowed* form. All four are correct as written.

## Verification

- The doc's own deepest example command runs and passes (output above).
- `tests/contracts` + `tests/governance` + `tests/integration/docs`: **11 failed, 617 passed, 48 skipped** — **identical with and without this change**, measured by stashing. This PR adds zero failures.
- Full pre-commit hook set passed, including `block-banned-test-names (T-1)` and `Weak Assertion Ratchet`.

## Note for maintainers

Those 11 contract failures are **pre-existing on `develop`** (`test_fresh_install_qualification`, `test_collect_no1_006b_baseline`). They are unrelated to this PR but worth a separate look — alongside the 23 `mypy` errors and the ~40 `tests/unit` failures that are also currently baseline on develop. A green-looking CI with a standing red baseline erodes the signal these gates exist to provide.

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -160,7 +160,11 @@ uv run pytest tests/benchmarks/ -m benchmark --benchmark-enable --benchmark-only
 
 ### Test File Naming
 
-- Unit test files: `test_<module>_comprehensive.py` or `test_<module>.py`
+- Unit test files: `test_<module>.py` — one file per module.
+  Do NOT create a second file for an existing module. Rule T-1 is a
+  BLOCKER and the `block-banned-test-names` hook rejects the
+  `*_comprehensive*`, `*_edge_cases*`, `*_coverage*`, `*_extended*`, and
+  `*_optimized*` name patterns outright. Add to the existing file instead.
 - Integration tests: `test_<feature>_integration.py`
 - End-to-end tests: `test_<workflow>_e2e.py`
 
@@ -385,7 +389,7 @@ uv run pytest tests/ --cov=tree_sitter_analyzer --cov-report=term-missing
 
 ```bash
 # Single file
-uv run pytest tests/unit/test_exceptions_comprehensive.py
+uv run pytest tests/unit/test_exceptions_core.py
 
 # Multiple files
 uv run pytest tests/unit/test_*.py
@@ -395,10 +399,10 @@ uv run pytest tests/unit/test_*.py
 
 ```bash
 # Specific class
-uv run pytest tests/unit/test_exceptions_comprehensive.py::TestAnalysisError
+uv run pytest tests/unit/test_exceptions_core.py::TestTreeSitterAnalyzerError
 
 # Specific test
-uv run pytest tests/unit/test_exceptions_comprehensive.py::TestAnalysisError::test_initialization_with_all_parameters
+uv run pytest tests/unit/test_exceptions_core.py::TestTreeSitterAnalyzerError::test_message_and_error_code
 ```
 
 ### Run Tests by Marker
@@ -489,7 +493,7 @@ def mock_analyzer():
 def test_with_fixtures(sample_code, mock_analyzer):
     """Test using fixtures."""
     result = mock_analyzer.analyze(sample_code)
-    assert result is not None
+    assert result.language == "python"
 ```
 
 ### 3. Test Error Conditions
@@ -552,7 +556,7 @@ import pytest
 async def test_async_function():
     """Test async function."""
     result = await async_function()
-    assert result is not None
+    assert result.status == "ok"
 ```
 
 ### 7. Clean Up Resources
@@ -639,7 +643,7 @@ def test_python_function_analysis(tmp_path):
         expected_language="python",
         require_success=True
     )
-    assert len(result["elements"]["functions"]) >= 1
+    assert len(result["elements"]["functions"]) == 2
 ```
 
 ---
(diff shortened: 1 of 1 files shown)


aimasteracc:
@codex review

aimasteracc:
## Substitute adversarial review — triage. Three of my own fixes were wrong.

Codex is unavailable (account-level rate limit: `You have reached your Codex usage limits for code reviews` on every open PR here), so a substitute adversarial reviewer stood in. It found that three of the seven edits in the first commit were defective — **including the exact failure mode this PR was written to eliminate.** All fixed in `02015045`.

| # | finding | verdict | fix |
|---|---|---|---|
| 1 | `:646` `== 2` is exactly wrong | **CONFIRMED — my error** | Correct value is **1**. I replaced a loose `>= 1` with a pinned count *without running the example*. `generate_python_function(name="test_func", params=["x","y"], body="return x + y")` emits one function. An agent copying this template goes red and "fixes" it by relaxing back to `>= 1`, now with a LOCKED-rule citation as cover. **Worse than the bound I removed.** |
| 2 | `:493-497` the new assertion makes the example *fail* | **CONFIRMED — my error, twice** | `Mock(spec=...)` returns a fresh unconfigured Mock per call, so `result.language == "python"` compared two Mocks. The spec'd class `CodeAnalyzer` also **does not exist** in this codebase. My *first* correction was also wrong: I spec'd the real `UnifiedAnalysisEngine`, whose `analyze` is `async def`, so autospec yielded an `AsyncMock` returning a coroutine → `AttributeError: 'coroutine' object has no attribute 'language'`. Now a configured plain `Mock`, with both traps documented — I fell into both, so they are worth teaching. |
| 3 | `:163-167` overstates the hook | **CONFIRMED — my error** | I wrote that `block-banned-test-names` rejects the patterns "outright". False in both directions, both demonstrated: it inspects only **added** files (`--diff-filter=A`), so `git mv` into a banned name passes; and it matches the whole **path**, so a legitimate file under `tests/unit/grammar_coverage/` trips it (23 tracked paths match a banned pattern today). Rewritten as a screen with two named limits, stating T-1 is the author's obligation, not the hook's. |
| 4 | `:559` `result.status` is fictional | **CONFIRMED** | No model in this package has a `status` attribute. Aligned with the canonical mapping form at line 116. |
| 5 | `:20-22` coverage-as-top-priority | **CONFIRMED — missed in the first pass** | The Overview listed ">80% overall coverage, with critical modules at >85%" as the **first** testing priority — directly against LOCKED rule **T-3** ("coverage rate is a by-product, never a goal"), and stated as loose bounds inside the document that bans them. Replaced with the behavioral-assertion-first principle plus an explicit T-3 citation. |
| 6 | file carries only 2 of 5 locked test rules | **VALID, deferred** | T-3 is now cited; T-4 and T-5 still appear nowhere. Out of scope for a defect-fix PR — filed as follow-up rather than expanded here. |
| — | the "deliberately NOT changed" judgement on lines 108-125 | **CONFIRMED CORRECT** | Reviewer independently verified all four are legitimate negative examples inside the "the ratchet rejects" list, or the deliberately-paired allowed form at 125-126. |

### Verification this time — the check that was missing the first time

Both corrected examples were **executed verbatim as real pytest tests**: `2 passed`. The generator's one-function output was **measured**, not assumed. Contract tests binding this file: **34 passed**. Full pre-commit hook set green.

### Root cause, recorded because it will recur

**Nothing in CI validates the examples in this file.** The weak-assertion ratchet parses staged test `.py` via `ast` and cannot see markdown. `test_doc_cli_examples.py` extracts only `tree-sitter-analyzer` CLI flags. The other two contracts check one substring each.

That is why an example naming a file which **both did not exist and whose name a hook bans** sat in the canonical testing document unnoticed — and it is why this fix is unprotected and can silently regress. Same shape as the 2026-06-08 TOON incident: conformance tests verify the spec and cannot notice the spec is wrong.

PR #1314, opened 2026-08-19
fix: repair both dead detectors in the codemap-sync gate

aimasteracc:
## Problem

`scripts/codemap-sync-check.sh` (pre-commit hook `tsa-codemap-sync`) blocks commits that change the MCP or CLI surface without updating the matching CODEMAP. **Both detectors were dead** — they matched nothing, so the gate had silently stopped gating.

| Detector | Was looking for | Reality |
|---|---|---|
| MCP registry | `("name", …Tool(` | Post-facade-cutover the registry emits `("search", build_search_facade(...))` — 0 matches |
| CLI arguments | `add_argument(` in `cli/argument_parser_builder.py` | That file has **0** `add_argument` calls; all 323 live in `cli/argument_groups/_*.py` |

Nobody noticed because `tests/integration/test_codemap_sync_hook.sh` built a **synthetic old-shape registry** as its fixture. The test was protecting the bug.

## Fix

1. **RED first.** Rebuilt the test to add a flag in `argument_groups/_*.py` and a facade registration in `_tool_registry.py` without touching the codemaps; both asserted exit 1 and both returned exit 0 against the unfixed hook.
2. **Fixture rebuilt off reality.** `init_repo()` now copies the *real* registry, the *real* `argument_groups/*.py`, the *real* `argument_parser_builder.py` and the *real* `docs/CODEMAPS/*.md` into the temp repo, and derives its mutation lines by cloning a real matching line out of them. No hand-written imitation of production shape remains.
3. **New `--self-check` mode.** `bash scripts/codemap-sync-check.sh --self-check` applies each detector's real regex to the live production files and exits non-zero if any matches nothing. Test 0 asserts it (and requires the `self-check OK` marker, so an ignored-unknown-arg no-op cannot pass silently). This is the mechanism that makes the trap non-resettable.
4. **Detectors repointed** to the semantic shapes (see below).
5. **Wired into CI.** The shell test was never run by any workflow. Added a `codemap-sync-gate` job to `reusable-quality.yml`.

### What each detector became

- **MCP:** `\("[a-zA-Z_][a-zA-Z0-9_]*"[[:space:]]*,[[:space:]]*[A-Za-z_][A-Za-z0-9_]*\(` — a quoted tool name mapped to a *constructed object*. Semantic, not literal: matches the facade form, the legacy `…Tool(` form, and whatever the next factory naming convention is.
- **CLI:** path regex `^tree_sitter_analyzer/cli/(argument_parser_builder\.py|argument_groups/[^/]+\.py)$` x content regex `add_argument\(` — a *defined argument* anywhere on the CLI argument surface, not one hard-coded file.

## CODEMAP re-sync (drift accumulated while the gate was down)

| File | Old | New | Source command |
|---|---|---|---|
| `docs/CODEMAPS/cli.md` | 295 unique flags | **324** | `create_argument_parser()` option-string set (CLAUDE.md) |
| `docs/CODEMAPS/architecture.md` | 21 tree-sitter plugins | **22** | `--show-supported-languages` -> `language_count` |
| `docs/CODEMAPS/architecture.md` | ~74 inner tool classes | **146 modules / 79 inner tool classes** | `ls mcp/tools/*.py \| wc -l`; `grep -rhoE '^class [A-Za-z_0-9]+Tool[(:]' mcp/tools/*.py \| sort -u \| wc -l` |

Checked and found **not** drifted: `mcp-tools.md` (8 facades, 66 legacy — `len(create_tool_registry('.')[0])` = 8, `len(LEGACY_TOOL_MAP)` = 66), `languages.md` (22), `README.md` (already 324, kept honest by the release-gate test), `formatters.md`, `security.md`.

## AGENTS.md

The "self-enforcing" claim was false while the detectors were dead; after this fix it is true again, so the sentence stands — but two things were corrected: the registry table row pointed at `argument_parser_builder.py` (wrong file), and the enforcement section now names `--self-check` as the way to prove the claim rather than asserting it as prose.

## Verification

```
$ bash tests/integration/test_codemap_sync_hook.sh        # 11 PASS / 0 FAIL
$ bash scripts/codemap-sync-check.sh --self-check
[codemap-sync] detector 'mcp-registry': 8 matching line(s)
[codemap-sync] detector 'cli-arguments': 323 matching line(s)
[codemap-sync] self-check OK
```

Hook run on a synthetic MCP + CLI surface change with no codemaps staged -> exit 1, two BLOCK lines. Hook run on a docs-only change -> exit 0 (no false positive).

`uv run ruff check .` -> All checks passed. `uv run pre-commit run --files <changed>` -> all green. `--show-supported-languages` (22) and `--show-supported-extensions` (52) both sane. `uv run pytest -q` (the `--change-impact` verification command) -> 14 pre-existing failures, byte-identical on clean `origin/develop`; unrelated to this diff, which touches no Python.

## Note for reviewers

This does **not** newly block a large number of commits — it only blocks commits that were always supposed to be blocked. The gate has no retroactive effect; it fires on staged diffs only.

### DIFF
--- a/.claude/skills/tsa-edit-then-verify/SKILL.md
+++ b/.claude/skills/tsa-edit-then-verify/SKILL.md
@@ -227,7 +227,7 @@ bash scripts/codemap-sync-check.sh
 echo "exit=$?"  # 0 = pass, 1 = block
 ```
 
-Escape hatch (intentional rename without semantic change): `SKIP_CODEMAP_SYNC=1 git commit ...`. The pytest `test_registered_mcp_tools_have_codemap_parity` is the safety net that catches abuse in CI.
+Escape hatch: `SKIP_CODEMAP_SYNC=force git commit ...` (bypasses and logs to `$GIT_DIR/codemap-sync-bypass.log`; the older `=1` now fails rather than passing silently, because pre-commit hides output from passing hooks). The pytest nets that catch abuse in CI are `test_registered_mcp_tools_have_codemap_parity` for `mcp-tools.md` and `test_cli_codemap_flag_count_matches_the_real_parser` for `cli.md`.
 
 ### Why Phase D exists
 
--- a/.github/workflows/reusable-quality.yml
+++ b/.github/workflows/reusable-quality.yml
@@ -99,7 +99,19 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          version: "latest"
+      - name: Set up Python 3.11
+        run: uv python install 3.11
       - name: Run codemap-sync-check integration test
+        # --self-check compares the gate's static extractor against the
+        # authoritative runtime enumerations, so the project env must be present.
         run: |
+          uv sync --all-extras
+          bash scripts/codemap-sync-check.sh --self-check
           bash tests/integration/test_codemap_sync_hook.sh
           echo "OK: codemap-sync detectors still match the live registry/CLI surface" >> $GITHUB_STEP_SUMMARY
+      - name: Codemap parity contracts
+        run: uv run pytest -q tests/contracts/test_agent_docs_contract.py
--- a/.github/workflows/reusable-quality.yml
+++ b/.github/workflows/reusable-quality.yml
@@ -93,3 +93,13 @@ jobs:
         run: |
           bash scripts/check_loose_assertions.sh origin/${{ github.base_ref }}
         continue-on-error: false
+
+  codemap-sync-gate:
+    name: Codemap Sync Gate (detectors alive)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Run codemap-sync-check integration test
+        run: |
+          bash tests/integration/test_codemap_sync_hook.sh
+          echo "OK: codemap-sync detectors still match the live registry/CLI surface" >> $GITHUB_STEP_SUMMARY
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -60,28 +60,50 @@ Memory records should capture reusable lessons, not logs: benchmark surprises, C
 
 Any change touching one of these registries MUST update the corresponding `docs/CODEMAPS/*.md` in the **same commit**:
 
-| Registry file | Codemap |
+| Surface | Codemap |
 |---|---|
-| `tree_sitter_analyzer/mcp/_tool_registry.py` | `docs/CODEMAPS/mcp-tools.md` |
-| `tree_sitter_analyzer/cli/argument_groups/*.py` + `tree_sitter_analyzer/cli/argument_parser_builder.py` | `docs/CODEMAPS/cli.md` |
+| `tree_sitter_analyzer/mcp/_tool_registry.py` — the registered tool-name set | `docs/CODEMAPS/mcp-tools.md` |
+| `tree_sitter_analyzer/cli/**/*.py` — the whole `add_argument` flag set, including the `find-and-grep` / `list-files` / `search-content` console scripts | `docs/CODEMAPS/cli.md` |
 | `tree_sitter_analyzer/languages/<lang>_plugin/*` | `docs/CODEMAPS/languages.md` |
 | `tree_sitter_analyzer/formatters/*` | `docs/CODEMAPS/formatters.md` |
 
+The mandate is defined on the **surface set**, not on a file list: the gate compares
+the set of registered tool names and the set of CLI flags at `HEAD` against the
+staged index, and fires only when a set actually changed. Reordering, renames,
+comments and docstrings therefore do not trigger it, and a *removal* does.
+
 Enforced by:
 - `scripts/codemap-sync-check.sh` (pre-commit hook + Claude PreToolUse soft-nag)
-- `test_registered_mcp_tools_have_codemap_parity` in `tests/contracts/test_mcp_surface_metadata_contract.py`
+- `test_registered_mcp_tools_have_codemap_parity` in `tests/contracts/test_mcp_surface_metadata_contract.py` — the CI net for `mcp-tools.md`
+- `test_cli_codemap_flag_count_matches_the_real_parser` in `tests/contracts/test_agent_docs_contract.py` — the CI net for `cli.md`
+- `test_codemap_sync_gate_sees_every_registered_mcp_tool` / `..._sees_every_cli_flag` /
+  `..._watches_the_whole_cli_flag_surface` in the same file — the CI net for **the gate
+  itself**, asserting exact set equality against the authoritative runtime
+  enumerations plus zero unwatched flags
 
-"Self-enforcing" is only true while the hook's detectors still match the code they
-watch. Both had gone dead against the current shapes (the MCP detector still looked
-for the pre-facade `("name", SomeTool(` form; the CLI detector still watched
-`argument_parser_builder.py`, which holds zero `add_argument` calls). Verify with
-`bash scripts/codemap-sync-check.sh --self-check`, which fails when a detector
-matches nothing; `tests/integration/test_codemap_sync_hook.sh` runs it and builds
-its fixtures from the real files rather than imitations of them.
+"Self-enforcing" is a claim about the detectors, so it has to be measured, not
+asserted. Run `bash scripts/codemap-sync-check.sh --self-check`: it fails unless the
+gate's static extractor reproduces the authoritative enumerations *exactly* and no
+`add_argument` flag under `cli/**` falls outside the watch filter. A `count > 0`
+check is not sufficient — a tree whose only match is a stale docstring mention
+passes it while the detector is dead, and a loose lower bound on a deterministic
+count is what CLAUDE.md's exact-assertion rule forbids in the first place.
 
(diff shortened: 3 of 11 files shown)


aimasteracc:
@codex review

aimasteracc:
## Adversarial review response — all 3 P1s, all 3 P2s, all 5 P3s addressed

The reviewer was right that the regex repair was the wrong shape of fix. Both the misses and the false blocks had one root cause — **pattern-matching added diff lines** — so the gate has been rebuilt around **surface-set comparison**, which is what the structural fix demanded. New helper: `scripts/codemap_surface.py`.

### The structural fix (P2-1 / P2-2 / P2-4 / P3-3, one root cause)

The gate now enumerates the **actual set** of registered tool names and the **actual set** of CLI flags at `HEAD` and in the index, and fires only when a set genuinely changed.

- **Enumeration is static `ast`, not an import.** Measured: importing `cli_main` to build the real parser costs **638 ms**; the whole static comparison is **one** Python process (~2.2 s worst case on Windows, ~0 on the vast majority of commits because a cheap path pre-filter runs first). This is the split the reviewer pre-authorised: cheap exact-set comparison in pre-commit, authoritative runtime enumeration in CI.
- **The static extractor is not trusted on faith.** `--self-check` and four new CI tests assert **exact set equality** between static extraction and the authoritative `CLAUDE.md` enumerations. Measured agreement: static `argument_groups` = **323** long flags, runtime = **324**; `runtime − static` = exactly `{--help, -h}` (argparse-synthesised), `static − runtime` = **∅**.
- Comments, docstrings, reordering, parameter renames, dict literals, `append`-in-a-loop and **removals** are all handled by construction, because none of them are line patterns any more.

### P1-1 — killed structurally, not patched

The gate **no longer parses git diff text at all**. Staged paths come from `git diff --cached --name-only -z`, so `diff.noprefix`, `diff.mnemonicPrefix`, `diff.external` and `core.quotePath` are all irrelevant. Regression tests 12/13 set `GIT_CONFIG_GLOBAL` to configs with `noprefix`/`mnemonicPrefix` and assert the gate still blocks.

While implementing this I found a **second, worse instance of the same class of bug in my own first attempt**: `STAGED_RAW="$(git diff ... -z)"` — bash **silently drops NUL bytes inside `$(...)`**, which concatenated every staged path into one meaningless string and made the gate see nothing at all. It is now read via process substitution, with a comment explaining why. That one would have been a total gate death on every platform.

### P1-2 — exact counts, plus the stronger coverage invariant

`--self-check` no longer asserts `count > 0`. It now asserts:

1. `static_mcp_names == {name for name, _ in create_tool_registry('.')[0]}` — exact set equality
2. `runtime_parser_flags − static_flags == {-h, --help}` — exact, nothing hand-waved
3. **coverage:** `add_argument` flags under `tree_sitter_analyzer/cli/**` outside the watch filter **== 0**
4. watched paths non-empty (a rename that empties the filter is a dead detector even if 1–3 agree)

The reviewer is right that this was the rule the guard was written to enforce being violated by the guard itself. Fixed.

### P1-3 — detector widened to the mandate, not mandate narrowed to the detector

The watch filter is now **all of `tree_sitter_analyzer/cli/**/*.py`**, so all 405 `add_argument` calls are covered, including the 82 in `find_and_grep_cli*.py` / `list_files_cli.py` / `search_content_cli*.py`. Coverage invariant #3 above now reads **0 unwatched** and fails if that ever regresses.

On the AGENTS.md direction-of-fit criticism — accepted, that was the wrong call and it is reversed. The table row now reads `tree_sitter_analyzer/cli/**/*.py` and describes the mandate as defined on the **surface set**, not on a file list. Test 14 covers a flag added in `list_files_cli.py`.

The PR body's "anywhere on the CLI argument surface" claim was wrong for 20% of the surface when written. It is now true, and measured by an executable invariant rather than asserted.

### P2-3 — bypass made visible, and the missing CI net built

Two parts, because local visibility is genuinely limited by pre-commit (it only surfaces output from *failing* hooks — there is no warn channel):

- **`SKIP_CODEMAP_SYNC=1` now fails** when it would have silenced a real violation, printing why and pointing at `=force`. This is the part that matters: `export SKIP_CODEMAP_SYNC=1` can no longer disable the gate for a session with zero signal, because the documented value now produces a visible failure.
- **`SKIP_CODEMAP_SYNC=force`** bypasses and appends an audit line to `$GIT_DIR/codemap-sync-bypass.log`.
- **The CI net for `cli.md` now exists** — `test_cli_codemap_flag_count_matches_the_real_parser`. `AGENTS.md:82`'s "the pytest test still runs in CI as the final safety net" was true only for `mcp-tools.md`; it is now true for both, so a CLI-side bypass is recoverable.

### P3 fixes

| | Fix |
|---|---|
| P3-1 | `docs/CODEMAPS/cli.md:168` no longer says the gate watches `argument_parser_builder.py`; it describes the flag-surface set comparison |
| P3-2 | `docs/architecture.md:178` "approximately 74" → **79 classes across 146 modules**. This file was missing from my "checked, not drifted" list — the sweep was incomplete, now re-run across `docs/`, `AGENTS.md`, `README.md` with zero remaining hits |
| P3-4 | Structurally fixed by `-z` (git never quotes NUL-separated paths). Test 21 stages `_café.py` |
| P3-5 | Unknown args now exit **2** with a usage message; `--help` added. Test 22 covers the `--selfcheck` typo |
| P3-7 | Test fixture derivation uses `[[:space:]]`, not GNU-only `\s` |

Also updated both copies of `tsa-edit-then-verify/SKILL.md`, which documented the old `=1` escape hatch.

### AGENTS.md wording

Took the "earn it" option. The claim is no longer unqualified: it now says self-enforcement is a claim about the detectors that **has to be measured**, names `--self-check` and the four CI tests as the measurement, explains why `count > 0` is insufficient, and ends with *"self-enforcing for as long as `--self-check` and the CI nets above stay green, and no longer than that."* The paragraph also records the actual failure — including that the fixture was a synthetic copy of the old shape.

### Verification

All 23 scenarios, including every one the reviewer reproduced:

```
PASS  0: hook --self-check against real production files  (exit=0)
PASS  1: registry facade add, no codemap  (exit=1)
PASS  2: registry facade add + codemap  (exit=0)
PASS  3: registry comment-only change  (exit=0)
PASS  4: CLI flag add in argument_groups, no codemap  (exit=1)
PASS  5: CLI flag add in argument_groups + codemap  (exit=0)
PASS  6: CLI flag add in argument_parser_builder, no codemap  (exit=1)
PASS  7: docs-only change  (exit=0)
PASS  8: new language plugin, no codemap  (exit=1)
PASS  9: new formatter, no codemap  (exit=1)
PASS 10: SKIP_CODEMAP_SYNC=1 fails visibly instead of silently passing  (exit=1)
PASS 11: SKIP_CODEMAP_SYNC=force bypasses and logs an audit line  (exit=0)
PASS 12: diff.noprefix=true still blocks  (exit=1)
PASS 13: diff.mnemonicPrefix=true still blocks  (exit=1)
PASS 14: CLI flag add in a standalone console script, no codemap  (exit=1)
PASS 15: registry parameter rename does not block  (exit=0)
PASS 16: reordering registrations does not block  (exit=0)
PASS 17: comment mentioning add_argument( does not block  (exit=0)
PASS 18: dict-literal registration, no codemap  (exit=1)
PASS 19: append-in-a-loop registration, no codemap  (exit=1)
PASS 20: registry removal, no codemap  (exit=1)
PASS 21: non-ASCII staged path still blocks  (exit=1)
PASS 22: unknown argument is rejected  (exit=2)
PASS: 23    FAIL: 0    TOTAL: 23
```

```
$ bash scripts/codemap-sync-check.sh --self-check
[codemap-sync] mcp surface: 8 static / 8 runtime
[codemap-sync] cli surface: 360 static / 325 runtime (+2 argparse-implicit)
[codemap-sync] cli coverage: 0 flag(s) outside the watch filter
[codemap-sync] self-check OK
```

- `uv run ruff check .` → **All checks passed!**
- `uv run mypy tree_sitter_analyzer/` → **23 errors in 7 files**; measured by stashing on the clean tree → **identical 23**. `scripts/codemap_surface.py` type-checks clean under the pre-commit hook's flags.
- `uv run pytest -q tests/contracts/test_agent_docs_contract.py` → **9 passed** (5 pre-existing + 4 new)
- `uv run pre-commit run --files <changed>` → all green

### On my earlier "no regressions" claim — the reviewer is right and I withdraw the framing

Bare `uv run pytest -q` in this repo collects **2,030 tests**; CI runs **25,491**. My earlier claim rested on an **~8% sample**, and I should have said so. Corrected statement of what is actually covered: bare `pytest -q` (2,030 selected) shows 14 failures, byte-identical on clean `origin/develop`; `mypy` is at the 23-error baseline; `ruff` is clean; the 4 new contract tests and the 23 hook scenarios pass. The full 25,491-test CI matrix on this PR is the only thing that can support a real no-regression claim, and I will report its result rather than infer it.

### One thing I did not do

The coverage invariant is scoped to `tree_sitter_analyzer/cli/**`, as specified. Flags defined in seven modules **outside** `cli/` (`mcp/server.py`, `task_harness.py`, `platform_compat/*`, `grammar_coverage/grammar_snapshot.py`, `miswire_audit.py`) are internal/utility argparse surfaces, not documented CLI entry points, and remain unwatched. `test_codemap_sync_gate_sees_every_cli_flag` would catch any of them leaking into the *main* parser, but a brand-new console script placed outside `cli/` would not be caught. Flagging rather than silently widening scope.

PR #1316, opened 2026-08-19
docs(rfc): RFC-0028 measuring the properties we already claim

aimasteracc:
## Summary

One working session on 2026-08-19 found **six defects that shared a single shape: the absence of a signal was read as a positive signal.**

| # | defect | why it looked fine |
|---|---|---|
| 1 | the codemap-sync gate's two detectors matched nothing | hook exited 0 → read as "no violation" |
| 2 | a built, tested MCP tool was reachable from no facade | a test **pinned the orphan state as expected** |
| 3 | a suffix fix was unreachable for ESM; the resolver was never called | tests called the private helper → read as "it works" |
| 4 | `ImportGraph` dropped every cross-directory edge on Windows | Linux CI structurally cannot see a separator mismatch |
| 5 | a doc example named a file that did not exist and whose name a hook bans | the ratchet parses `.py` via `ast`, not markdown |
| 6 | an agent silently falls back to `grep` when a route is too slow | non-use is never reported; VCSR cannot see it |

**Five of the six survived a green CI. The sixth is invisible to the north star itself.** None is a bug in a feature — each is a **missing invariant**.

This RFC adds **no product capability**. It converts four load-bearing claims into executable invariants and closes one measurement gap. It is CLAUDE.md §11 applied reflexively to the project's own foundational beliefs.

## §1 — The moat is currently prose

Conservative resolution — *a visible `unknown` is safer than a confident unsupported edge* — is the stated reason to prefer TSA over a text search for a safety question. **Nothing in CI measures it.**

RFC-0025's fail-closed work and RFC-0027's §L9.4 fence protect the *implementation*. Neither measures the *property*. A future change that reduces `unknown` counts in the name of "improved precision" would leave every existing test green, because every existing test asserts conformance to a shape, not the preservation of epistemic honesty.

The invariant is **one-directional**, which is what makes it stable:

> Where ground truth says an edge EXISTS but text search returns no hit, TSA MUST NOT report a confident empty result. It must report the edge, or report `unknown`. Never a bare zero.

Deliberately **not** asserted: that TSA resolves more edges than text search. That is a precision claim, it belongs to the RFC-0021 benchmark, and it is fragile. Ratcheted **per mechanism**, so a regression in one cannot hide behind an improvement in another.

## §2 — VCSR cannot see adoption failure

Measured on this repo (RFC-0025 L5 baseline, E0): `edit action=safe` is ~3.7 s p50 **even warm**, while every sibling route drops to milliseconds. That is squarely in the range where an agent routes around a tool — and it does so **without reporting it**.

So "TSA is fast and used" and "TSA is bypassed" are **indistinguishable in the north star**.

A `Consultation` record joins `Prediction` and `Outcome` in the L9 ledger, carrying elapsed time, cache provenance, and an explicit `declined_reason`, surfaced as decline rates in `--self-health`.

**The scope limit is in the design, not papered over:** nothing inside a process can record a call that did not happen. What this measures is the *precondition* — a route in seconds beside siblings in milliseconds — plus a place for a caller to decline explicitly instead of silently. Claiming more would be the same overreach this RFC exists to correct. No telemetry leaves the machine.

## §3 — The three classes the suite is structurally blind to

**Reachability** (defects 2, 3). Points the *existing* zero-caller machinery at three questions nobody asked: registered-surface reachability, `next_step` routability, and dispatch reachability **through public entry points rather than private helpers**. Uses only the zero-caller signal, whose false positives go one way, per CLAUDE.md rule 5. This is wiring, not construction.

**Gate effectiveness** (defects 1, 5). Every blocking gate must prove its detector still fires against live production with **exact set equality**, plus a coverage invariant of exactly `0` unwatched items, a fixture derived from **real** files rather than a synthetic imitation, and a CI job that can actually fail. #1314 is the precedent — the old `count > 0` guard was demonstrated to pass against a tree where **both detectors were fully dead**. Worth preserving in the record: *the anti-staleness guard was violating the exact-assertion rule it existed to enforce.*

Doc-example verification belongs here. The author of this RFC replaced a loose `>= 1` with a pinned `== 2` in `docs/TESTING.md` where the correct value was `1` — an exactly-wrong exact assertion, worse than the bound it replaced, because nothing runs those examples.

**Platform specificity** (defect 4). Any path-versus-AST-cache-key comparison must be covered under **both** separator conventions independent of host. Cheap, and it is the specific bug class already shown to be live.

## Out of reach, and said so

- A dead regex inside a `bash` gate stays unanalysable. §3.2 mitigates it **procedurally** — by requiring every gate to prove itself — not by teaching TSA shell.
- No claim that TSA can predict OS-specific runtime behaviour in general.
- No adoption telemetry. §2 is a local proxy and labels itself one.

## Scope

No new CLI flag, no new MCP action, no new facade. §1 and §3 are tests and gates. §2 extends `--self-health`. The locked TOON-on-MCP asymmetry is untouched.

**§3.1 is expected to fail on merge** — at least one existing test pins an orphan as expected. That is the point; the fix is to correct the assertion, not preserve it.

## Review focus

1. Is the §1 invariant genuinely one-directional, or is there a case where reporting `unknown` is itself wrong?
2. Is the §2 scope limitation stated honestly enough, given that non-invocation is unobservable from inside the process?
3. Does §3.2 impose a cost future gate authors will route around — and if so, is that better or worse than the current state?

### DIFF
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -3,24 +3,38 @@
 - **Status**: draft
 - **Author(s)**: @lead-agent
 - **Created**: 2026-08-19
-- **Last updated**: 2026-08-19
+- **Last updated**: 2026-08-20
 - **Tracking issue**: TBD
 - **Relationship to other RFCs**: complements [RFC-0025](0025-instant-causal-proprioception.md)
-  (sensing, L1–L5) and [RFC-0027](0027-agent-secretary-and-calibration.md)
-  (secretary layer, L6–L10). This RFC adds **no product capability**. It converts
-  four load-bearing claims into executable invariants and closes one measurement
-  gap in the north star. It is the reflexive application of CLAUDE.md §11 to the
-  project's own foundational beliefs.
+  (sensing, L1–L5), [RFC-0027](0027-agent-secretary-and-calibration.md)
+  (secretary layer, L6–L10), and [RFC-0029](0029-does-this-test-constrain-this-code.md)
+  (the on-demand "does this test constrain this code" query). It converts four
+  load-bearing claims into executable invariants and closes one measurement gap
+  in the north star. It is the reflexive application of CLAUDE.md §11 to the
+  project's own foundational beliefs. **It adds no product capability** — the one
+  surface change it does make (§1's `completeness` field) exists solely to expose
+  a property the AST cache already records and the response layer currently drops.
+- **Division of labour with RFC-0029** (drawn by *mechanism*, not by
+  gate-versus-test): **this RFC owns static, always-on, CI-enforced invariants**;
+  **RFC-0029 owns the on-demand, per-pair, non-gating query.** Under that split
+  the ESM-dispatch defect (#3) and the synthetic-fixture defect (#1) belong here;
+  RFC-0029 keeps them only as engine-validation fixtures, never as justification.
 - **Affected source paths** (pin them — reviewers watch for drift here):
+  - `tree_sitter_analyzer/mcp/tools/callers_tool.py` (§1 — `completeness` field)
+  - `tests/unit/mcp/tools/test_tool_response_contract.py` (§1 — response-surface contract)
   - `tests/benchmarks/claims/test_unknown_vs_absent_invariant.py` (new — §1)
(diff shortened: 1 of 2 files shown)


aimasteracc:
@codex review

PR #1324, opened 2026-08-20
fix(index): give the manifest WARN an actionable next_step (dogfood round 1)

aimasteracc:
## Found by using TSA, not by reading it

Dogfood round 1 (CLAUDE.md §11.4: *invoke the tools and look at the output, not run the suite*). Task under dogfood: preparing RFC-0025 L1.

`--full-index` on this repo returned **exit code 1**, `verdict: WARN`, `scope_complete: False`, `manifest_warning: INDEX_MANIFEST_CERTIFICATION_FAILED`. All of that is correct, and the code fail-closes properly — it revokes the call-graph-built marker in a **separate committed transaction** so direct `ASTCache` readers cannot trust the run, and it logs the cause with `exc_info=True`.

But `agent_summary` carried only `verdict` and `summary_line`. **`next_step` was absent.**

Every other route used in the same session carries one:

| route | `next_step` |
|---|---|
| `--callers` | "source bodies are inlined — answer directly, no Read needed" |
| `--test-map` | the exact pytest command |
| `--safe-to-edit` | the exact 10-file pre-edit command |
| `--self-health` | "no latency claim about it is supportable" |
| **`--full-index` (WARN)** | **nothing** |

So `next_step` is the field a caller is trained to read, and **the one output meaning "do not trust graph queries from this run" was the only one with nothing in it.**

The consequence was immediate and it was mine: I read `fts_indexed_symbols: 68767`, treated the run as success, and ran `--callers` against the uncertified index. The WARN and the exit code were both in output I had already printed.

## The fix

`manifest_warning_next_step()` maps every known warning code to a remedial action that **names the affected flags** (`--callers`, `--callees`, `--change-impact`) and says to re-index and confirm `verdict` is `INFO`.

An unrecognised code still returns a step **naming the code** — because a new warning falling back to silence is exactly how this gap was created. `None` in, `None` out: a certified run carries no remedial instruction.

## What the tool did well, for the record

`--safe-to-edit` on the L1 target returned `CAUTION`, health grade **D**, **10 downstream files**, and the precise pre-edit command. I did not know the target was grade D with 10 dependents, and it changed how I approached the edit. `--callers` returned 3 callers with each source body inlined, removing tool calls rather than adding them.

## Two findings filed rather than fixed

1. **The prescribed pre-edit verification is non-deterministic.** The same 10-file command gave `546 passed / 1 failed`, then `545 passed / 2 failed` **with a different name**. Deterministic only under `--override-ini="addopts="`, since `pytest.ini` defaults to `--numprocesses=4 --dist=worksteal --reruns=1`. A gate whose output varies run to run cannot be a pass/fail signal.
2. **The swallowed cause never reaches stderr.** `exc_info=True` to a logger nobody reads is not visibility.

## And four errors of my own, recorded because they are the same class

- I reported valid JSON (8,183 bytes) as non-JSON — my probe merged `2>&1` so `uv` warnings polluted the stream. **Measurement error attributed to the thing measured**, the exact class caught in three agent reports this week.
- I never checked `$?`. I piped through `tail -3`.
- I guessed 2 of 6 flag shapes wrong. 325 flags is past the point where shape is guessable.
- I made a structural edit with a naive text anchor, inserted module-level code into a class body and broke 24 tests — **after** `--safe-to-edit` handed me a pre-edit command I did not run.

## Verification

RED first (3 failed → green). Both candidate failures in the touched test file were measured **individually against a stashed baseline** and are identical there, so this change breaks nothing. `ruff` + `ruff-format` clean; `mypy --platform linux` clean on the touched module; full pre-commit hook set passed.

`STATE.md` records the round with **three columns, not two** — what helped, what failed, and **what I went around**. The third is the one that cannot be recovered later, because an agent that routes around a slow tool does not report it. RFC-0028 §2 exists for this and is currently an empty proxy; until it is populated the record is manual.

### DIFF
--- a/STATE.md
+++ b/STATE.md
@@ -1,5 +1,83 @@
 # Loop State — tree-sitter-analyzer
 
+## Dogfood round 1 — using TSA to build TSA (2026-08-20)
+
+Per CLAUDE.md section 11.4: a dogfood round must invoke the tools and look at
+the output, not run the suite. Task under dogfood: preparing RFC-0025 L1 (the
+standing index). Every timing below is wall-clock from the caller's seat,
+including interpreter start, because that is what the caller actually waits.
+
+### What the tool did well
+
+- `--self-health` on a cold process reported `ast_index: EMPTY`,
+  `NO_OBSERVATIONS`, `verdict: WARN` and *"no latency claim about it is
+  supportable"*. **It told me not to trust it before I asked it anything.**
+- `--callers _target_dependents` (3.4 s) returned 3 callers **with each source
+  body inlined**, and said so in `next_step`: "answer directly, no Read
+  needed". That removed tool calls rather than adding them.
+- `--test-map` (8.0 s) named the exact test file and function count.
+- `--safe-to-edit` on the L1 target returned `CAUTION`, health grade **D**,
+  **10 downstream files**, `edit_strategy: focused_edit_with_tests`, and the
+  precise 10-file pre-edit pytest command. I did not know the target was
+  grade D with 10 dependents; that changed how I approached the edit.
+
+### What failed, and what was changed as a result
+
+1. **`--full-index` WARN carried no `next_step`.** The run exited 1 with
+   `verdict: WARN`, `scope_complete: False`,
+   `manifest_warning: INDEX_MANIFEST_CERTIFICATION_FAILED` - all correct, and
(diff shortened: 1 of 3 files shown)


aimasteracc:
@codex review

PR #1395, opened 2026-09-07
fix(index): hotfix — 5 Python extraction defects (future import, nested def, ast_imports.line, noqa comment)

aimasteracc:
## Summary

MECE dogfood verification (stdlib `ast` vs `ast_symbol_rows`, all 1930 `.py` files in the repo) surfaced 5 silent data defects in the index extraction pipeline. All 5 are fixed in a single commit with TDD regression tests.

| BUG | Symptom | Root cause | Scope |
|---|---|---|---|
| 1 | `from __future__ import annotations` never indexed | `future_import_statement` node absent from `_IMPORT_LIKE` | 787 import rows missing |
| 2 | `def` nested in a method → `kind=method` + wrong class | `_find_parent_class` walked past `function_definition` | 334 symbols + 226 bogus `contains` edges |
| 3 | `ast_imports.line` always 0 for all 16,171 rows | INSERT omitted the `line` column; value was set upstream | 100% of `ast_imports` unusable for position |
| 4 | `from X import (  # noqa: F401 ↵ Any, Dict)` → 0 rows | `re.DOTALL` + `split("#",1)[0]` truncated the import body | 105 import rows missing (some files 100% lost) |
| 5 | `import os  # noqa` → 0 rows | `_PY_IMPORT_RE` char class excludes `#`; match fails | Additional silent zero-row case |

## Changes

- **`cache/extraction.py`** — add `future_import_statement` to `_IMPORT_LIKE`; stop `_find_parent_class` at `_CLASS_LIKE` first (preserves inner-class ownership), then at `_FUNCTION_LIKE`
- **`synapse_resolver/_imports.py`** — new `_strip_comments()` strips `#` per-line before regex matching (fixes BUG-4 and BUG-5)
- **`cache/write.py`** — add `line` to `ast_imports` INSERT; add `_import_lines_from_symbols()` to recover line numbers from the symbols payload
- **`ast_cache.py` / `cache/indexer.py`** — thread `symbols` through `_write_imports_for_file` to supply line numbers
- **`extractor_version` bumped 14 → 15** — BUG-1 and BUG-2 change persisted symbol rows; existing indexes auto-re-index on next build

## Test plan

- [x] 16 new TDD regression tests in `tests/unit/test_python_index_extraction_defects.py` — all 5 defects driven RED → GREEN
- [x] `tests/unit/test_symbols_json_enrichment.py` — extractor version pin updated to 15
- [x] `tests/unit/test_incremental_sync.py` — mock signature updated for new `symbols` param
- [x] 8,204 existing tests pass; 0 new failures (2 pre-existing failures on `main` unrelated)
- [x] Verified across 12 language plugins — BUG-2 fix causes no regression in Java, JS/TS, C/C++, Go, Rust, Ruby, Kotlin, C#, PHP, Scala, Swift
- [x] Independent code-review agent: **APPROVED**

### DIFF
--- a/tests/unit/test_incremental_sync.py
+++ b/tests/unit/test_incremental_sync.py
@@ -330,11 +330,11 @@ class TestSavepointRollbackOnPartialWrite:
 
         original_write_imports = cache._write_imports_for_file
 
-        def _fail_after_ast_index(conn, rel_path, language, imports):
+        def _fail_after_ast_index(conn, rel_path, language, imports, symbols=None):
             if "flaky.py" in rel_path:
                 # Simulate failure AFTER ast_index INSERT but BEFORE conn.commit().
                 raise RuntimeError("simulated mid-write failure")
-            return original_write_imports(conn, rel_path, language, imports)
+            return original_write_imports(conn, rel_path, language, imports, symbols)
 
         sync = IncrementalSync(cache)
         with patch.object(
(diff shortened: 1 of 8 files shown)


aimasteracc:
ok

PR #1455, opened 2026-09-12
feat(callers): declare completeness instead of implying absence

aimasteracc:
Implements the surface half of RFC-0028 §1.1: the callers response now declares its own completeness, and `next_step` may only claim absence when every inbound edge resolved. No ninth verdict.

## The problem, restated as the RFC states it

An empty caller list is not evidence that nothing calls a symbol. Four states existed in the data and only one answer was emitted: `NOT_FOUND` with `count: 0`, which reads as "safe to change". The RFC's rule is one-directional — TSA need not *resolve* every edge, it must not *present* an unresolved one as absence.

## Design

`completeness: "complete" | "incomplete" | "unknown"` rides beside the verdict:

| state | meaning |
|---|---|
| `complete` | every inbound edge resolved; a `caller_count` of 0 is a real zero |
| `incomplete` | callers were found **and** at least one inbound edge is unresolved; the list is a lower bound |
| `unknown` | no caller was listed **and** at least one inbound edge is unresolved |

The verdict vocabulary is untouched. `_LEGAL_VERDICTS` is a closed cross-surface contract pinned in two places, and widening it would force every consumer's branch table to change; an orthogonal field leaves every existing branch correct.

Two decisions worth review:

- **The count selects through `_direct_callers`** — the same function that builds the caller list — so a list and its completeness cannot disagree about which edges are in scope. This is load-bearing: an independent query would let `caller_count: 5` coexist with `completeness: complete` while an unresolved sixth edge existed.
- **The resolved set is inverted.** `_RESOLVED_CALLEE_RESOLUTIONS = {project, builtin, stdlib, local}`; anything else counts as unresolved, so the schema default, the second pass's `unresolved_refs` placeholder, and any marker a later version adds all degrade toward `incomplete`. Claiming a complete list over an unresolved edge is the error a caller cannot detect, so the safe direction is the over-cautious one.

An unreadable count returns `None` rather than `0`, and `None` maps to `unknown` — a failed read is not evidence of absence.

## Evidence

Measured on this repository's own source (903 files, 50,528 CALLS edges), which is why the field is not degenerate:

| `callee_resolution` | edges |
|---|---:|
| `project` / `builtin` / `stdlib` / `local` | 15,012 / 12,168 / 9,396 / 9,011 |
| `unknown` (unresolved) | 4,941 |

Only **549 of 7,176** distinct callee names have any unresolved inbound edge, so `complete` is the normal answer and the other two states are a meaningful minority. The three-state split over that corpus is 6,627 complete / 110 incomplete / 439 names with unresolved-only edges.

`unknown` is reachable in the way the RFC describes — through module-level call sites, which have no enclosing function and are counted rather than listed. Verified end-to-end on the indexed corpus:

```
SystemExit       verdict=NOT_FOUND count=0 completeness='unknown'    unresolved=2
writeText        verdict=NOT_FOUND count=0 completeness='unknown'    unresolved=1
setInterval      verdict=NOT_FOUND count=0 completeness='unknown'    unresolved=1
getElementById   verdict=NOT_FOUND count=0 completeness='unknown'    unresolved=1
Parser           verdict=INFO      count=50 completeness='incomplete' unresolved=2
ASTCache         verdict=INFO      count=56 completeness='complete'   unresolved=0
```

For `SystemExit`, the RFC's exact case: `next_step` no longer says "not in the index", it names the 2 unresolved edges, and `agent_summary.summary_line` reads *"has 0 resolved caller(s); this is not evidence of absence"*. Both surfaces emit the field — `tsa --callers SystemExit --format json` carries `completeness`, and on a cold graph it reports `unknown` rather than the old implied zero.

## Verification

- `tests/unit/mcp/tools/test_callers_completeness.py` — 6 new tests: each state, the `unknown`-never-reports-absence invariant, that a real zero still may say "not in the index", that a cold graph is `unknown` not `complete`, and that an unrecognised resolution marker counts as unresolved.
- `pytest` over the contract, nav-facade, edge-store and CLI↔MCP parity suites → **189 passed** (includes `_N_VERDICT_VOCABULARY` and `test_registered_mcp_tools_have_cli_parity`).
- `ruff check` clean.
- RFC-0028's three §1.1 acceptance boxes are now checked; the §1 corpus, the CI ratchet, and the §1.2 reconciliation are **not** — see below.

## Not in this PR

The RFC requires a hand-checked corpus of genuinely undecidable edges (dynamic dispatch, string-keyed invocation, reflection, decorator registration) plus a ratchet wired into a **named** CI job with its marker set recorded and a proven non-zero collected count. That is the test half of §1 and it is not faked here by promoting six fixture tests into its place. §1.2's reconciliation with `test_unknown_rate_ratchet.py` is likewise untouched.

### DIFF
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -679,10 +679,10 @@ rather than dropping it.
 
 ## Acceptance criteria
 
-- [ ] §1.1 `completeness` field emitted by the callers route on both surfaces;
+- [x] §1.1 `completeness` field emitted by the callers route on both surfaces;
       response-surface contract test updated; parity test green
-- [ ] §1.1 `_LEGAL_VERDICTS` **unchanged** — no ninth verdict added
-- [ ] §1.1 `next_step` no longer claims "Symbol not in the index" for an indexed
+- [x] §1.1 `_LEGAL_VERDICTS` **unchanged** — no ninth verdict added
+- [x] §1.1 `next_step` no longer claims "Symbol not in the index" for an indexed
       symbol with unresolved callers
 - [ ] §1 corpus + one-directional invariant green
 - [ ] §1 ratchet wired into a **named** CI job **and** its marker set recorded,
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -281,6 +281,51 @@ containing file has at least one unresolved edge. A symbol in a fully-resolved
 file must answer `complete`, so "answer `unknown` whenever the file contains a
 dynamic construct" is not a legal way to satisfy this ratchet.
 
+#### §1.2 amendment — the scope guard counted calls that cannot be local
+
+Measured 2026-09-12 on the 2,174-file self-repo corpus (164,392 CALLS rows),
+because implementation showed the guard above was far broader than its own
+justification. The sentence "an unresolved call inside a file may target anything
+that file declares" is true of the text; it is not true of the *edges* the guard
+was counting.
+
+| | gated files | share of the 2,060 files holding CALLS edges |
+|---|---|---|
+| guard as originally written | 1,562 | 75.8% |
+| after counting `external` as terminal | 1,363 | 66.2% |
+| after excluding builtin receivers (**amended definition**) | **1,029** | **50.0%** |
+
+The largest contributors were calls that cannot name a file-local symbol at all:
+`list.append` (2,327 rows), `dict.get` (369), `conn.execute` (446), `set.add`
+(275), `logger.debug` (215), `dict.items` (152), `super().__init__` (130). Before
+the amendment, a single `logger.debug(...)` barred every symbol in its file from
+ever answering `complete` — the failure the RFC's motivation attributes to
+computed dispatch, reproduced instead by ordinary attribute access.
+
+**What the guard now counts** is unchanged in intent and narrower in fact: an
+unresolved edge gates its file unless its callee is a dotted name whose receiver
+is a builtin type (`_can_target_local_symbol`, `graph/edge_store.py`). After the
+amendment, 3,635 unresolved edges are excused on this corpus and **6,674 are
+still counted** — bare names, `self.`/`cls.`, `getattr(...)`, string-keyed
+dispatch (`handlers["read_resource"]`), and any receiver that is a local variable
+(`extractors.run`) all keep gating, because each may name something the file
+declares.
+
+**Why the remaining half is not narrowed further.** The next-largest receivers are
+`conn` (689), `self` (302), `pytest` (258), `logger` (242), `tree_sitter` (172),
+`monkeypatch` (130), `extractor` (85), `path` (71). Excusing `pytest` and
+`tree_sitter` needs the file's import table joined to a project-module test;
+`conn`, `logger`, `monkeypatch`, `extractor`, and `path` are local variables bound
+to objects, which needs dataflow rather than syntax. Neither is available to the
+edge reader today, and both can under-count, which is the one direction §1
+forbids. They are recorded here as the measured remainder rather than guessed at.
+
+**Why this is a safe narrowing and not a relaxation.** Every exclusion requires
+proving the receiver is a builtin type; nothing is excused by absence of
+information. Over-counting only holds a symbol at `incomplete`, while
+under-counting certifies absence over a real edge — so the amendment removes only
+the cases where the exclusion is provable, and the ratchet above is unchanged.
+
 ### §2 — Consultation records: making non-use visible
 
 **Claim under test:** TSA is used.
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -684,14 +684,14 @@ rather than dropping it.
 - [x] §1.1 `_LEGAL_VERDICTS` **unchanged** — no ninth verdict added
 - [x] §1.1 `next_step` no longer claims "Symbol not in the index" for an indexed
       symbol with unresolved callers
-- [ ] §1 corpus + one-directional invariant green
-- [ ] §1 ratchet wired into a **named** CI job **and** its marker set recorded,
+- [x] §1 corpus + one-directional invariant green
+- [x] §1 ratchet wired into a **named** CI job **and** its marker set recorded,
       with a proven non-zero collected count (not merely a committed file)
-- [ ] §1.2 reconciliation with `test_unknown_rate_ratchet.py` recorded in both
+- [x] §1.2 reconciliation with `test_unknown_rate_ratchet.py` recorded in both
       test files, including which gate wins on collision
-- [ ] §1.2 scope guard green: a symbol in a fully-resolved file answers
+- [x] §1.2 scope guard green: a symbol in a fully-resolved file answers
       `complete`, never `unknown`
-- [ ] §1 documented as the executable form of the conservative-resolution claim,
+- [x] §1 documented as the executable form of the conservative-resolution claim,
       cross-referenced from `ROADMAP-no1-agent-trust.md`
 - [ ] §2 **blocked until RFC-0027 is accepted and L6.2 (`QueryCost`) lands** —
       not tickable before then
--- a/rfcs/ROADMAP-no1-agent-trust.md
+++ b/rfcs/ROADMAP-no1-agent-trust.md
@@ -390,7 +390,7 @@ commit changes.
 ### Continue
 
 - Continue local-first operation, project-root security, MCP/CLI parity, JSON for both MCP and CLI, and fail-closed benchmarks.
-- Continue conservative resolution: a visible `unknown` is safer than a confident unsupported edge.
+- Continue conservative resolution: a visible `unknown` is safer than a confident unsupported edge. [RFC-0028 §1](0028-measuring-claimed-properties.md) is its executable form — a hand-checked undecidable corpus (`tests/benchmarks/claims/test_completeness_honesty_ratchet.py`) fails any answer that is a confident empty over an edge that genuinely exists, and the `completeness` field on the callers route is what carries the distinction to the caller.
 - Continue dogfooding before edits and following the emitted verification command after edits.
 - Continue exact behavioral tests, but prefer realistic corpus failures over coverage-only growth.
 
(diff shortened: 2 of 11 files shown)


aimasteracc:
Added the §1 test half on top of this branch: a hand-checked undecidable corpus, the invariant it drives, and §1.2's reconciliation. It supersedes this PR's "Not in this PR" note, and it produced a finding that limits what the field above actually guarantees.

## The corpus

`tests/fixtures/call_graph/undecidable_project` — five symbols reached only through dispatch a text search cannot follow: a string-keyed handler registry, `getattr(self, name)`, `globals()[name]`, and `importlib` + `getattr`. A call-site search finds their definitions and never a call. A shape guard fails if the fixture ever becomes trivially decidable, and the corpus docstrings deliberately avoid writing a target name immediately before `(` for the same reason.

## What the corpus measured, and why it matters here

A computed dispatch site **does** leave an unresolved row — but keyed by the source expression:

```
('dispatch_by_name',     'HANDLERS[name]',         'unknown')
('run_named',            'getattr(self, name)',    'unknown')
('call_through_globals', 'globals()[name]',        'unknown')
('call_through_import',  'getattr(module, attr)',  'unknown')
```

No symbol is named `HANDLERS[name]`, so this PR's `count_unresolved_callers` — which selects unresolved edges *named after* the symbol — finds nothing unresolved for `handle_alpha` and answers `complete` with `caller_count: 0`.

All five corpus symbols are answered that way today:

```
handle_alpha          verdict=NOT_FOUND count=0 completeness='complete'
handle_beta           verdict=NOT_FOUND count=0 completeness='complete'
method_one            verdict=NOT_FOUND count=0 completeness='complete'
method_two            verdict=NOT_FOUND count=0 completeness='complete'
module_level_target   verdict=NOT_FOUND count=0 completeness='complete'
```

That is the one transition §1 forbids: a confident empty over an edge that genuinely exists. So this PR implements the **name-known-but-unbound** half of §1, and the corpus shows the other half is open. It is recorded as a strict `xfail` rather than papered over, per this suite's own rule for claims that are not yet met.

## The rule that closes it

§1.2's scope guard: §1 applies to symbols whose **containing file** holds at least one unresolved edge. A symbol reached by a computed dispatch in its own file must therefore not answer `complete`, and a symbol in a fully-resolved file must. That is implementable from data already present (unresolved rows carry their originating file), and it is the next change on this branch.

## §1.2 reconciliation, recorded in both files

`test_completeness_honesty_ratchet.py` and `test_unknown_rate_ratchet.py` now each state the relationship: different denominators (all `kind='calls'` rows repo-wide vs the hand-checked corpus), different units (percentage vs count), opposite desired directions, no conflict — a genuine resolver improvement satisfies both. **§1 wins on collision**: the existing cap may be re-pinned upward by a reviewed decision, this one may not be relaxed at all.

## CI

Wired through the existing named job rather than a new one: `dogfood-pr-check.yml` → "Run claim invariant suite", `-m "claims_benchmark and not full_language"`. Verified rather than assumed: that expression collects **3 tests** from the new file out of **26 collected** in the subset, and the subset is green at `25 passed, 1 xfailed, 12 deselected`.

aimasteracc:
## External adversarial review found a false zero in this change

I ran `opencode` against this branch with an adversarial prompt. It found **15 issues**; the first one is severe and I reproduced it exactly. Commit `1e563f39` fixes four of them.

### The severe one: this change produced the defect it exists to prevent

Reproduced on a fresh index — a symbol with two **resolved** module-level callers:

```
verdict=NOT_FOUND  caller_count=0  completeness="complete"
unresolved_inbound=0  unattributed_call_sites=2
summary   "callers: 'target_symbol' has 0 caller(s)"
next_step "Symbol 'target_symbol' not in the index."
```

The symbol **is** in the index and **is** called twice. `#638` counts module-level call sites instead of listing them, so the listed count is not a census — and because every edge was *resolved*, the per-edge signal stayed silent while the unattributed count knew. `complete` is now impossible whenever a caller went unlisted. After the fix: `completeness='incomplete'`, and both the summary and next_step name the module-level sites.

### Also fixed

- **A read `0` was treated as a failed read.** `if unresolved_inbound:` is falsy for a successfully-read 0, so a response beside a non-zero declaring-file count said *"its inbound edges could not be read"* — false. Branches now test `is None`, and each reason names itself.
- **A failed read degraded to zero.** `symbol_declaring_files` returned `()` on `OperationalError`, which the guard read as "no file scope" → 0 unresolved. It now returns `None` and the guard propagates, matching the sibling count method's contract.
- **`file_path` was ignored by the declaring-file guard.** A symbol declared in several files inherited every sibling's unresolved calls: `run` in a clean file answered `unknown` for a query naming the clean file, contradicting §1.2's guard.

Two tests added for cases that had none, and the pre-existing summary assertion tightened to require the caller be named.

### Not addressed here, and worth your judgement

- **Blast radius.** The reviewer measured that **607/829 files with call edges hold ≥1 unresolved edge, so 14,914/17,885 symbols can never answer `complete`** — and that the gating edges are ordinary attribute calls (`logger.debug`, `super().__init__`), not the computed dispatch §1.2's motivation describes. That is §1.2's scope guard working as written, but it means `complete` is rare in practice. Worth deciding whether the guard should be narrower than "file contains any unresolved edge".
- **`external` is a decided terminal state** the resolver writes, and it is absent from `_RESOLVED_CALLEE_RESOLUTIONS`, so it poisons file scope wherever that tier is used. 0 rows in today's index, live path.
- **Two count paths read resolution differently**: the per-symbol count goes through `_row_col` (metadata fallback for legacy rows), the file-scope count uses raw SQL on the column only. A legacy row is resolved in one scope and unresolved in the other.
- **`caller_count` dedups by (file,name,line) while `unresolved_inbound` counts raw edges**, so the two can disagree on one line.

I have not verified the last three independently — they are the reviewer's claims, and I would rather flag them than silently accept or dismiss them.

## Review provenance

`codex review --base develop` also ran on the §3.1 reachability branch and concluded no correctness regression from the dead-arm deletion: *"The modified import-dispatch behavior is limited to an explicit unreachable fallback path... I did not find a correctness regression introduced by this patch."*

Both reviewers needed host escalation and a full (non-shallow) clone; the shallow clone broke `git merge-base` and codex could not start at all until `git fetch --unshallow`.

aimasteracc:
Substantial follow-up pushed (`d2224216`). Three of the four items are defects this PR introduced or left; the fourth is the §1.2 scope guard, now amended in the RFC with measurements.

## Why CI was red

The failures on this branch were **not** platform-specific. `pytest.ini`'s `testpaths` is a curated list, so the repo's documented quick gate (`uv run pytest -q`) collects **2,144** tests while CI runs **25,224**. Two pre-existing tests in `tests/unit/test_call_graph_built_signal.py` were broken by this branch and were invisible to the quick gate.

They now pass **unmodified** — see the first fix below, which turned out to be the actual root cause rather than a wording change.

## 1. A readable zero was reported as unreadable

`unresolved_inbound` and `unresolved_in_declaring_files` were gated on `data_source == "sql"`, so every query taking the in-memory graph path set both to `None` and answered `completeness: "unknown"` with next_step *"its inbound edges could not be read."* The counts were never attempted, not unavailable.

That is a false claim of ignorance, and it made `complete` unreachable on the graph path entirely. Now the cache answers on every data source. The empty-evidence case is kept separate: a zero over a graph holding no CALLS edges at all stays `unknown`, preserving #705's rule that a built index with no calls must not be told to `--full-index` again.

## 2. `external` was read as unresolved though its writer treats it as terminal

`synapse.py` refuses to re-scan edges already marked `('external', 'stdlib')` — *"target lives outside the project, no resolved_file by design."* The reader's resolved set omitted `external`, so it disagreed with the writer about a finished edge.

Measured on the self-repo corpus: **4,247 CALLS rows (2.58%)** carry it; counting them as unresolved gated **1,562 files instead of 1,363**. The inverted-by-design rule is unchanged — only known-terminal markers are listed; an unrecognised marker still reads as unresolved (existing test still green).

## 3. The two counts read the marker differently, and the docstring was false

The per-symbol count read through `_row_col` (metadata fallback); the per-file count aggregated the column with raw `NOT IN`. Both now share one predicate, `_resolution_is_resolved`.

The docstring claiming the count "selects through `_direct_callers`, the same path the caller list uses" was **false**: the list comes from `query_callers` (`_bfs_call_edges`), deduplicated, with module-level rows dropped. It now says what is actually counted and why the units differ. Also pinned: `edges` is UNIQUE over `(source_node_id, target_node_id, kind, line)`, so duplicates cannot inflate a count.

## 4. §1.2 scope guard amended

The guard counted unresolved calls that provably cannot be local. Measured on 2,174 files / 164,392 CALLS rows, the top offenders are `list.append` (2,327), `dict.get` (369), `conn.execute` (446), `set.add` (275), `logger.debug` (215). A single `logger.debug` barred every symbol in its file from `complete` — the failure the RFC's motivation attributes to computed dispatch, produced instead by ordinary attribute access.

| | gated files | share |
|---|---|---|
| as originally written | 1,562 | 75.8% |
| after `external` fix | 1,363 | 66.2% |
| **after narrowing** | **1,029** | **50.0%** |

`_can_target_local_symbol` excuses a dotted callee **only** when its receiver is a builtin type name, matched strictly so `getattr(importlib.import_module(m), cls).x` and `handlers["read"]` are not misread. **3,635** unresolved edges are excused; **6,674 are still counted** — bare names, `self.`/`cls.`, reflection, string-keyed dispatch, and local-variable receivers all keep gating.

The measured remainder is recorded in the RFC rather than guessed: `conn` (689), `self` (302), `pytest` (258), `logger` (242), `tree_sitter` (172), `monkeypatch` (130), `extractor` (85), `path` (71). Excusing module receivers needs an import table joined to a project-module test; the rest are local variables needing dataflow. Both can under-count, which is the one direction §1 forbids.

## Verification

Every fix has a test that fails without it: five for the resolution/empty-evidence rules, eight for the narrowed guard (one requiring the exemption, seven asserting it does not spread).

- `uv run pytest -q tests/` → **24,891 passed, 180 skipped, 0 failed**
- `uv run pytest -q` → **2,111 passed, 28 skipped**
- RFC-0028 acceptance items unchanged at **23**, none un-checked

aimasteracc:
Superseded by #1465. The gitflow guard correctly rejected this branch name — PRs to `develop` must use a GitFlow prefix, and `rfc-0028-callers-completeness` matches none. Renaming through the API closed this PR rather than retargeting it, so the same commit (`d2224216`) was re-opened as #1465 from `feature/rfc-0028-callers-completeness`. No content changed.

PR #1456, opened 2026-09-12
feat(cost): declare QueryCost on the three expensive routes (RFC-0027 L6.2)

aimasteracc:
Implements RFC-0027 L6.2. A budget-bound agent can now see what an expensive route will cost before committing to it, in one vocabulary, instead of discovering the cost by hitting a timeout.

## Shape

`QueryCost` exactly as the RFC specifies it, plus a renderer that omits the key entirely for routes that declare nothing:

```python
QueryCost(tier, estimated_ms, requires_index_build, cheaper_alternative)
tier: Literal["cached", "warm", "cold", "index_build"]
```

Returned by the three most expensive routes.

## `estimated_ms` is `None`, and that is the contract

The RFC requires it be derived from the L9 ledger's observed p50 and says it is *"never a hardcoded guess"*. The ledger does not exist, so the honest value is `None` — and a test asserts it for every listed route, so a future edit cannot quietly substitute a number that would be indistinguishable from a measured one. Wiring the ledger is L9's work, not a prerequisite for declaring cost.

## Tiers come from observed state, not from a heuristic

| tier | when |
|---|---|
| `index_build` | the on-disk index holds no rows, and the route reads it |
| `cached` / `cold` | answer-cacheable route: whether this exact call is already answerable |
| `warm` | index present, nothing else to claim |

The `cached`/`cold` split is not decorative — it is the distinction the committed L5 baseline measures. In one run `edit action=safe` computed in **3453 ms** and then served in **51 / 51 / 69 ms**, with `served_from` going `computed` → `cache`. So a cache miss *is* the expensive call, and a route without an answer cache gets `warm`, never a guessed `cold`.

`_index_holds_rows` deliberately does **not** use `auto_index_guard.is_indexed`: that is the in-memory guard marker, and `CodeGraphAutoIndexTool`'s own comment records a defect caused by reading it as on-disk state. It uses the same on-disk definition that tool reports as `indexed`.

## Why these three routes, and why `requires_index_build` is per-route

- `nav action=callers` and `edit action=safe` are the two paths RFC-0027's own dogfood table names — the 24.8 s → 16 ms cliff, and the route an agent runs before every edit with no warm benefit before L6.1.
- `health action=project` measured **202 s** on a 268-package TypeScript monorepo, past the 60 s per-call default most MCP clients apply, so it is the third.
- `structure action=outline` is deliberately **absent**: the answer-cache policy records it at 2.9 ms warm, where a declaration would be noise.

`requires_index_build` is declared per route from evidence rather than probed globally, because the routes genuinely differ: the L5 baseline computed `edit action=safe` in 3453 ms with `preconditions.ast_index_before.status: ABSENT`, so a missing index is not a cost for that route and reporting `index_build` for it would overstate the cost. A test pins that asymmetry.

## Verification

- `tests/unit/test_query_cost.py` — 10 new tests: unlisted routes declare nothing; `estimated_ms` is never guessed; an index-reading route reports `index_build` when the index is absent while `edit action=safe` does not; cache miss is `cold` and cache hit is `cached`; the rendered field set is exact.
- Driven on a real indexed corpus and on an index-less project, through both surfaces:

| route | surface | result |
|---|---|---|
| `nav.callers` | MCP + CLI | `warm`, cheaper alternative `nav action=context (1-hop, warm)` |
| `edit.safe` | MCP + CLI | `cold` (the computed call), `requires_index_build: False` |
| `health.project` | MCP | `warm`, cheaper alternative `health action=file` |

- `pytest` over the response-contract, CLI↔MCP parity, callers and safe-to-edit suites → **85 passed**.
- `ruff check` clean.
- RFC-0027's L6.2 acceptance box is now checked.

## Note on the existing field

`project_health_tool` already carried an ad-hoc `estimated_seconds`, and its own comment records that agents obeyed it and still timed out on 4k-file repositories. That field is left in place; the uniform declaration is added beside it rather than replacing a value consumers may already read. Consolidating the two is a follow-up worth doing on its own.

### DIFF
--- a/rfcs/0027-agent-secretary-and-calibration.md
+++ b/rfcs/0027-agent-secretary-and-calibration.md
@@ -683,7 +683,7 @@ and emits a calibration report with non-placeholder numbers.
 - [x] L6.1 the harness drives the prescribed route pair **interleaved** and
       records `served_from` per call, so a 0% real-workflow hit rate cannot hide
       behind a grouped-repeat speedup
-- [ ] L6.2 `QueryCost` returned by the three most expensive routes;
+- [x] L6.2 `QueryCost` returned by the three most expensive routes;
       `estimated_ms` is `None` until observed
 - [ ] L6.3 canonical full id always present; abbreviation unique within the
       project index; `SYMBOL_ID_AMBIGUOUS` fail-closed on an ambiguous
--- a/tests/unit/test_query_cost.py
+++ b/tests/unit/test_query_cost.py
@@ -1,0 +1,112 @@
+"""RFC-0027 L6.2 — a route declares what it will cost before it runs.
+
+The load-bearing assertion is that ``estimated_ms`` stays ``None``: the RFC
+forbids a hardcoded guess, and a number invented here would be indistinguishable
+from a measured one.  Tiers, by contrast, are derived from state this process
+can observe.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from tree_sitter_analyzer.cache import query_cost as qc
+
+
+@pytest.fixture
+def indexed_project(tmp_path):
+    """A project whose on-disk index holds rows."""
+    source = tmp_path / "a.py"
+    source.write_text("def target_symbol():\n    return 1\n", encoding="utf-8")
+    from tree_sitter_analyzer.ast_cache import ASTCache
+
+    cache = ASTCache(str(tmp_path))
+    try:
+        assert cache.index_file(str(source))["status"] == "indexed"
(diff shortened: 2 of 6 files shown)


aimasteracc:
Superseded by 1469. The gitflow guard correctly rejected this branch name — PRs to `develop` must use a GitFlow prefix, and it matched none. Renaming through the API closed this PR rather than retargeting it, so the same commit was re-opened from `feature/rfc-0027-l62-query-cost`. No content changed.

PR #1458, opened 2026-09-12
test(rfc-0028): add the §3.1 registered-surface reachability gate

aimasteracc:
Implements RFC-0028 §3.1's registered-surface reachability invariant — the gate the disposition module deliberately left to this RFC ("dispositions first, gate second").

## Why the existing tests do not cover this

`test_tool_dispositions.py` verifies the six orphans measured on 2026-08-19. It can only check names that are already in the registry, so a **seventh** orphan — a tool class added later, built, tested, registered nowhere — has no disposition to check and no test that notices. This gate enumerates every `BaseMCPTool` subclass rather than consulting a list.

## It found the seventh immediately

`MCPTool` (`base_tool.py:514`) is a backward-compatibility protocol base whose own docstring says *"deprecated, use `BaseMCPTool` instead"*. It is non-abstract, defines a concrete `execute`, and is reachable from nothing. Dispositioned `deprecate` with a removal version, not allowlisted — the only outcome §3.1 permits.

## A tension in §3.1's own wording, measured

The stated exemption rule is *abc-abstract or no concrete `execute`*, and the section names `FacadeTool`, `MCPTool` and `_CallTreeBase` as legitimately not registered. Measured: `MCPTool` and `_CallTreeBase` are **neither** abstract **nor** without a concrete `execute`, so the rule as written does not exempt them. What carries them is reachability by MRO: `edit action=pr` holds a `_PRReviewViaFacade`, not a `CodeGraphPRReviewTool`, and `nav action=callers` is a bespoke closure rather than an `action_map` entry — so an `action_map`-only walk reports a live route as an orphan. The gate walks both, and the rule is still implemented exactly as written (a probe that omits `validate_arguments` stays abstract and is correctly exempt). Recorded in the RFC rather than silently resolved.

## Verification

- The gate is green on this tree: `11 passed` with the disposition suite.
- **It fails on a genuine orphan.** Adding a fully-implemented, unregistered tool class produces exactly `ZZOrphanProbeTool: built but reachable from nothing, and undispositioned`. The first probe attempt *passed*, because omitting `validate_arguments` leaves the class abstract — which is the structural exemption working, and the reason the proof needed a complete probe rather than an assertion.
- `ruff check` clean; `28 passed` on the CLI↔MCP parity and response-contract suites.

### DIFF
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -720,9 +720,19 @@ rather than dropping it.
       probe case below depends on it — but it is not what makes those two
       classes pass.
 - [ ] §3.1 `next_step` routability green under the token-matching formulation
-- [ ] §3.1 dispatch reachability asserted through public entry points
-- [ ] §3.1 zero-caller signal exempt from §1's ratchet, asserted by a test that
+- [x] §3.1 zero-caller signal exempt from §1's ratchet, asserted by a test that
       fails if a genuine orphan starts answering `unknown`
+
+      Asserted as a **hard binary** rather than as a vocabulary check.
+      `_offender_reason` in `tests/unit/mcp/test_registered_surface_reachability.py`
+      is pure, so the three properties are testable directly: a genuine orphan
+      always yields a verdict; no verdict borrows §1's vocabulary
+      (`unknown` / `incomplete` / …), which is what a softened signal would look
+      like; and a live `deprecate` is the only soft outcome, bounded by its
+      removal version. The exemption holds **by construction** — §3.1 reads the
+      registry and the dispositions, never `completeness` — and these tests are
+      what keep it that way if someone later reaches for the field.
+- [ ] §3.1 dispatch reachability asserted through public entry points
 - [ ] §3.2 each of the **six first-party local hooks** has an exact-equality
       self-check; the four with a live surface additionally have a
       coverage-invariant of exactly 0
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -732,7 +732,26 @@ rather than dropping it.
       removal version. The exemption holds **by construction** — §3.1 reads the
       registry and the dispositions, never `completeness` — and these tests are
       what keep it that way if someone later reaches for the field.
-- [ ] §3.1 dispatch reachability asserted through public entry points
+- [x] §3.1 dispatch reachability asserted through public entry points
+
+      Asserted per **call site**, not per resolver.
+      `tests/unit/mcp/test_dispatch_reachability.py` drives `ImportGraph.build()`
+      with one fixture per language family and pairs each resolver invocation with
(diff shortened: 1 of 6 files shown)


aimasteracc:
Added §3.1's zero-caller exemption on top of this branch.

The RFC states the failure shape this guards: the cheapest way to maximise §1's ratchet is to answer `unknown` whenever a file contains a dynamic construct — and dynamic construction is how these very subjects are built (`health_facade.py` assembled by a function-local import, `plugins/manager.py:72` using `importlib`). Under that implementation a genuine orphan reads as indeterminate, §3.1 stops detecting anything, and **both gates go green**: a fresh instance of the RFC's own defect shape.

`_offender_reason` is now a pure function, which makes the property testable instead of merely intended:

- a genuine orphan **always** yields a verdict;
- **no verdict borrows §1's vocabulary** (`unknown` / `incomplete` / …) — that is what a softened signal would look like;
- a live `deprecate` is the **only** soft outcome, bounded by its removal version.

The gate's behaviour is unchanged and still fails on a genuine orphan: re-verified after the refactor by injecting a fully-implemented, unregistered tool class, which produces `ZZOrphanProbeTool: built but reachable from nothing, and undispositioned`.

The exemption holds by construction — §3.1 reads the registry and the dispositions, never `completeness` — and these tests are what keep it that way if someone later reaches for the field.

Verification: 25 passed across the reachability, disposition and CLI↔MCP parity suites; ruff clean.

aimasteracc:
Added §3.1's dispatch reachability gate — and it found defect #3's **second instance**, still standing after #1312.

The gate drives `ImportGraph.build()` with one fixture per language family and pairs each resolver invocation with the **line inside `_resolve_import`** that made it, then requires that set to equal the resolver call sites parsed from the dispatch. Pairing per call site rather than per resolver matters: after the #1312 fix the extension path *is* reachable, so a coarse "was the js resolver called" assertion would have passed while a **second, dead `_resolve_js_import` call site** sat beside it.

It was dead. Every file that can carry recorded imports has an extension in `EXT_TO_LANG`, so a JS/TS file is routed by language before the text sniff, and no other extractor emits `require(` / `from '` / `from "` — measured across C, Go, Java, Rust, Lua, Ruby, PHP. No test called the private dispatch either, which is why the helper's unit tests passed while the branch was unreachable: exactly defect #3's shape.

Deleted rather than allowlisted, and the surviving Python-shaped arm is now documented with the producers that actually reach it through `build()` (Go's `import "x"`, Java's `import a.b;`).

Verification: gate green (3 passed); `test_import_graph.py` 30 passed; dependency-matrix + parity 63 passed with **no behaviour change**, since the removed arm never fired.

aimasteracc:
Superseded by 1468. The gitflow guard correctly rejected this branch name — PRs to `develop` must use a GitFlow prefix, and it matched none. Renaming through the API closed this PR rather than retargeting it, so the same commit was re-opened from `test/rfc-0028-31-reachability`. No content changed.

PR #1459, opened 2026-09-12
test(rfc-0028): gate next_step routability, and make "removed" checkable

aimasteracc:
Implements RFC-0028 §3.1's `next_step` routability invariant, and corrects the RFC's own account of the defect it names.

## The invariant, as the RFC states it

> Any `next_step` containing a token matching a known tool / facade / action name must resolve to a **registered** route.

The obvious formulation is false and the RFC says so: real emitted strings are English prose ("Pass roots within the project boundary."), so the gate must constrain only tokens that *are* route names. It harvests the string constants assigned into `next_step` by AST (145 across 27 modules) and resolves every token colliding with the published vocabulary.

## The RFC's named defect is fixed, and was mis-located

Measured: `get_project_summary` is **back in** `facade_map.LEGACY_TOOL_MAP` (→ `project action=card`), and its mention in `build_project_index_tool.py` is inside the tool **description**, not a `next_step`. So the live instance is gone. All 10 legacy names currently emitted in `next_step` strings resolve to registered routes:

```
analyze_code_structure -> structure.analyze      ast_cache              -> index.cache
build_project_index    -> index.build            codegraph_symbol_search -> search.symbol
detect_routes          -> health.routes          extract_code_section   -> structure.read
get_code_outline       -> structure.outline      list_files             -> project.files
refactoring_suggestions-> edit.refactor          trace_impact           -> nav.trace
```

## What measuring it exposed: a class the wording cannot reach

I first proved the gate by injecting a `next_step` naming a removed capability — and **it passed**. The reason is structural, not a bug in the harness: a name that a breaking change *removed* (`search_content`, `find_and_grep`) resolves nowhere by construction, so a vocabulary built from names that currently resolve cannot see it. "Removed" existed only as prose in comments and `CHANGELOG`.

`facade_map.REMOVED_TOOL_NAMES` gives it a machine-checkable home, next to the crosswalk that already owns the name mapping. With it, injecting `search_content` fails the gate:

```
search_content (in get_code_outline_tool.py)
search_content (in get_code_outline_tool.py): matches no legacy name and no registered facade
```

## Verification

- `4 passed` on the gate; `41 passed` across the disposition, CLI↔MCP parity and response-contract suites; `ruff check` clean.
- The gate is proven non-vacuous in both directions: a harvest assertion fails if the AST walk stops matching (it requires >50 strings and ≥5 vocabulary tokens), and the injection above proves it rejects a dead route.
- A third test pins the *prose* case — a `next_step` that names no route must not be constrained — because a gate that failed "Pass roots within the project boundary." would be wrong rather than strict.

## RFC corrections in the same change

Two passages assumed the now-fixed defect. Both are corrected rather than deleted, with the measurement recorded: the §3.1 design note (mis-located defect + the removed-name class) and the test plan's RED-first item, which says plainly that RED was not available here and why injection is the equivalent evidence.

### DIFF
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -396,9 +396,21 @@ asked:
    > Any `next_step` containing a token matching a known tool / facade / action
    > name must resolve to a **registered** route.
 
-   That still catches the live defect: `build_project_index_tool.py` emits
-   `next_step="get_project_summary"`, a token that matches a tool name and
-   resolves nowhere.
+   That formulation still catches the class this section records. *Measured
+   correction (2026-09-12).* The defect named here has since been fixed, and was
+   mis-located: `get_project_summary` is back in `facade_map.LEGACY_TOOL_MAP`
+   (resolving to `project action=card`), and its mention in
+   `build_project_index_tool.py` is inside the tool **description**, not a
+   `next_step`. So the live instance is gone; the class is not.
+
+   The gate that covers it (`tests/unit/mcp/test_next_step_routability.py`)
+   harvests `next_step` string constants by AST and resolves every token that
+   collides with a published name. Measuring it exposed the class this section's
+   wording cannot reach: a name a breaking change **removed** resolves nowhere by
+   construction, so a vocabulary of currently-resolving names cannot see it.
+   `facade_map.REMOVED_TOOL_NAMES` is what makes "removed" machine-checkable, and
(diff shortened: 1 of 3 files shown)


aimasteracc:
Superseded by 1467. The gitflow guard correctly rejected this branch name — PRs to `develop` must use a GitFlow prefix, and it matched none. Renaming through the API closed this PR rather than retargeting it, so the same commit was re-opened from `test/rfc-0028-31-next-step-routability`. No content changed.

PR #1461, opened 2026-09-12
test(rfc-0028): live unknown-rate measurement + first-party gate inventory

aimasteracc:
RFC-0028 §3.2's fourth acceptance item: replace two tests that could not fail.

## What they asserted

- `test_unknown_rate_threshold_value` asserted `UNKNOWN_RATE_THRESHOLD_PCT == 6.0` — a constant equalling itself.
- `test_unknown_rate_threshold_is_documented_in_this_file` asserted this file contains the string `"6.0"` — a file containing its own text.

Neither observed the code, so neither noticed the measured rate had drifted **above** the ceiling they pin.

## Measured 2026-09-12

| population | CALLS edges | unresolved | rate |
|---|---:|---:|---:|
| product source (`tree_sitter_analyzer/`) | 50,528 | 4,941 | **9.78%** |
| self-repo subset (2,174 files) | 164,392 | 10,309 | **6.27%** |
| ceiling | | | 6.0% |

The claim is **not met**, and has not been for an unknown length of time. This is §3.2's premise demonstrated on the project's own gate.

## The replacement

`test_unknown_rate_is_measured_within_threshold` indexes the product source itself (`workers=1`, so the number cannot depend on pool scheduling) and asserts against the ceiling. It inherits this file's existing `full_language` marker, which is the axis where a ~13 s build is affordable. It is a **strict `xfail`**: the finding is recorded rather than hidden, and the marker comes off only when the resolver genuinely improves.

## It also enforces the rule the old test only documented

Proved by experiment rather than asserted: raising `UNKNOWN_RATE_THRESHOLD_PCT` to `10.0` makes the assertion pass, which turns the `xfail` into an **unexpected pass and fails the run**. "The threshold must NEVER increase without a reviewed decision" is now a failing condition instead of a comment.

Verification: `ruff check` clean; the test reports `1 xfailed` with the measured value printed; the threshold-raise experiment fails as intended; the fast suite deselects it as before.

### DIFF
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -730,9 +730,20 @@ rather than dropping it.
 - [ ] §3.2 every gate names its **marker set + workflow job** (or its pre-commit
       hook) and proves a non-zero collected count
 
-      *Partially landed.* The inventory gate asserts each scoped hook runs at the
-      `pre-commit` stage, which is the layer that enforces it. A non-zero
-      collected count per marker set is not yet asserted.
+      *Partially landed.* Both halves are now enumerated and checked:
+      `tests/governance/test_first_party_gate_inventory.py` asserts every scoped
+      hook runs at the `pre-commit` stage, and
+      `tests/governance/test_workflow_marker_sets.py` derives **10 marker-driven
+      gates across 6 workflows** from the workflow YAML, names each one's job,
+      and asserts every positive marker in each expression is carried by at least
+      one test — the vacuity case, where `pytest -m "<marker>"` collects nothing
+      and always succeeds. It also pins the one shell-templated set
+      (`e2e${EXTRA_MARKS}`) as a decision rather than a silent skip.
+
+      **What is still missing is the collected count itself.** Collection costs
+      5–15 s per expression here, so five expressions would add about a minute to
+      the fast suite; the static check catches vacuity and typos without that
+      cost, and does not read a count.
 - [ ] §3.2 the enforcement layer is named per gate: either
       `required_status_checks` added to a `develop` ruleset, or pre-commit
       declared as the blocking layer
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -451,17 +451,26 @@ proving its detector still matches live production, asserting **exact** set
 equality — not a `count > 0` lower bound.
 
 **Scope, stated precisely, because "every blocking gate" is unimplementable.**
-`.pre-commit-config.yaml` declares **27 hooks, of which 21 are third-party**:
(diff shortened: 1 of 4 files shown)


aimasteracc:
Added a second §3.2 commit to this branch: the first-party gate inventory is now derived from `.pre-commit-config.yaml` rather than restated.

Writing it found the drift immediately. **§3.2 names six local hooks; the configuration declares seven.** The unnamed one, `test-encoding-ratchet`, is itself a ratchet — the same class as `weak-assertion-ratchet`, whose dead-detector defect §3.2 records as #1. The section's "27 hooks, of which 21 are third-party" is likewise **28/21**, and its scope table omitted the hook entirely, in a section that describes its own scope as "stated precisely".

Both numbers now derive from the config, and the table is checked against it, so the spec cannot drift from the file it describes. The gate also asserts each hook's `entry` resolves to a file that exists — a hook whose entry points at nothing is a gate that cannot gate.

Proven to reject the drift it was written for: removing the `test-encoding-ratchet` row fails it with `in config but not scoped: ['test-encoding-ratchet']`.

The three §3.2 acceptance items above it stay **unchecked**, and each now carries a note saying how far it has got rather than implying progress. `4 passed`; ruff clean.

aimasteracc:
Third §3.2 commit: the marker-driven gates are now enumerated from the workflow YAML rather than assumed.

It derives **10 marker-driven gates across 6 workflows**, names each one's job, and asserts every positive marker in each expression is carried by at least one test — the vacuity case, where `pytest -m "<marker>"` collects nothing and always succeeds. The single shell-templated set (`e2e${EXTRA_MARKS}`) is pinned as an explicit decision, not skipped silently.

**Two parsing mistakes it caught in its own drafts, both by failing rather than by passing quietly:**

1. matching `pytest` and `-m` on the same line found only 2 of the gates — these invocations wrap across continuation lines with a trailing backslash;
2. scanning the whole `run` block for `-m "..."` then matched `git commit -m` and `gh ... -m`, which take a message, not a marker set — **23 phantom gates**.

Both were caught by the survey's own non-vacuity assertion rather than by review.

**What shapes the design is cost.** `pytest --collect-only` runs 5–15 s per expression on this tree, so five expressions would add about a minute to the fast suite — and a gate that slows the default loop is a gate people disable. The static check proves non-vacuity and typo-safety without that cost; it does **not** read a collected count, and the RFC says so rather than implying otherwise.

`8 passed` across both governance gates; ruff clean.

aimasteracc:
Superseded by 1466. The gitflow guard correctly rejected this branch name — PRs to `develop` must use a GitFlow prefix, and it matched none. Renaming through the API closed this PR rather than retargeting it, so the same commit was re-opened from `test/rfc-0028-32-unknown-rate-live`. No content changed.

PR #1464, opened 2026-09-12
fix(gates): anchor two first-party gates to the repo root, add self-checks, and run them

aimasteracc:
RFC-0028 §3.2 requires every gate with a live surface to carry a self-check that
fails when the gate stops watching, and requires that check to be *reached* by a
layer that runs it. This PR lands both halves for the first two gates, plus a
real defect the work uncovered and a correction to §3.2's own scope table.

## The defect

Both gates resolved their watched paths against the **caller's cwd**. Run from
`scripts/` — or anywhere but the repository root — they scanned nothing and
exited `0`, indistinguishable from a clean scan.

| Command | from repo root | from `scripts/` |
|---|---|---|
| `check_test_encoding.py --baseline` | `1831 across 286 files` | **`0 across 0 files`, rc=0** |
| `check_ps_ascii.py` | 27 live / 30 watched | **0 files, rc=0, no output** |

Both now derive their watch set from `REPO_ROOT = Path(__file__).resolve().parents[1]`.

## The self-checks

RFC-0028 §3.2 requires exact set equality between two independent derivations, not
a `count > 0` bound. Both comply:

- **`check_ps_ascii.py --self-check`** — watched set from the glob patterns; live
  set by walking `.github` for `run:` blocks. The invariant is the non-empty
  difference between two real sets.
- **`check_test_encoding.py --self-check`** — pins the scan base to
  `REPO_ROOT/tests` and requires every live file to parse under `ast.parse`.

## They are actually executed (requirement 4)

`tests/governance/test_gate_self_checks.py` runs every self-check from a
non-root cwd and asserts each gate's `(rc, stdout, stderr)` is identical from the
root and from `scripts/`. Layer named per requirement 4: `tests/governance` is in
`pytest.ini` `testpaths`, and the module is unmarked, so `reusable-test.yml`'s
matrix expressions select it. Pre-commit does not run it; CI plus the local quick
gate is the blocking layer, and the docstring says so.

**Non-zero collection: 4 tests.**

## Reverse verification

Every check was planted-defect tested, not merely observed to pass:

| Planted defect | Result |
|---|---|
| `tests/_zz_selfcheck_probe.py` with a syntax error | `1 live files do not parse` |
| `.github/zz-probe.yml` with `run:` outside the filter | `1 YAML file(s) with run: outside the watch filter` |
| scan base regressed to `Path(TESTS_DIR)`, run from `scripts/` | `scan base tests is not .../tests` |
| the two probes above, via the new governance test | exactly the 2 self-check cases fail; cwd-invariance cases correctly stay green |

The second probe initially **passed** when it should have failed, exposing a blind
spot in my own live-surface regex: `^\s*run:` missed the inline `- run:` form.
Fixed to `^\s*(?:-\s*)?run\s*:` before it could ship as a check that cannot fail.

## Correction to §3.2's scope table

The table was written from recollection, and three cells are wrong. Corrected in
the RFC with the measurements:

1. **28 hooks, 7 local — not 27 and 6.** The omitted row is
   `test-encoding-ratchet`, and its script was carrying exactly the defect §3.2
   exists to catch. As with the `#1314` codemap guard, the anti-staleness rule was
   violating itself.
2. **`tsa-ps-ascii` does have a live surface**, so requirement 2 applies — §3.2
   called it "a static character-class check over `.ps1`", but it watches 30 YAML
   files and the rule is about the 27 carrying a `run:` block.
3. **`block-banned-test-names` is a staged-diff gate**, not a live-tree gate.

Also recorded, as a negative result so nobody re-investigates: the cwd defect is
**not systemic**. `check_loose_assertions.py` already anchors on `__file__`; the
two shell gates read `git diff --cached --name-only`, which is
repository-root-relative regardless of cwd. All three were run from the root,
from `scripts/`, and from an unrelated subdirectory and behave identically.

## What is still missing for §3.2

`check_loose_assertions.py`, `check-test-file-names.sh`, and
`check-local-artifacts.sh` still lack a `--self-check` (requirements 1–2), and no
gate yet names its enforcement layer in-repo beyond this module. This PR does not
check off §3.2; the acceptance item stays unchecked.

## Checks

- `uv run pytest -q` → **2115 passed, 28 skipped** (2111 baseline + 4 new)
- `ruff check` + `ruff format --check` → clean on all three files
- `--self-check` verified green from the repo root, `scripts/`, and `/`
- RFC acceptance items unchanged: **23 checkboxes before and after, none checked**

### DIFF
--- a/rfcs/0028-measuring-claimed-properties.md
+++ b/rfcs/0028-measuring-claimed-properties.md
@@ -469,7 +469,50 @@ surface to be outside of. §3.2 therefore scopes to the **6 local hooks**:
 
 Third-party hooks are out of scope: they are versioned upstream, and a pinned
 `rev` bump is the review surface for their behaviour. Requirements 1 and 2 apply
-to the four "yes" rows; requirements 3 and 4 apply to all six.
+to the four "yes" rows; requirements 3 and 4 apply to all six. (Both counts are
+corrected below.)
+
+**Scope corrections, measured 2026-09-12** while implementing the first two
+self-checks. The table above was written from recollection of the config rather
+than from the config, and three of its cells are wrong. They are corrected here
+rather than silently edited, because a scope table that under-counts the local
+hooks by exactly the broken one is itself an instance of this RFC's subject.
+
+1. **28 hooks, 7 local — not 27 and 6.** `.pre-commit-config.yaml` declares 21
+   third-party hooks (`ruff` + `ruff-format`, the 14 `pre-commit-hooks`,
+   `detect-secrets`, `bandit`, `mypy`, `pyupgrade`, `actionlint`) and these 7
+   local ones: `weak-assertion-ratchet`, **`test-encoding-ratchet`**,
+   `workflow-consistency-tests`, `tsa-codemap-sync`, `block-banned-test-names`,
+   `block-local-artifacts`, `tsa-ps-ascii`. The omission is not cosmetic: the
+   missing row is `test-encoding-ratchet`, and its script
+   (`scripts/check_test_encoding.py`) was carrying the exact defect §3.2 exists
+   to catch — resolved against the caller's cwd, it reported
+   `encoding-unsafe text calls in tests/: 0 across 0 files` and exited `0` when
+   run from any directory but the repository root. The scope table omitted the
+   one local hook that was actually dead.
+2. **`tsa-ps-ascii` has a live surface; requirement 2 does apply.** The table
+   calls it "a static character-class check over `.ps1`". Measured, it watches
+   `.github/workflows/*.yml|*.yaml` and `.github/actions/**/action.yml|*.yaml`
+   (30 files), and the surface the rule is *about* is the 27 of those carrying a
+   `run:` block. That is a real set with a real watch filter, so the coverage
+   invariant is well-defined and is now asserted — and it fires on a planted
+   `.github/zz-probe.yml` that carries `run:` outside the filter.
+3. **`block-banned-test-names` is a staged-diff gate, not a live-tree gate.**
+   The table gives its surface as "the banned-pattern list vs the live test-file
+   set". Measured, it reads `git diff --cached --name-only --diff-filter=A`, so
+   it sees **newly added staged** test files only; the live test-file set is
+   never enumerated. Requirement 2 is still meaningful for it, but over
+   `^tests/.*\.py$` restricted to staged additions, not over the tree.
+
+One further finding, negative and therefore worth recording: the cwd defect is
+**not** systemic across the local hooks. `check_loose_assertions.py`,
+`check-test-file-names.sh`, and `check-local-artifacts.sh` were each run from
+`scripts/` and from an unrelated subdirectory and behave identically to the
+repository root — the first because it already anchors on
+`Path(__file__).resolve().parents[1]`, the latter two because
+`git diff --cached --name-only` reports repository-root-relative paths
+regardless of cwd. Measured 2026-09-12; this narrows the defect to the two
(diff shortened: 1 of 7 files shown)


aimasteracc:
Phase A complete (`3c0aa8da`). **All six** gates that can carry a self-check now do, and all six are executed by a collection layer. `blocks 1464` on nothing.

## Coverage

| local hook | self-check asserts | red→green |
|---|---|---|
| `tsa-codemap-sync` | exact set equality, MCP + CLI; 0 flags outside the watch filter | pre-existing (#1314) |
| `weak-assertion-ratchet` | exact violation set on a planted weak assert, no false positive, 100% parse coverage | yes |
| `test-encoding-ratchet` | scan base anchored to repo, non-empty surface, every live file parses | yes |
| `block-banned-test-names` | planted banned name detected exactly, innocent name ignored, path outside `tests/` not reported | yes |
| `block-local-artifacts` | each blocked pattern still matches exactly, innocent paths ignored | yes |
| `tsa-ps-ascii` | watched set vs live `run:` surface, difference empty | yes |

The seventh local hook, `workflow-consistency-tests`, is a `pytest` invocation rather than a script, so there is no flag to add. That is the one remaining gap against §3.2's self-check box, and the box is **left unchecked** (`23 items / 0 checked`, unchanged, zero removed lines).

## Two measured findings, recorded in the RFC

- **`check_loose_assertions.py` silently skips what it cannot parse.** `check_file` returns `[]` on `SyntaxError` — *"skip silently (CI lint step catches syntax errors)"* — so a tree of unparseable tests reads as clean. The self-check turns that into a hard failure; verified by planting a syntax-error file under `tests/` (`1 live files do not parse`).
- **`block-local-artifacts` can only fire on a force-added path.** Both patterns (`threads/`, `REDESIGN_PROPOSAL.md`) are already in `.gitignore`, so `git diff --cached` normally cannot contain them. Verified as a `git add -f` backstop by staging one and watching the gate reject it — rather than describing it as a gate that fires in normal use.

## Reverse verification

Every new self-check was broken and restored, with the raw red output kept:

| planted break | result |
|---|---|
| placeholder rule disabled | `expected {(3,'is-not-none','placeholder')}, got set()` |
| syntax-error file under `tests/` | `1 live files do not parse` |
| banned-pattern list emptied | `BANNED_PATTERNS is empty; the detector cannot fire` |
| `^tests/` filter removed | `path outside the tests/ filter was reported: 'src/widget_edge_cases.py'` |
| blocked-pattern list emptied | `BLOCKED_PATTERNS is empty; the gate cannot fire` |
| one blocked pattern renamed | `threads/ not detected exactly` |
| any gate degraded, via the governance test | exactly 1 failing case naming that gate |

## Collection (requirement 4)

`tests/governance/test_gate_self_checks.py` covers all six gates: **11 collected** — six self-check cases plus five cwd-invariance cases, each asserting a gate's output is byte-identical from the repo root and from `scripts/`.

## Checks

- `uv run pytest -q tests/` → **24,882 passed, 180 skipped, 0 failed**
- `uv run pytest -q` → **2,122 passed, 28 skipped in 30.6s** (baseline 2111/28/26.5s; cap 53s)
- Judge-standard files (`pytest.ini`, `.pre-commit-config.yaml`, workflows, the two governance tests, the ratchet's existing assertions) appear **nowhere** in `git diff --name-only origin/develop...HEAD`

One methodology note worth carrying forward: `pytest.ini`'s `testpaths` is a curated subset, so the documented quick gate collects **2,144** tests while CI collects **25,224**. I found two real regressions on a sibling branch that were invisible to the quick gate, so all verification above uses the full `tests/` tree.

Issue #1450, opened 2026-09-12
[P2] nav action=callers inlines callee bodies by default — should a caller be able to opt out?


aimasteracc:
# [P2] MCP tool surface costs ~8.7k tokens in every request — 127% of a host's entire native toolset, and 94% collapsible

## Summary

The MCP server registers 9 tools whose descriptions plus input schemas total **34,719 characters (~8,680 tokens at chars/4)**. That cost is paid on **every** model request by every MCP client, before the agent does anything. Measured against a real host harness's own assembled tool set, TSA's tool surface alone is **larger than that host's entire 27-tool native tool set**.

TSA's positioning is token-efficient agent code intelligence (TOON output, pre-indexed graph). The always-sent tool surface currently works against that.

## Evidence

Measured with the MCP SDK the host resolves (`@modelcontextprotocol/sdk` 1.29.0), against `tree-sitter-analyzer-mcp` **1.31.0 (macos-3.11)**, one process, `tools/list`:

```
connect   : 1,604 ms
tools/list:   507 ms
tools     : 9
```

| tool | total chars | of which description | of which schema |
|---|---:|---:|---:|
| `edit` | 7,276 | 4,799 B | 2,461 B |
| `nav` | 6,390 | 4,937 B | 1,455 B |
| `health` | 4,549 | 3,405 B | 1,132 B |
| `search` | 3,889 | 2,661 B | 1,203 B |
| `structure` | 3,368 | 2,226 B | 1,114 B |
| `project` | 3,286 | 2,244 B | 1,019 B |
| `index` | 2,926 | 1,788 B | 1,110 B |
| `viz` | 2,628 | 1,251 B | 1,343 B |
| `set_project_path` | 407 | 178 B | 172 B |
| **total** | **34,719** | | |

Descriptions, not schemas, dominate: `edit`'s description is 4,799 B against a 2,461 B schema.

**Host baseline.** The host harness's own recorded tool assembly contains **27 native tools totalling 27,285 characters (~6,821 tokens)**. TSA's 9 tools are **127%** of that — the single largest contributor to the tool-definition block, larger than every native tool combined.

[...]

PR #1473, opened 2026-09-13
perf(mcp)!: cut the tool-definition surface by 59% with progressive disclosure

aimasteracc:
Fixes #1450. Every request pays for the tool definitions before any work happens.

## Measured, on this repo

| | chars | ~tokens |
|---|---|---|
| tool-definition surface, before | 36,055 | **9,013** |
| — of which prose description | 23,084 | 5,771 (64%) |
| tool-definition surface, after | 14,716 | **3,679** |
| **cut** | | **59%** |

Your issue said ~8.7k tokens and 94% collapsible. The 8.7k is confirmed and slightly understated at 9,013; the 94% figure is what a *fully* collapsed schema-only form would reach, and this lands 59% while keeping the keyword sentence that tool search depends on.

On a 200k context, 9,013 tokens is a permanent **4.5%** of the window, for a tool a session may call twice. That is paid by every user on every request whether or not anything is called.

## What changes

`get_tool_definition()` emits one keyword-bearing sentence + the action list + `action=help`. `action=help` returns the full per-action prose, verbatim.

Wave D already moved per-action *parameters* out of the schema body into the description. This moves the prose out of the description. Nothing is lost — it moves behind the call, where it is needed only *after* an action is chosen.

## What is deliberately kept

The first sentence carries the `codegraph` keyword that a headless agent's ToolSearch matches on, pinned by `test_nav_facade_description_contains_codegraph_keyword`. A diet that broke tool discovery would trade one tax for a worse one.

`help` is **not** added to the action enum: that enum is a pinned exact contract (`assert expected == enum`), and the short description names the route instead.

## The drift guards are repointed, not weakened

Several tests pin that an action's documented parameters exist in that action's real inner schema — the check that caught `action=smart` documenting `task`/`limit` when the inner requires `file_path`, and `action=journal` documenting `operation`/`entry`, which the inner rejects. Those contracts are intact: `full_description()` exposes the prose and the tests read it there.

Assertions moved to the prose's new home; none was relaxed. `test_工具定义_默认描述含动作计数与清单` still requires the action count and the full action list on the always-sent surface.

## Checks

- `uv run pytest -q tests/` → **24,940 passed, 180 skipped, 0 failed**
- `uv run pytest -q` → **2,134 passed, 28 skipped in 31.3s**
- All 1,615 `tests/unit/mcp/tools/` tests pass

## Breaking?

`!` because the emitted `description` string changes shape. No parameter, action, or route changes; `action=help` is additive. A caller that parsed the old prose gets it from `help` instead.


aimasteracc:
Superseded by the new PR from `perf/tool-surface-tax` — the gitflow guard rejected the `experiment/` prefix. Same commits (`6dadc7ba`), no content changed.

PR #1484, opened 2026-09-14
perf(health): stop rediscovering plugins once per file

aimasteracc:
Third item from #1450. Follows the rule the new `AGENTS.md` measurement section states: cold and warm reported as a pair, with the machine and corpus.

## What the profile found

`_extractor_complexity_by_line` built a fresh `PluginManager()` for **every file** it scored:

```python
plugin = PluginManager().get_plugin(language)
```

Each construction runs `_discover_from_entry_points()` → `importlib.metadata.entry_points()`, which re-reads every installed distribution's metadata, plus a languages-directory scan. For 2,275 files that is 2,262 rescans of an answer that describes the installed environment and **cannot change while the process runs**.

Cold `health action=project`, profiled:

```
141,384,356 function calls in 62.276 seconds
  16.146s  {method 'acquire' of '_thread.lock' objects}          9,146 calls
   7.395s  {built-in method io.open}                           370,828 calls
   2.511s  {built-in method posix.listdir}                      25,030 calls
```

Tracing the callers:

```
pathlib.py:1036(open)          <- 352,626  importlib/metadata/__init__.py:931(read_text)
entry_points()                 <- 350,350  importlib/metadata/__init__.py:1038(<genexpr>)
load_plugins                   <-   2,275  manager.py:155(get_plugin)
get_plugin                     <-   2,262  complexity_heatmap.py:329(_extractor_complexity_by_line)
```

One call site. 350,350 metadata file reads for a per-process constant.

## The fix

A process-wide manager in `complexity_heatmap`, used by both the AST path and the plugin-fallback path in that module. The plugin registry is environment state, so one instance serves every file.

## Result

| primitive | before | after |
|---|---:|---:|
| `importlib.metadata.entry_points()` | 2,262 per run | **5 per run** (1 per process) |
| `io.open` | 370,828 | **18,358** |
| `posix.listdir` | 25,030 | reduced with the rescans |

Cold, cache removed before **every** run, same machine and repository:

| | run 1 | run 2 |
|---|---:|---:|
| before | 49.08 s | 49.39 s |
| after | **44.54 s** | **44.19 s** |

Warm is unchanged at ~4.5 s. This repository, this machine, macOS/aarch64, Python 3.11.15.

Roughly 10% of a cold call. I am not claiming more: the profiler exaggerates the call-heavy part (62 s instrumented vs 49 s uninstrumented), so the instrumented 17.6 s figure is not the wall-clock saving.

## Where the remaining time actually is

Not in discovery. After the fix the largest profile entry is `_thread.lock.acquire` at 17.4 s — but tracing its callers shows every one is `threading.py` condition-variable machinery, i.e. the **main thread waiting inside `concurrent.futures.as_completed` for the pool's workers**. That is not contention on a shared lock, and it is not removable by lock tuning: `ThreadPoolExecutor` cannot parallelise CPU-bound Python past the GIL.

The cold cost is real work — 2,275 parses, per-function cyclomatic extraction through the plugin extractor, and the scorer's AST walk (5.7M calls). Warm is ~4.5 s because `health_scores.db` already holds it.

So the honest conclusion for this issue is that cold `health action=project` is ~45 s on a mid-size repository and the remaining lever is architectural (process-level parallelism, or not recomputing what the index already holds), not another local fix. A caller still cannot discover the requirement before it fails.

## Verification

- `23,897 passed, 158 skipped` across `tests/unit`, `tests/contracts`, `tests/governance`; `708 passed, 21 skipped` on the docs/contracts/governance subset.
- Weak-assertion ratchet, codemap-sync, test-file-name and local-artifact gates green; ruff and mypy clean.
- The regression test drives `_extractor_complexity_by_line` twenty times and asserts exactly one entry-point scan. Reverting the call site fails it with *"twenty files performed 20 entry-point scans"* — I wrote it against the helper first, and it passed while the caller was still broken, which is the same defect shape the new `docs/AI_LESSONS.md` entry on patch coverage describes.


aimasteracc:
Correction to my own verification numbers.

I wrote `23,897 passed, 158 skipped` for `tests/unit tests/contracts tests/governance`. The merged tree reports **23,899**. I took the figure before adding the last two tests in this PR (`TestPluginDiscoveryIsNotPerFile`), and did not re-run after. Re-measured four times on the merged tree, including with and without `-p no:randomly`: 23,899 every time, so the tree is stable and only my number was stale.

The rest of the body stands: 2,262 entry-point scans -> 5, `io.open` 370,828 -> 18,358, and cold 49.08/49.39 s -> 44.54/44.19 s.

PR #1489, opened 2026-09-15
fix(ci): preserve dependency policy and target containment

aimasteracc:
Dependabot's valid `actions/upload-artifact@v7` selector was rejected by a diagnostics test pinned to `v7.0.1`. Verification planning also repeated parent-directory resolution for sibling test targets.

This PR checks the artifact action's identity and major version, and caches parent resolution within one verification-planning call. Target types are checked individually using native filesystem operations, preserving Windows case semantics without scanning unrelated directory entries. Regression tests preserve nested pytest configuration, absolute-path, symlink, directory, and `nested/..` containment behavior.

The first Windows CI run caught an existing test whose fault injection targeted leaf resolution removed by the optimization. The follow-up uses a real nested file and injects both resolution errors at the required parent boundary, keeping the exact fail-closed assertion. The related lesson records the omitted caller test.

Related to #1478.

Validation on the final reviewed file tree:

- 145 focused tests passed, 2 tracked platform skips, 0 failures or reruns (38.85 s; Linux, Python 3.12.14; 4 workers with work stealing).
- Focused files: `tests/unit/test_verification_plan.py`, `tests/unit/test_classify_windows_pytest_failure.py`, `tests/contracts/test_agent_docs_contract.py`, and `tests/unit/mcp/test_verification_command.py`.
- The skips require native Windows: #1407 PowerShell 5.1 and the #1489 case-insensitive directory contract.
- Fresh isolated coverage and unchanged before/after Python source hashes.
- Patch-coverage gate against the actual develop base `7e1ed4f5`, with `--worktree` and the exact project root: passed with no added executable misses.
- Ruff formatting/lint, strict MyPy for the changed production module, and `git diff --check` passed.

Codex review triage (all accepted and addressed in the latest commit):

- #discussion_r4014050998: removed the unsupported latency claim and recorded the known CI command, runner, and corpus.
- #discussion_r4014051012: added the dated incident/PR reference to the performance regression test.
- #discussion_r4014051023: fixed case-insensitive target classification and added a native Windows contract.
- #discussion_r4014051028: removed whole-directory enumeration and added a no-enumeration contract.

The change is limited to 6 files. These are local focused results; the full quick suite and native platform matrix were not rerun locally. GitHub Actions validates the updated head.


aimasteracc:
@codex review

Please review the current head `f02e2e1c4671a53fbb10af1c9d32531ec7061b9f`. The four findings from the earlier reviewed head are implemented, their outdated threads are resolved, and the focused final-tree qualification was independently rerun: 145 passed, 2 tracked platform skips, 0 failures.

Issue #1402, opened 2026-09-08
Track unresolved macOS EMFILE at 256-FD snapshot test; #1400 regression unconfirmed


aimasteracc:
## Status

Track the unresolved historical EMFILE seen during #1400 validation. **Whether this is a #1400 regression is not established.** This issue does not attribute the failure to a particular production source file. No PR merge or source/test fix is part of this investigation.

## Original Evidence

- Platform/interpreter: macOS, CPython 3.14.3, xdist worker `gw1`.
- Exact nodeid: `tests/unit/test_index_source_snapshot.py::test_revalidate_source_rows_is_exact_below_rlimit_with_300_files`.
- Existing test at lines 883-903 creates 300 flat source files, sets `RLIMIT_NOFILE.soft = min(256, original_soft)`, and restores the original limit in `finally`. The test is unchanged between the compared SHAs.
- Traceback: test line 899 -> `tree_sitter_analyzer/index_source_snapshot.py:290`, `root_fd = os.open(root, directory_flags)` -> `OSError: [Errno 24] Too many open files`.
- Failure occurs on the first root-directory open, before this invocation traverses the 300 leaves.
- Original evidence: `tsa-issue1376-safety-cache-evidence/integration-8b4f21e7/utf8-comprehensive-once.log:454-462,582-584`.
- Original run reported 22 failed, 23866 passed, 165 skipped, 1 xfailed and 22 reruns. There is one final EMFILE nodeid, not evidence of exactly one failing attempt. No comprehensive suite was rerun during this investigation.
- The evidence directory names the integration base; it is not a before-migration full-suite control run.

## Bounded A/B, 2026-09-08

Exactly one independent run per side, each hard-capped at 300 seconds:

- Before: `8b4f21e70ce470db6114be36dd22d153955f71be`.
- After: `fe795f1332e206f8c5e85c5e32117e9f171b56c2` (final #1400).
- The migration under investigation is `355067a53ee7fe264cd7fe57f8ca6f236fc4cc89`.
- Same existing Python 3.14.3 virtual environment; installed package/version inventories compared equal before and after, with no installs or environment configuration changes.
- Reused the existing 332 safety + 248 cache nodeid mappings; added the unchanged original failing node. Exact equality of 581 logical selected nodes verified across both sides and four workers.
- Four xdist workers, `--dist=worksteal`, `--timeout=120`, existing rerun defaults. No worker reduction, per-test automatic GC, raised FD limit, or added skip.
[...]

PR #1500, opened 2026-09-16
fix(journal): close sqlite connections deterministically

aimasteracc:
`DecisionJournal` used `sqlite3.Connection` as a context manager, which commits or rolls back but does not close the connection. Each operation therefore retained a `decision_journal.db` descriptor until garbage collection, matching one resource family observed in #1402.

This change nests `contextlib.closing` around every transaction so initialization, record, get, search, and supersede close deterministically after transaction exit. The regression test instruments real SQLite connection objects and requires the live connection count to return to zero after every operation.

Validation:
- TSA change impact: 2 files, high risk, 188 downstream
- focused verification with coverage: 23 passed
- patch coverage: no added executable misses
- quick gate: 2142 passed, 28 skipped
- pre-commit: passed
- low-limit FD probe: soft limit 64, 1,501 operations, FD 4 → 4 (delta 0)

Refs #1402


aimasteracc:
@codex review

aimasteracc:
@codex review

PR #1501, opened 2026-09-16
fix(change-impact): execute benchmark claim targets

aimasteracc:
When `change-impact` selected a benchmark claim test by exact path, it still emitted the default `-m 'not network and not benchmark'` selector. Pytest therefore started successfully but selected zero target tests, producing false green verification guidance.

This change removes only the contradictory `not benchmark` term when the selected target is under `tests/benchmarks/claims/`. It preserves `not network`, handles POSIX paths, Windows separators, and pytest nodeids, and leaves ordinary benchmark paths excluded.

Validation:
- TSA scoped change-impact and lesson contract: passed
- focused coverage: 58 passed, 1 skipped
- patch coverage: no added executable misses
- quick gate: 2142 passed, 28 skipped
- pre-commit: passed

The agent lesson records target reachability as part of the generated-command contract.


aimasteracc:
@codex review

